Issue #3809, opened 2018-04-26
Compression interferes with SSE events


paulmillar:
If a compression content encoder (such as GZipEncoder) is registered and the client requests a supported compression when making an SSE GET request then the client (apparently) does not receive events although the server has sent them.

The effect is caused by the compression encoder buffering data before writing output to the client.  Although the server issues OutputStream#flush after writing the event, this does _not_ trigger sending the outstanding data.

Note that there is a flag (`syncFlush`) that would support sending outstanding data on flush; however, it is currently not possible to specify this flag.  Moreover, setting the flag would affect all responses, including those targeting non-SSE endpoints.  The syncFlush flag (potentially) has a detrimental effect on the compression ratio.  Therefore, syncFlush should be enabled for SSE responses.

mseele:
We ran into this issue, too. Because of `EncodingFilter` and `ContentEncoder.aroundWriteTo` being final we decided to create this **ugly** workaround:

Adding the following filter AFTER EncodingFilter will not compress SSR requests.

```java
/**
 * Workaround https://github.com/eclipse-ee4j/jersey/issues/3809 for SSE requests
 * (by removing the accept/encoding header on SSE requests).
 */
@Priority(Priorities.HEADER_DECORATOR)
public static class ServerSentEventEncodingSpeciality implements ContainerResponseFilter {
    @Override
    public void filter(ContainerRequestContext requestContext, ContainerResponseContext responseContext) throws IOException {
        MediaType type = responseContext.getMediaType();
        if (type != null && type.isCompatible(MediaType.SERVER_SENT_EVENTS_TYPE)) {
            requestContext.getHeaders().remove(HttpHeaders.ACCEPT_ENCODING);
        }
    }
}
```

I understand the point of @jansupol in https://github.com/eclipse-ee4j/jersey/pull/3810#issuecomment-400718602 that he wants to avoid a general `syncFlush` for all SSE requests. But having a **special** media type to avoid encoding is IMHO also not a good solution for this problem. This might cause errors with client implementations who rely on the fact, that a SSE request respond with `text/event-stream`.

The best and universal way would be to trigger the `syncFlush` via an annotation at the endpoint method. Is there any annotation we can use for this?
Or we add a possibility to enable/disable the special SSE functionality with a global variable in `EncodingFilter`.
Another solution would be to remove the `final` keyword in the `ContentEncoder` class, btw. That would make it possible to override the encoders with an own solution.

[...]

PR #3810, opened 2018-04-26
Bug 3809 Fix compression not to interfere with SSE

paulmillar:
Motivation:

The gzip and deflate compression implementation, if enabled and
requested by the client, accepts some 8 kiB of data before sending any
compressed data to the client.  It may take considerable time (with the
server sending many SSE events) before sufficient data is collected and
any compressed data is sent to the client.

Modification:

Add support for detecting time-critical responses, based on the declared
MIME type of the response.

Add the possibility for a ContentEncoder to provide a different encode
wrapping for time-critical responses.

Update GZip and Deflate ContentEncoders to enable syncFlush flag for
time-critical responses.  With this flag, a flush results in any pending
data being compressed and sent.  This will likely result in poorer
compression ratio.

Result:

Compressed SSE data is delivered without unnecessary delays.

Issue: https://github.com/eclipse-ee4j/jersey/issues/3809
Signed-off-by: Paul Millar <paul.millar@desy.de>

### DIFF
--- a/core-common/src/main/java/org/glassfish/jersey/message/DeflateEncoder.java
+++ b/core-common/src/main/java/org/glassfish/jersey/message/DeflateEncoder.java
@@ -84,6 +84,18 @@ public class DeflateEncoder extends ContentEncoder {
     @Override
     public OutputStream encode(String contentEncoding, OutputStream entityStream)
             throws IOException {
+        return encode(entityStream, false);
+    }
+
+    @Override
+    public OutputStream timeCriticalEncode(String contentEncoding, OutputStream entityStream)
+            throws IOException {
+        return encode(entityStream, true);
+    }
+
+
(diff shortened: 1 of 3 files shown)


jansupol:
You have a good point with this issue, but the solution basically says that anybody who is using sse media type starts syncFlush the  DeflaterOutputStream. While for your time-critical application this can be useful, for other sse use cases this would not be preferred.  My comments to this PRs are:
1) The decision would better be set on mediaType such as "text/event-stream+timecritical", or it would be even better set on an Application level. With upcoming Microprofile config/config jsr we may also think on setting the config. I understand its a long-term run, though.
2) I would stick on jdk naming "syncFlush" rather then "time-critical", as there may be other use-cases for syncFlush than "time-critical", the media type "x/y+timecritical" is merely one of them.

PR #3812, opened 2018-04-27
issue 3804 fix (SelectableEntityFiltering when the fields specified is . character)

senivam:
Tokenizer is fixed to allow valid empty filters. The fix will prevent Jersey to fail after empty filter request. The empty object is returned instead. 

Issue #3804 


jansupol:
+1

PR #3813, opened 2018-04-30
Upgrade netty dependency to 4.1.24.Final, Replease Http2Codec which r…

### LINKED ISSUE #3697: Using unstable netty api that has been removed broke jersey-container-netty-http
I do not have a solution at this time. Not that familiar with netty, and seems replacement class is not a drop in replacement. Jersey is using a unstable Netty api that was present in <=4.1.14, but removed in versions since >=4.1.15, in this [commit](https://github.com/netty/netty/commit/74f24a). Which breaks [HttpVersionChooser](https://github.com/jersey/jersey/blob/master/containers/netty-http/src/main/java/org/glassfish/jersey/netty/httpserver/HttpVersionChooser.java#L72) and [JerseyServerInitializer](https://github.com/jersey/jersey/blob/master/containers/netty-http/src/main/java/org/glassfish/jersey/netty/httpserver/JerseyServerInitializer.java#L140) usage of the removed class, Http2Codec.

I briefly looked into replacing Http2Codec with Http2MultipleCodec but was unsuccessful. I may see about a PR time permitting.Till then mentioning for others to be aware, and for the record.

Noodle05:
…emoved since netty 4.1.15.Final. Create a netty jersey server builder to use builder pattern

Fix #3697

### DIFF
--- a/containers/netty-http/src/main/java/org/glassfish/jersey/netty/httpserver/HttpVersionChooser.java
+++ b/containers/netty-http/src/main/java/org/glassfish/jersey/netty/httpserver/HttpVersionChooser.java
@@ -20,7 +20,7 @@ import java.net.URI;
 
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.handler.codec.http.HttpServerCodec;
-import io.netty.handler.codec.http2.Http2Codec;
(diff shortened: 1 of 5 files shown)


jansupol:
Closing in favor of #3983

Issue #3700, opened 2017-10-17
Registering resources in ResourceConfig leads to warnings


jerseyrobot:
Hi, 

we register Resources using the ResourceConfig.register(Object object) method. This works and has worked until now. However now (2.26) we are getting a warning:

`Oct 17, 2017 1:47:00 PM org.glassfish.jersey.internal.inject.Providers checkProviderRuntime
WARNING: A provider war.HelloWorldResource registered in SERVER runtime does not implement any provider interfaces applicable in the SERVER runtime. Due to constraint configuration problems the provider war.HelloWorldResource will be ignored. 
`

It looks like [the check assumes that no resources are registered in this way](https://github.com/jersey/jersey/blob/master/core-common/src/main/java/org/glassfish/jersey/internal/inject/ProviderBinder.java#L199). Is this a bug or are we misusing the ResourceConfig API?

I've attached a [minimal reproducing case](https://github.com/jersey/jersey/files/1390890/register-resource-warning.tar.gz) based on the jersey-example-java8-webapp archetype. Running war.App leads to the warning above.

Thanks & best regards,
  Silvestre

Tibor17:
Can you pls test this commit and push it to master? The dropwizard project wants to have a fix, see https://github.com/dropwizard/dropwizard/issues/2418, https://github.com/dropwizard/dropwizard/pull/2395

bcalmac:
The same is true about spring-boot.

jansupol:
@Tibor17 We cannot merge unless the commit contains the sign-off, I am sorry.

serhiypal:
@jansupol how do we get a sign-off?

jansupol:
@serhiypal `git commit -s`. Please see the [wiki](https://github.com/eclipse-ee4j/jersey/wiki/Missing-ECA,-Missing-Sign-off).

serhiypal:
@jansupol. Ah, thanks. Then what about this one: https://github.com/eclipse-ee4j/jersey/pull/3918.
It looks pretty similar to what is in https://github.com/objecttrouve/jersey-1/commit/f81ab991ca28d373aeca322b4ad5b58c67ffc6f6 but is more accurate replication of missing logic from [ApplicationHandler](https://github.com/jersey/jersey/blob/2.25.1/core-server/src/main/java/org/glassfish/jersey/server/ApplicationHandler.java#L982)

senivam:
[...]

PR #3817, opened 2018-05-04
#3700: Treat resource instances as resources.

objecttrouve:
Cosmetic elimination of a warning. 

The intuition is that if resource classes are OK, resource instances should be, too. 

The predicate that tells if a component class is `correctlyConfigured` only checked the `resourceClasses` collection to determine whether the resource is actually a resource.
As a consequence a warning, was logged when an _instantiated_ resource class was tested:

`WARNING: A provider war.HelloWorldResource registered in SERVER runtime does not implement any provider interfaces applicable in the SERVER runtime. Due to constraint configuration problems the provider war.HelloWorldResource will be ignored.
`

The change is as conservative as possible.
Changed the result of the relevant predicate only in the context where the changed result values should not have any impact.
But the side-effect of logging a warning should go away.

No additional test since the change shouldn't have any observable effect other than the absence of a warning.

### DIFF
--- a/core-server/src/main/java/org/glassfish/jersey/server/ResourceModelConfigurator.java
+++ b/core-server/src/main/java/org/glassfish/jersey/server/ResourceModelConfigurator.java
@@ -116,12 +116,23 @@ public class ResourceModelConfigurator implements BootstrapConfigurator {
         /*
          * Check the {@code component} whether it is correctly configured for client or server {@link RuntimeType runtime}.
          */
-        java.util.function.Predicate<Class<?>> correctlyConfigured =
+        java.util.function.Predicate<Class<?>> correctlyConfiguredComponentClass =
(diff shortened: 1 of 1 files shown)


jansupol:
References #3700

serhiypal:
It is not an intuition it is a regression in the logic. This block https://github.com/jersey/jersey/blob/2.25.1/core-server/src/main/java/org/glassfish/jersey/server/ApplicationHandler.java#L982 was not ported over to `2.27`.

senivam:
please do not use instance registration to register resource instances without any additional container implementation. I was explaining difference between instance and class registrations (they are really not the same) in bug #3700 and in PR #3918 the warning is thrown intentionally and there is no need to fix that.

objecttrouve:
Thanks for the clarification. Closing the request.

PR #3820, opened 2018-05-04
Allow to register the resource by instance or proxy.

DangCat:


mkarg:
@DangCat Can you please provide an _at least minimal_ description of this PR?

dhenneke:
@DangCat are you planning to update the PR with the requested changes?
@senivam is there a chance for us to resubmit the PR in case the original author abandoned it?

I would like to see this PR merged since we need this change for our use-cases.

senivam:
@dhenneke it would be IP violation if we re-submit someone's code by another author... So it is very preferable for original author to perform requested changes...

dhenneke:
Ok. I already expected that this would not be possible.

jansupol:
@dhenneke You can use the original commit, grab the files and commit it with your sign-off, but with the original author `git commit --author="DangCat <fan.shutian@zte.com.cn> -s"`. Then you can add your changes on top of this commit.

raphw:
Not sure if this PR is still alive but I really wish there was a solution to this. I'd happily reimplement this feature "in my own words", if this would remmedy the missing sign off?

PR #3822, opened 2018-05-05
Removed reference to JavaFX class javax.util.Pair

chkal:
There is currently a single class in the Jersey code which depends on a JavaFX class: The test `GenericEntityTest` uses `javax.util.Pair`. Unfortunately this prevents the Jersey build to work with OpenJDK, because OpenJDK doesn't ship with JavaFX (yet?).

This pull requests replaces `javax.util.Pair` with a simple implementation which was added as a static inner class to `GenericEntityTest`. The test runs find with OpenJDK after applying this patch.


arjantijms:
The custom `Pair` class is a good no-dependeny option. An other option is possibly: `java.util.AbstractMap.SimpleEntry<K, V>`, which is effectively a Pair as well with interface `java.util.Map.Entry<K, V>`.

Issue #3803, opened 2018-04-09
Base64 encoding with non ASCII characters results in ArrayIndexOutOfBoundException


jerseyrobot:
To calculate the authentication string in org.glassfish.jersey.client.authentication.BasicAuthenticator the method org.glassfish.jersey.internal.util.Base64.encodeAsString() is used to do the Base64 encoding.
Unfortunately this results in a ArrayIndexOutOfBoundException when the password contains non ASCII characters like § (the paragraph sign) or a German Umlaut.

Some examples:
  * Base64.encodeAsString(new byte[]{(byte)128})
  * Base64.encodeAsString("§")

mdurawa:
The easiest way to fix it, would be to replace org.glassfish.jersey.internal.util.Base64 with java.util.Base64. That would affect about 10 files. (I could try to do that, but I'm not familiar with github yet.)

Btw. in the test classes the behavior not to support bytes > 127 seems to be intended. Is there any reason for that?

jansupol:
@srbala Why deprecate? The usual contract is that whatever is internal package is not be used by users, it can be changed/dropped any release...

srbala:
In that case, class can be dropped.

maffe:
This issue can be closed, #3829 was merged, the Base64 class does not exist anymore and a fixed version was released.

PR #3829, opened 2018-05-15
Replaced internal Base64 by java.util.Base64 (#3803)

mdurawa:
Using java.util.Base64 instead of core-common.src.main.java.org.glassfish.jersey.internal.util.Base64 allows to encode and decode characters > ASCII 127 (e. g. German umlauts).


srbala:
@mdurawa Add test method to validate #3803 issue.

mdurawa:
@srbala the tests are already there.
I updated core-common/src/test/java/org/glassfish/jersey/internal/util/Base64Test.java (it's the test class for the deleted class - maybe that could be marked as deprecated and removed in a future release):

As mentioned in the issue these tests checked that non Ascii 127 characters are not supported, I don't know the reason for this. Now they are fine with those characters/bytes.

- testRoundtripOneByteGreaterThan127()
- testRoundtripAssortedValues()
- testEncodeByteArray()

jansupol:
I am not sure why the IP validation check did not take place here, but I do not see the sign-off. Can you please resubmit with -s ? Otherwise it looks good. Thanks

mdurawa:
After some error messages on the client side, the sign-off message seems to be in the repository.

mdurawa:
How can I fix the test of e2e-entity?
I get the same error with a clean copy of the project master.

Issue #3816, opened 2018-05-03
Failed test: ClientDestroyTest.testLifecycleListenerProvider:224 Filter was expected to be still open.


mkarg:
# Problem

Cloned Jersey then did `mvn package`. The result is:

```
Mai 03, 2018 11:00:38 PM org.glassfish.grizzly.http.server.NetworkListener shutdownNow
INFORMATION: Stopped listener bound to [localhost:9998]
Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 5.355 sec - in org.glassfish.jersey.tests.e2e.client.TimeoutTest

Results :

Failed tests:
  ClientDestroyTest.testLifecycleListenerProvider:224 Filter was expected to be still open.

Tests run: 223, Failures: 1, Errors: 0, Skipped: 8
```

# Environment
```
$ mvn -version
Apache Maven 3.5.3 (3383c37e1f9e9b3bc3df5050c29c8aff9f295297; 2018-02-24T20:49:05+01:00)
Maven home: C:\Program Files\apache-maven-3.5.3
Java version: 1.8.0_144, vendor: Oracle Corporation
Java home: C:\Program Files\Java\jdk1.8.0_144\jre
Default locale: de_DE, platform encoding: Cp1252
OS name: "windows 8.1", version: "6.3", arch: "amd64", family: "windows"
```

mkarg:
@jansupol Was this my fault? Does Jersey need special build handling?

jansupol:
No special build handling I know of. `mvn package` works just fine.

mkarg:
I was just asking because some years back I was told by the Jersey team that mvn package will fail unless more memory is provided AFAIK. So this limitation got removed?

mkarg:
@chkal Maybe it is only a feeling but I think that the Jersey test suite is rather unstable. We maybe should focus on stabilizing this first before addressing other issues. At least for me, I was not able to pass the tests a single time so far. :-(

jansupol:
There is public jenkins job building Jersey https://ci.eclipse.org/jersey/job/jersey-master-continuous/. So far there are two JAXB tests failing (JDK8b161+ issue) +2 tests failing with Felix (not sure why yet), but  ClientDestroyTest seems to work fine

chkal:
[...]

PR #3830, opened 2018-05-15
Added Travis CI.

Tomas-Kraus:
Another attempt which finally seems to work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,16 @@
+language: java
+
+jdk:
+  - oraclejdk8
+
+cache:
+  directories:
+    - .autoconf
+    - $HOME/.m2
+
+install: true
+
+script:
+  - mvn -e -U -B clean install -fn 2>&1 | tee jersey-build.log | grep -E '(---)|(Building)|(Tests run)|(T E S T S)'
+  - echo '------------------------------------------------------------------------'
+  - tail -100 jersey-build.log
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,6 @@ cache:
 install: true
 
 script:
-  - mvn -e -U -B clean install -fn 2>&1 | tee jersey-build.log | grep -E '(---)|(Building)|(Tests run)|(T E S T S)'
+  - travis_wait 120 mvn -e -U -B clean install -fn 2>&1 | tee jersey-build.log | grep -E '(---)|(Building)|(Tests run)|(T E S T S)'
   - echo '------------------------------------------------------------------------'
   - tail -100 jersey-build.log


mkarg:
@jansupol Is there a good reason for _neither_ commenting _nor_ reviewing this PR for 12 days by _any_ committer? As you know, over at JAX-RS API we have this 14-days-limit. It would be good if we can come to _frequent and prompt_ reviews in Jersey, too. If there are too few _active_ committers, @chkal and me can help out -- but we need to be committers then, certainly.

Tomas-Kraus:
Markus, that's fine. We are still having some issues with Travis besause of its various limitations.
Unfortunately this version is still not working perfectly. But we are still trying to find some solution.
Also I'm sitting some 2 meters from Jan in the same office so don't worry, I have some good ways how to push him if really needed. :D

mkarg:
@Tomas-Kraus Good to hear. In fact the question was not limited on this particular PR. I noticed that there are several PRs without comments / reviews and I wonder why.

Tomas-Kraus:
This code is included in PR 3835 so let's close this one.

chkal:
@jansupol @Tomas-Kraus 

> Also I'm sitting some 2 meters from Jan in the same office so don't worry, I have some good ways how to push him if really needed. :D

I absolutely understand that you are communicating a lot directly if you are in the same office. However, there are a lot of people who want to help and want to start contributing. And it feels weird that there is some kind of developing happening behind the scenes without any public communication. That's why [I already asked](https://github.com/eclipse-ee4j/jersey/issues/3816#issuecomment-388121185) if there is some other communication channel used by the committers. There are currently 32 committers listed on the Eclipse Jersey page. And I guess not all of them are in the same office.

Don't get me wrong. I really appreciate the work you are doing. But it would be great if the community could participate more.

Tomas-Kraus:
Community has this available as PR 3835 - there are diffs and comments. We would like to see all of you helping us to move forward.
But at this stage it's not that easy. We don't have even basic CICD things like PR checks working.
So first of all we need to get those things up to make sure that every PR passes build, tests, coding style and other checks. Then we can start processing other things.

mkarg:
Agreed in general, but in this special case Christian maybe could be helpful. He set up the Travis stuff for JAX-RS API recently.

mkarg:
Proposal: (A) I could talk to the Travis guys, maybe they allow the Eclipse Foundation to use larger limits on STDOUT size and execution time. (B) Maybe "mvn -T 2C" will speed up processing, as Maven is rather I/O intensive, hence, its default sole thread blocks most of the time.

Tomas-Kraus:
(A) would be great. If you have an option to ask for it, please give it a try. (B) making it even more parallel when there is already enough I/O load may make it worse. But I have no idea what's the default for -T and we won't loose anything with giving it a try.

Tomas-Kraus:
Challenge accepted. :) Jersey tests are not best candidate at this moment because of server ports. But we have a lot of other projects with clean junits where it makes sense.

mkarg:
@Tomas-Kraus The Travis CI guys told me they will soon extend Jersey's timeout to 3 hours. Is this long enough? They cannot make it larger than that. They also do not have a STDOUT length limitation but say that there is a timeout limitation that stops logging; they also extend it to 3 hours. Let's see if this helps. They agreed to do this change for all projects I name them. So if it helps please let me know so I can ask the PMC if others had similar issues. Shall I ask them if they can set up a facility for data only images?

Tomas-Kraus:
The only problem I currently know about is Jersey. It will be great if they can extend stdout timeout to 3 hours for it. Another big projects are glassfish (not yet donated) and eclipselink.
Now the Docker things. EclipseLink and Jersey builds can be optimized if I can do
- build without tests first 
- and then use it to run matrix of tests using several isolated instances of already built sources
Data only image is a good way to achieve this. But I also need tools to build such an image using docker and create container from it and link it as java/ant/maven container directory.

With EclipseLink situation is even more complicated because we need to test it with matrix of databases. I use set of Docker images with various databases to do it.
So typical EclipseLink Docker setup are 3 containers linked togeher:
- built sources (instance of data only image)
- java/ant/maven/... environment
- database (MySQL, Oracle, Derby, DB2, ...)

Without data only image we have to clone and build project in each of the container which is just wasting of resources.

mkarg:
@Tomas-Kraus Regarding the idea of compiling once, storing the binaries and then running the tests I would say that this is possible. You need [Build Stages (BETA)](https://docs.travis-ci.com/user/build-stages/) with a first stage doing mvn -DskipTests=true install and then storing to a cache. The second (parallel) stages will run the tests using the binaries stored in the cache. If you need multiple caches, you can name them. For larger data to be shared, there is an [example using S3](https://docs.travis-ci.com/user/build-stages/share-files-s3/). If you need help with this, I can forward the Travis guy to you, but I doubt that he will mount docker data volumes. Maybe I miss something, by why shouldn't that work?

Tomas-Kraus:
I can give it a try. If there are stages and it's possible to pass sources tree and maven repo between them, there is a chance that it may work.
I'm quite busy those days, but I can try some sample to understand how it works.

Tomas-Kraus:
I was doing some experiments on https://github.com/Tomas-Kraus/jsonp
The small hack I did is in adding $HOME/build into cache directories. It breaks git so I have to update branch and revision manually, but is seems to work.
In this case it would be nice to disable git in 2nd stage where whole soource tree is restored from cache. But I have no idea how.

Issue #3828, opened 2018-05-15
SseEventSource#target(null) should throw NPE


jansupol:
Javadoc on SseEventSource#target says:

> NullPointerException in case the supplied web target is {@code null}.

Need to add a check on `JerseySseEventSource.Builder#target()` for null.

PR #3831, opened 2018-05-16
SseEventSource#target(null) should throw NPE

senivam:
Fix for issue: 

#3828 


jansupol:
+1

PR #3833, opened 2018-05-21
Test issue related to JRE version

senivam:
Tests for one JAXB-JSON binding class shall be excluded from executions due to JRE internal issue which prevents JSON correct marshalling/unmarshalling. The JRE issue occurs in 3 patches of the JDK8: 1.8.0_161, 1.8.0_162, 1.8.0_172. Other JRE/JDK versions are just fine. 

Thus tests (3 in total) for one JSON class shall be excluded from testing for indicated JRE versions.


jansupol:
We should not remove the code after 181+ is out, first of all people should be able to run the tests no matter the jdk bug and the jdk can be given on a provided environments such as Travis. 
But +1 for spotting 171

PR #3835, opened 2018-05-21
Travis CI build script

senivam:
Configuration for Travis CI build. Including changes in pom files (additional profiles for sequential output are added), configuration files (separate bash script and travis.yml config files are added).

For now the build is fully sequential, but it's supposed to modify it to be parallel (multiple build threads in Travis CI).

The whole build time for now is 45 minutes. The whole output (log) is 2.5 MB

Build is run as clean install targets for main Jersey sources (with main travis profile) and when this part is finished the only several tests (e2e and integration) are run as test targed (e2e travis profile) 

### DIFF
--- a/connectors/jdk-connector/src/main/java/org/glassfish/jersey/jdk/connector/internal/HttpConnection.java
+++ b/connectors/jdk-connector/src/main/java/org/glassfish/jersey/jdk/connector/internal/HttpConnection.java
@@ -127,7 +127,7 @@ class HttpConnection {
         });
     }
 
-    synchronized void close() {
+    void close() {
         if (state == State.CLOSED) {
             return;
         }
@@ -218,6 +218,9 @@ class HttpConnection {
     }
 
     private void changeState(State newState) {
+        if (state == State.CLOSED) {
+            return;
+        }
         State old = state;
         state = newState;
 
@@ -414,9 +417,11 @@ class HttpConnection {
 
         @Override
         void processConnectionClosed() {
-            cancelAllTimeouts();
-            changeState(State.CLOSED_BY_SERVER);
-            HttpConnection.this.close();
+            synchronized (HttpConnection.this) {
+                cancelAllTimeouts();
+                changeState(State.CLOSED_BY_SERVER);
+                HttpConnection.this.close();
+            }
         }
 
         @Override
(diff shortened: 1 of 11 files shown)


Tomas-Kraus:
Added Jan to check this too.

pom.xml
```diff
@@ -1172,6 +1173,19 @@
                 </pluginManagement>
             </build>
         </profile>
+        <profile>
+            <id>travis_main</id>
```
jansupol: This profile is used for skipping e2e tests, that take long. From `main` it is not clear what it does, is travis_e2e_skip better?

pom.xml
```diff
@@ -1172,6 +1173,19 @@
...
+            <properties>
+                <run.e2e>true</run.e2e>
```
jansupol: <skip.e2e>?

connectors/jdk-connector/src/main/java/org/glassfish/jersey/jdk/connector/internal/HttpConnection.java
```diff
@@ -127,7 +127,7 @@ public void completed(HttpRequest result) {
         });
     }
 
-    synchronized void close() {
```
jansupol: `close` should remain synchronized to prevent running `filterChain.close();` multiple times.
The deadlock can be better solved by disallowing a change from State.CLOSED to other states

jansupol:
@mkarg The java code modifications are due to issues that Jersey hits on Travis. So it makes sense to put the changes here. The first change is due to a deadlock that is caused when JerseyTest cleans up. A fix for that can actually help with other reported test failures. I agree that the pull requests should be explained.

Tomas-Kraus:
Last build failed. ...giving it one more try, but this is not good.

mkarg:
@jansupol Thanks a lot for the explanations. I understand that it is some kind of challenge for your team to work "publicly", but descriptions like yours are needed so new people like Christian and me can get onboard eventually. :-)

connectors/jdk-connector/src/main/java/org/glassfish/jersey/jdk/connector/internal/HttpConnection.java
jansupol: Actually this is ok, it is already called from a synchronized blocks, except `processConnectionClosed`.

PR #3839, opened 2018-05-26
Support for JAX-RS 2.2 Java SE Bootstrapping API

mkarg:
Implementation of the new *Java SE Bootstrapping API* of **JAX-RS 2.2**.

### DIFF
--- a/containers/grizzly2-http/pom.xml
+++ b/containers/grizzly2-http/pom.xml
@@ -41,6 +41,12 @@
             <groupId>org.glassfish.grizzly</groupId>
             <artifactId>grizzly-http-server</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.hamcrest</groupId>
+            <artifactId>hamcrest-junit</artifactId>
+            <version>2.0.0.0</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
--- a/containers/grizzly2-http/pom.xml
+++ b/containers/grizzly2-http/pom.xml
@@ -43,9 +43,7 @@
         </dependency>
         <dependency>
             <groupId>org.hamcrest</groupId>
-            <artifactId>hamcrest-junit</artifactId>
-            <version>2.0.0.0</version>
-            <scope>test</scope>
+            <artifactId>hamcrest-library</artifactId>
         </dependency>
     </dependencies>
 
--- a/containers/netty-http/src/main/java/org/glassfish/jersey/netty/httpserver/NettyHttpContainerProvider.java
+++ b/containers/netty-http/src/main/java/org/glassfish/jersey/netty/httpserver/NettyHttpContainerProvider.java
@@ -58,7 +58,7 @@ public class NettyHttpContainerProvider implements ContainerProvider {
      * Create and start Netty server.
      *
      * @param baseUri       base uri.
-     * @param configuration Jersey configuration.
+     * @param container     Jersey container.
      * @param sslContext    Netty SSL context (can be null).
      * @param block         when {@code true}, this method will block until the server is stopped. When {@code false}, the
      *                      execution will
@@ -66,25 +66,59 @@ public class NettyHttpContainerProvider implements ContainerProvider {
      * @return Netty channel instance.
      * @throws ProcessingException when there is an issue with creating new container.
      */
-    public static Channel createServer(final URI baseUri, final ResourceConfig configuration, SslContext sslContext,
+    public static Channel createServer(final URI baseUri, final NettyHttpContainer container, SslContext sslContext,
                                        final boolean block)
             throws ProcessingException {
+
+        final ServerBootstrap serverBootstrap = createServerBootstrap(baseUri, container, sslContext);
+        return startServer(getPort(baseUri), container, serverBootstrap, block);
+    }
+
+    /**
+     * Create but not start Netty server.
+     *
+     * @param baseUri       base uri.
+     * @param container     Jersey container.
+     * @param sslContext    Netty SSL context (can be null).
+     * @return Netty bootstrap instance.
+     * @throws ProcessingException when there is an issue with creating new container.
+     */
+    static ServerBootstrap createServerBootstrap(final URI baseUri, final NettyHttpContainer container, SslContext sslContext) {
 
         // Configure the server.
         final EventLoopGroup bossGroup = new NioEventLoopGroup(1);
         final EventLoopGroup workerGroup = new NioEventLoopGroup();
-        final NettyHttpContainer container = new NettyHttpContainer(configuration);
+
+        ServerBootstrap b = new ServerBootstrap();
+        b.option(ChannelOption.SO_BACKLOG, 1024);
+        b.group(bossGroup, workerGroup)
+         .channel(NioServerSocketChannel.class)
+         .childHandler(new JerseyServerInitializer(baseUri, sslContext, container));
+
+        return b;
+    }
+
+    /**
+     * Start Netty server.
+     *
+     * @param port          IP port to listen.
+     * @param container     Jersey container.
+     * @param serverBootstrap Netty server bootstrap (i. e. prepared but unstarted server instance)
+     * @param block         when {@code true}, this method will block until the server is stopped. When {@code false}, the
+     *                      execution will
+     *                      end immediately after the server is started.
+     * @return Netty channel instance.
+     * @throws ProcessingException when there is an issue with creating new container.
+     */
+    static Channel startServer(final int port, final NettyHttpContainer container, final ServerBootstrap serverBootstrap,
+                                       final boolean block)
+            throws ProcessingException {
 
         try {
-            ServerBootstrap b = new ServerBootstrap();
-            b.option(ChannelOption.SO_BACKLOG, 1024);
-            b.group(bossGroup, workerGroup)
-             .channel(NioServerSocketChannel.class)
-             .childHandler(new JerseyServerInitializer(baseUri, sslContext, container));
+            final EventLoopGroup bossGroup = serverBootstrap.config().group();
+            final EventLoopGroup workerGroup = serverBootstrap.config().childGroup();
 
-            int port = getPort(baseUri);
-
-            Channel ch = b.bind(port).sync().channel();
+            Channel ch = serverBootstrap.bind(port).sync().channel();
 
             ch.closeFuture().addListener(new GenericFutureListener<Future<? super Void>>() {
                 @Override
@@ -105,6 +139,24 @@ public class NettyHttpContainerProvider implements ContainerProvider {
         } catch (InterruptedException e) {
             throw new ProcessingException(e);
         }
+    }
+
+    /**
+     * Create and start Netty server.
+     *
+     * @param baseUri       base uri.
+     * @param configuration Jersey configuration.
+     * @param sslContext    Netty SSL context (can be null).
+     * @param block         when {@code true}, this method will block until the server is stopped. When {@code false}, the
+     *                      execution will
+     *                      end immediately after the server is started.
+     * @return Netty channel instance.
+     * @throws ProcessingException when there is an issue with creating new container.
+     */
+    public static Channel createServer(final URI baseUri, final ResourceConfig configuration, SslContext sslContext,
+                                       final boolean block)
+            throws ProcessingException {
+        return createServer(baseUri, new NettyHttpContainer(configuration), sslContext, block);
     }
 
     /**
@@ -172,7 +224,7 @@ public class NettyHttpContainerProvider implements ContainerProvider {
         }
     }
 
-    private static int getPort(URI uri) {
+    static int getPort(URI uri) {
         if (uri.getPort() == -1) {
             if ("http".equalsIgnoreCase(uri.getScheme())) {
                 return 80;
--- a/core-server/pom.xml
+++ b/core-server/pom.xml
@@ -170,7 +170,6 @@
         <dependency>
             <groupId>jakarta.ws.rs</groupId>
             <artifactId>jakarta.ws.rs-api</artifactId>
-            <version>2.2-SNAPSHOT</version>
         </dependency>
 
         <dependency>
--- a/core-server/pom.xml
+++ b/core-server/pom.xml
@@ -170,6 +170,7 @@
         <dependency>
             <groupId>jakarta.ws.rs</groupId>
             <artifactId>jakarta.ws.rs-api</artifactId>
+            <version>2.2-SNAPSHOT</version>
         </dependency>
 
         <dependency>
@@ -227,6 +228,13 @@
             <artifactId>jersey-hk2</artifactId>
             <version>${project.version}</version>
             <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.eclipse.microprofile.config</groupId>
+            <artifactId>microprofile-config-api</artifactId>
+            <version>${microprofile.config.version}</version>
+            <scope>provided</scope>
         </dependency>
     </dependencies>
 
--- a/core-server/pom.xml
+++ b/core-server/pom.xml
@@ -229,13 +229,6 @@
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
-
-        <dependency>
-            <groupId>org.eclipse.microprofile.config</groupId>
-            <artifactId>microprofile-config-api</artifactId>
-            <version>${microprofile.config.version}</version>
-            <scope>provided</scope>
-        </dependency>
     </dependencies>
 
     <profiles>
--- a/core-server/src/main/java/org/glassfish/jersey/server/internal/RuntimeDelegateImpl.java
+++ b/core-server/src/main/java/org/glassfish/jersey/server/internal/RuntimeDelegateImpl.java
@@ -35,7 +35,6 @@ import javax.ws.rs.core.Application;
 import javax.ws.rs.JAXRS;
 import javax.ws.rs.JAXRS.Instance;
 
-import org.eclipse.microprofile.config.Config;
 import org.glassfish.jersey.internal.AbstractRuntimeDelegate;
 import org.glassfish.jersey.message.internal.MessagingBinders;
 import org.glassfish.jersey.server.ContainerFactory;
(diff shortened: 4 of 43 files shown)


mkarg:
@jansupol This is only a *very, very initial* proof-of-concept I maintain in sync with my progress on the API drafts. If nobody objects or already is planned for, I would like to extend it in the next weeks to become a *full-monty implementation* of the Java SE Bootstrap API. Is that OK for your team or are other people working on this already?

mkarg:
@jansupol Great to hear, because I already have it halfway done (runs fine with Grizzly, Jetty and JDK already). So I will finish it, add tests and keep it in sync with the ugly-named interface.

mkarg:
@jansupol Regarding [your recent comment on auto-selection of ports](https://github.com/eclipse-ee4j/jaxrs-api/pull/619#issuecomment-393857384), I have two questions: (a) Do you mean _auto-selection_ of port (= _range scan_), or do you mean _reporting the native API's default_ port (= 80 on Jersey)? If you mean range scan, can you please point me to the code in Jersey that provides a port scanner?

jansupol:
The range scan would be great, would it not? But I would prefer to leave that functionality optional on the implementation builder and not make that mandatory that by JAX-RS. I just meant a single number, which the implementation can choose. The JAXRS interface implementation should set the port independently on the container implementations, to mandate that port on any (current or future) container the user chooses. I personally favour 8080.

mkarg:
Five years ago range scan would have been a great addition to Jersey, but with docker being umbiquitus and doing range scan on his own, I think only very few people will have an _actual_ benefit of that at time of publishing Jersey 2.28.  OTOH if someone likes to spent the time to add this feature, great, then I will +1 on that PR. :-)

If we really want to return a single number instead of getting one provided, then I wonder why Jersey shall select it _globally_ instead of using any _existing_ defaults already existing in the _container providers_? I mean, why shall we return 8080 in case someone explicitly requests Grizzly and that one has 80, or explicitly requests HTTPS on Jetty and that one picks 8333? That makes no sense to me. So either the default is 8080 defined _by the JAX-RS specification_, _or_ it is whatever _the container_ decided to use. Or I do not understand the purpose here?

mkarg:
@jansupol You convinced me. I Just updated the PoC. It now does auto-selection of 80 / 443 dependend of protocol HTTP / HTTPS. More code will follow next week.

mkarg:
Supports sixth draft of JAXRS interface:
* TLS configuration.

Tested using Grizzly2, Jetty, Netty, JDK-HTTP, Simple. For SimpleFramework _mutual_ authentication (i. e. enabling SSL client authentication) is not yet implemented. I ran out of time and will look into that in the next days.

As requested, this PR now is based on the `JAXRS_2_2` branch.

jansupol:
@mkarg This is rather large commit., sorry it took me so long to get to it. I had an idea to review all at once, but it seems too difficult. I try to add comments when I got hit by something.

core-server/src/main/java/org/glassfish/jersey/server/spi/ServerProvider.java
```diff
@@ -0,0 +1,90 @@
...
+     *            The secure socket configuration to be used with HTTPS.
+     * @param sslClientAuth
+     *            Whether the server needs SSL client authentication.
+     * @return the server, otherwise {@code null} if the provider does not support
+     *         the requested {@code type}.
+     *
+     * @throws ProcessingException
+     *             if there is an error creating the server.
+     */
+    public <T extends Server> T createServer(Class<T> type, URI uri, ResourceConfig resourceConfig,
```
jansupol: This would be better  `createServer(Class<T> type, ResourceConfig resourceConfig, JAXRS.Configuration)`. The `Configuration` can contain all the necessary information for SSL, it even can contain the URI.
I would prefer `Application` over `ResourceConfig`. I know that the `*ServerFactory` classes work with `ResourceConfig`, but the `Container` classes are fine with `Application`, so a change on factories would be needed. This is could be done later.

core-server/src/main/java/org/glassfish/jersey/server/spi/Server.java
```diff
@@ -0,0 +1,13 @@
+package org.glassfish.jersey.server.spi;
+
+import java.util.concurrent.CompletionStage;
+
+public interface Server {
```
jansupol: I really miss the start() method here. I know that some servers do not have start, but the others do, and are started immediatelly. My idea is that the start could be optionally left on start() based on `JAXRS.Configuration start=false` property.

jansupol:
Please do not forget the license headers.

jansupol:
The more I think about it the more I am convinced that `Server` should extend `Container`. The `Server` implementations can hold the `Container` that they create, so that the methods `reload` are accessible. Anyway, the `Server` hold the actual server implementation instances private (for instance grizzly HttpServer), I believe there should be a getter for these instances in the `Server` implementations.

core-server/src/main/java/org/glassfish/jersey/server/spi/ServerProvider.java
mkarg: So it is ok that we directly use a JAX-RS interface inside of the Jersey backends? Good to know. :-)

core-server/src/main/java/org/glassfish/jersey/server/spi/Server.java
mkarg: Good point. But in case a backend cannot create a non-started server this would imply that Jersey simply creates a "fake" server instance. That is ok for me, if it is ok for you. :-)

mkarg:
I will have to check the actual code what it implies to implement `Container`. Thanks for pointing this out.

core-server/src/main/java/org/glassfish/jersey/server/spi/ServerProvider.java
jansupol: In this case, it would be aligned with ContainerProvider interface that is using the Application as well.

core-server/src/main/java/org/glassfish/jersey/server/spi/Server.java
jansupol: Not sure what you mean. If the backend cannot create a non-started server the start() method would throw an Unsupported Operation Exception...

core-server/src/main/java/org/glassfish/jersey/server/spi/ServerProvider.java
mkarg: I followed your proposals and pushed some additional commits reflecting them:
* `createServer` accepts `JAXRS.Configuration` instead of individual parameters.
* `ResourceConfig` was replaced by `Application`
* Factory methods have been added where needed.

core-server/src/main/java/org/glassfish/jersey/server/spi/Server.java
mkarg: I will work on non-started containers this week.

mkarg:
License headers and improved Javadocs will follow in a subsequent commit (I planned to do that at last, just wanted to get the actual code correct first).

I thought about `Server` implementing `Container` and just pushed a commit with a different design: It does not extend the container, but it provides it. I do not see that a server is a special form of a container from the birds view, but instead both are separate assets working together. So I hope you don't mind that I keep the individual concepts, but just added `Server.container()` so one can request the container via the server if needed.

Regarding you proposal to query the _native_ server instance: This was already existing, see `Server.unwrap()`.

core-server/src/main/java/org/glassfish/jersey/server/spi/Server.java
mkarg: Non-started servers done. I refactored the factories a bit, so now you can even boot unstarted instances of _all_ backend types (even Netty and Simple which previously could only boot started).

mkarg:
I will review my code this week. After that I will update Javadocs and licence headers, and squash outdated commits.

mkarg:
@jansupol I would rather slow down with more additions to the JAXRS Bootstrap. For me implementing the 7th draft worked liked a charm for Jersey and a PoC was hacked within 30 minutes in RESTeasy. So maybe it would be a good idea to _first_ merge the 7th draft (+ docs + tests) and collect ideas for a future PR _after_ the vendors are done with implementing the 7th draft? It makes decision easier having something in the hands to discuss about. Such a "post-implementation-PR" could also cover ideas like thread pools and optional support for CDI 2.0 SeContainer (RESTeasy is working an that AFAIK). Besides that, my gut instinct say that Configurable is something different to JAXRS.Configuration (one is the application, the other is the boilerplate; I'd rather not mix it).

mkarg:
@jansupol I'd like to propose the project lead should rebase branch `JAXRS_2_2` ontop `master` so it will pass Travis CI tests, which is essential to finally merge *this* PR. :-)

mkarg:
I hereby **request reviews for approval** of merging this PR, as the status of the implementation today changed from *WIP* to *complete*.

The implementation is *feature complete*, (hopefully) *production quality*, sufficiently *documented* using Javadocs, and mostly covered by *unit tests*.

mkarg:
@jansupol @asoldano @arjantijms @andymc12 Please review for approval of merge. :-)

mkarg:
Note that this PR not only contains the Java SE Bootstrap API, but ontop it also has three additional commits adding pretty closely related functionality:
* Unit tests covering the *previously existing* functionality "free port scan".
* Support for the value `0` as test container port, effectively running unit tests on a *free* port.
* Setting `0` as the *default*  when running unit tests, effectively using a free port *by default*.

I thought it is a good idea to handle all those in a single PR as it is so closely related. If you do mind, I can split the PRs. Just tell me.

mkarg:
@jansupol Can you please review this PR? I have more changes in the queue which only will work *ontop* of it. So unless you want to *intentionally* reduce my pace, it would be cool if you could approve this PR. ;-)

jansupol:
@mkarg Sorry I plan to review this as a first thing next week

core-server/src/main/java/org/glassfish/jersey/server/internal/RuntimeDelegateImpl.java
```diff
@@ -39,11 +60,126 @@ public RuntimeDelegateImpl() {
...
+    @SuppressWarnings("unchecked")
+    @Override
+    public CompletableFuture<JAXRS.Instance> bootstrap(final Application application,
+            final JAXRS.Configuration configuration) {
+        return CompletableFuture.supplyAsync(() -> {
+            final Class<Server> httpServerClass = (Class<Server>) configuration
+                    .property(ServerProperties.HTTP_SERVER_CLASS);
+
+            return new JAXRS.Instance() {
+                private final Server server = ServerFactory.createServer(httpServerClass, application, configuration);
```
jansupol: usually, `configuration.property(ServerProperties.HTTP_SERVER_CLASS)` won't be set. so httpServerClass will be `null` and  ServerFactory.createServer will cast to httpServerClass==null

containers/grizzly2-http/pom.xml
```diff
@@ -41,6 +41,12 @@
             <groupId>org.glassfish.grizzly</groupId>
             <artifactId>grizzly-http-server</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.hamcrest</groupId>
+            <artifactId>hamcrest-junit</artifactId>
+            <version>2.0.0.0</version>
```
jansupol: CQ is for 1.3. Do we need 2.0?

core-server/src/main/java/org/glassfish/jersey/server/spi/ServerProvider.java
```diff
@@ -0,0 +1,87 @@
...
+ * <p>
+ * An implementation can identify itself by placing a Java service provider
+ * configuration file (if not already present) -
+ * {@code org.glassfish.jersey.server.spi.ServerProvider} - in the resource
+ * directory {@code META-INF/services}, and adding the fully qualified
+ * service-provider-class of the implementation in the file.
+ * </p>
+ *
+ * @author Markus KARG (markus@headcrashing.eu)
+ * @since 2.28
```
jansupol: 2.28 is EE4J_8

core-server/pom.xml
```diff
@@ -169,6 +169,7 @@
         <dependency>
             <groupId>javax.ws.rs</groupId>
             <artifactId>javax.ws.rs-api</artifactId>
+            <version>2.2-SNAPSHOT</version>
```
jansupol: This should be in pom.xml rather than in core-server/pom.xml?

containers/netty-http/src/main/java/org/glassfish/jersey/netty/httpserver/NettyHttpContainerProvider.java
```diff
@@ -58,33 +58,67 @@
...
+     * @param container     Jersey container.
      * @param sslContext    Netty SSL context (can be null).
      * @param block         when {@code true}, this method will block until the server is stopped. When {@code false}, the
      *                      execution will
      *                      end immediately after the server is started.
      * @return Netty channel instance.
      * @throws ProcessingException when there is an issue with creating new container.
      */
-    public static Channel createServer(final URI baseUri, final ResourceConfig configuration, SslContext sslContext,
+    public static Channel createServer(final URI baseUri, final NettyHttpContainer container, SslContext sslContext,
```
jansupol: You are making backwards incompatible change. Why?

core-server/src/main/java/org/glassfish/jersey/server/spi/ServerProvider.java
mkarg: What is the planned Jersey version number for adopting JAX-RS 2.2?

core-server/src/main/java/org/glassfish/jersey/server/internal/RuntimeDelegateImpl.java
mkarg: This won't happen, because the property has a default set *by the config builder*, see:
https://github.com/eclipse-ee4j/jersey/blob/fa9d0fe2d39be5e475035ac626a1bf19e0290753/core-server/src/main/java/org/glassfish/jersey/server/internal/RuntimeDelegateImpl.java#L94

containers/grizzly2-http/pom.xml
mkarg: Cannot find a version 1.3 for the Maven coordinates `hamcrest-junit` on [Maven Central] (https://search.maven.org/search?q=g:org.hamcrest%20AND%20a:hamcrest-junit&core=gav). It only lists `1.0.0.0-RC1`, `1.0.0.0`, `1.0.0.1` and `2.0.0.0`. So which Maven coordinates shall I use instead?

core-server/pom.xml
mkarg: Right! In fact it already exists in pom.xml so it simply is useless in core-server/pom.xml.

containers/netty-http/src/main/java/org/glassfish/jersey/netty/httpserver/NettyHttpContainerProvider.java
mkarg: Actually I didn't. I *added an additional method*, but the diff viewer displays this in a rather confusing way, so you got misleaded.

The original signature still exists see, line 156:
https://github.com/eclipse-ee4j/jersey/blob/fa9d0fe2d39be5e475035ac626a1bf19e0290753/containers/netty-http/src/main/java/org/glassfish/jersey/netty/httpserver/NettyHttpContainerProvider.java#L156

containers/grizzly2-http/pom.xml
jansupol: Why you need hamcrest-junit? Only `org.hamcrest.CoreMatchers` and `org.hamcrest.Matchers` are used, neither is sourced in hamcrest-junit, We use hamcrest-library, which should cover the used classes

core-server/src/main/java/org/glassfish/jersey/server/spi/ServerProvider.java
mkarg: "ping"? ;-)

jansupol: It actually depends on when JAX-RS 2.2 is to be released. 2.28 should be released this year. 2.29 contains a bunch of fixes that we would like to release, too. Should JAX-RS 2.2 be released around 2.29 time, 2.30 sounds just right

mkarg: Understood. Sad but true, the release date of JAX-RS 2.2 in turn is mostly dependend on the missing contract between Oracle and Eclipse Foundation. So all we can do is to simply keep this PR open, and as soon as the contract is there I will strive for releasing JAX-RS 2.2 and then update the `@since` markers.

mkarg: As JAX-RS 2.1.4 was published two weeks ago, but 2.28 is not published still, I'd like to ask again, how to go on with this topic? I would like to speed up the pace of JAX-RS's new features now that 2.1.4 seems to be the final release.

mkarg:
@jansupol I could rebase this ontop of the recently updated branch `JAXRS_2_2` and force-push it then. Would that be beneficial for you?

jansupol:
@mkarg Please go ahead. If we should ever finish this review, it would be good to have it buildable.

mkarg:
@jansupol Rebased Java SE Bootstrap API ontop of branch JAXRS_2_2. I kindly request to continue the review. :-)

mkarg:
@jansupol What is holding you back from merging this? Anything you want me to change?

core-server/src/main/java/org/glassfish/jersey/server/spi/ServerProvider.java
jansupol: We have a [plan](https://github.com/eclipse-ee4j/jersey/wiki/Road-Map) to support this bootstrap in 2.30, sometimes in September. Does it sound good for you?

mkarg: The version "Jersey 2.30" and "having a public plan" sounds really great, but waiting until September makes me cry. ;-)

core-server/src/main/java/org/glassfish/jersey/server/spi/ServerProvider.java
mkarg: As you didn't respond for months, I assume you accept these changes, so I hereby mark this discussion as resolved. :-)

core-server/src/main/java/org/glassfish/jersey/server/spi/ServerProvider.java
mkarg: I pushed another commit replacing 2.29 by 2.30, so this discussion is resolved. :-)

mkarg:
> Do not create integration with microprofile. We can create the integration in some extension module, but not as a default in jaxrs-ri. The provided scope does not solve this. Let's focus on the bootstrap iteslf for now, we can deal with the mp config later. In a separate module.

All requested changes are made. I would be more than happy to see this review being finished and the changes finally merged into the `JAXRS_2_2' branch. Otherwise please tell me what more you want me to change. :-)

mkarg:
@arjantijms @asoldano @andymc12 Review please. :-)

jansupol:
@mkarg It looks good. Thank you.

PR #3845, opened 2018-06-04
addAnnotatedType now provide ID to prevent warning

jGauravGupta:
Author: arjantijms <arjan.tijms@gmail.com>
Signed-off-by: jGauravGupta <gaurav.gupta.jc@gmail.com>

### DIFF
--- a/connectors/jetty-connector/src/test/java/org/glassfish/jersey/jetty/connector/AsyncTest.java
+++ b/connectors/jetty-connector/src/test/java/org/glassfish/jersey/jetty/connector/AsyncTest.java
@@ -110,29 +110,32 @@ public class AsyncTest extends JerseyTest {
                 }
             });
             asyncResponse.setTimeout(1, TimeUnit.SECONDS);
+            asyncResponse.resume(Response.status(Response.Status.SERVICE_UNAVAILABLE)
+                    .entity("Operation time out.").build());
 
-            new Thread(new Runnable() {
-
-                @Override
-                public void run() {
-                    String result = veryExpensiveOperation();
-                    asyncResponse.resume(result);
-                }
-
-                private String veryExpensiveOperation() {
-                    // very expensive operation that typically finishes within 1 second but can take up to 5 seconds,
-                    // simulated using sleep()
-                    try {
-                        Thread.sleep(5 * OPERATION_DURATION);
-                        return "DONE";
-                    } catch (InterruptedException e) {
-                        Thread.currentThread().interrupt();
-                        return "INTERRUPTED";
-                    } finally {
-                        LOGGER.info("Async long-running get with timeout finished on thread " + Thread.currentThread().getName());
-                    }
-                }
-            }).start();
+//            new Thread(new Runnable() {
+//
+//                @Override
+//                public void run() {
+//                    String result = veryExpensiveOperation();
+//                    asyncResponse.resume(result);
+//                }
+//
+//                private String veryExpensiveOperation() {
+//                    // very expensive operation that typically finishes within 1 second but can take up to 5 seconds,
+//                    // simulated using sleep()
+//                    try {
+//                        Thread.sleep(5 * OPERATION_DURATION);
+//                        return "DONE";
+//                    } catch (InterruptedException e) {
+//                        Thread.currentThread().interrupt();
+//                        return "INTERRUPTED";
+//                    } finally {
+//                        LOGGER.info("Async long-running get with timeout finished on thread " +
+//            Thread.currentThread().getName());
+//                    }
+//                }
+//            }).start();
         }
 
     }
--- a/connectors/jetty-connector/src/test/java/org/glassfish/jersey/jetty/connector/AsyncTest.java
+++ b/connectors/jetty-connector/src/test/java/org/glassfish/jersey/jetty/connector/AsyncTest.java
@@ -113,29 +113,28 @@ public class AsyncTest extends JerseyTest {
             asyncResponse.resume(Response.status(Response.Status.SERVICE_UNAVAILABLE)
                     .entity("Operation time out.").build());
 
-//            new Thread(new Runnable() {
-//
-//                @Override
-//                public void run() {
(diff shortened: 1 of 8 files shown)


ext/cdi/jersey-cdi1x-servlet/src/main/java/org/glassfish/jersey/ext/cdi1x/servlet/internal/CdiExternalRequestScopeExtension.java
```diff
@@ -57,7 +57,7 @@
      */
     private void beforeBeanDiscovery(@Observes BeforeBeanDiscovery beforeBeanDiscoveryEvent, final BeanManager beanManager) {
         requestScopeType = beanManager.createAnnotatedType(CdiExternalRequestScope.class);
-        beforeBeanDiscoveryEvent.addAnnotatedType(requestScopeType);
+        beforeBeanDiscoveryEvent.addAnnotatedType(requestScopeType, "Jersey " + CdiExternalRequestScope.class.getName());
```
japod: IIUC this is the only relevant change. Is there any test available for it?

connectors/jetty-connector/src/test/java/org/glassfish/jersey/jetty/connector/AsyncTest.java
```diff
@@ -110,29 +110,32 @@ public void handleTimeout(AsyncResponse asyncResponse) {
...
+//                        return "DONE";
+//                    } catch (InterruptedException e) {
+//                        Thread.currentThread().interrupt();
+//                        return "INTERRUPTED";
+//                    } finally {
+//                        LOGGER.info("Async long-running get with timeout finished on thread " +
+//            Thread.currentThread().getName());
+//                    }
+//                }
+//            }).start();
```
japod: If the above code is not important any more, you should remove it. Could you please provide some explanation why you think it should be gone?

arjantijms:
>Seems to me there is a lot of refactoring, some test code gone and only a one line update that is related to the PR title itself. But maybe i am missing something?

Yes, there's one observation missing and that's the addition of the `ProcessJAXRSAnnotatedTypes` and `ProcessAllAnnotatedTypes` CDI extensions. These have been split out from the existing code, hence some refactoring there.

These allow an integration container (such as GlassFish, Payara, Jboss), to first scan if an `Hk2CustomBoundTypesProvider` is present, and depending on that install either `ProcessAllAnnotatedTypes` or `ProcessJAXRSAnnotatedTypes`. 

The other functional change is indeed the `beforeBeanDiscoveryEvent.addAnnotatedType` one.  The old method (without the ID) is deprecated since CDI 1.1, see:

```java
    /**
     * <p>
     * Adds a given {@link javax.enterprise.inject.spi.AnnotatedType} to the set of types which will be scanned during bean
     * discovery.
     * </p>
     * 
     * <p>
     * This method is deprecated from CDI 1.1 and {@link #addAnnotatedType(AnnotatedType, String)} should be used instead.
     * </p>
     * 
     * @throws IllegalStateException if called outside of the observer method invocation
     * @param type The {@link javax.enterprise.inject.spi.AnnotatedType} to add for later scanning
     */
    void addAnnotatedType(AnnotatedType<?> type);
```

There's no specific test for that. `void addAnnotatedType(AnnotatedType<?> type, String id);` fully replaces the 1 parameter one. If the potentially existing test testing the 1 parameter version still passes, the replacement is successful.

connectors/jetty-connector/src/test/java/org/glassfish/jersey/jetty/connector/AsyncTest.java
arjantijms: It's some time ago I did the original commenting out. This one probably should be un-commented out again. The reason I commented it out in the original commit is that it's a very fragile test. It would fail a lot of times even before any changes were made to the code.

So to make testing easier I commented it out, but it's unrelated to this PR.

jansupol:
@arjantijms This pull request should not have been merged. First of all, as you agreed with @japod the warning prevention a single line change, but there are other changes splitting the Extensions into two. That has nothing to do with the warning and should have been in a separate commit. The file header that is used in PR contains CDDL. The copyright header for portions copyright should not be in a comment _// Portions Copyright [2018] Payara Foundation and/or affiliates_ but in the header:

>  /*
>   * Copyright (c) 2015, 2018 Oracle and/or its affiliates. All rights reserved.
>   * Copyright (c) 2018  Payara Foundation and/or its affiliates. All rights reserved.

There is copyright plugin v 1.49 that passes with this copyright. We have PR #3855 for this.

Overall, we appreciate the help from Payara, but  PRs should be submitted with better care. or if without care, not approved and merged by the same person while the discussion still goes on.

Can you please submit a new PR with license and copyright header fix?

arjantijms:
@jansupol Sorry for not responding earlier. I somehow completely misses your comment. My apologies for that.

Indeed, the header is wrong. The PR was picked from a revision it seems, or, the changes that I initially did had been applied to the wrong branch. I absolutely agree that PRs should be submitted with better care than this. If you look at our history for other commits, be it MP or Mojarra, you'd see this is normally the case.

The splitting however IS about warning prevention. It's split, since without it, Weld *will* complain about scanning potentially too much types. If I'm not mistaken your original comment there mentioned that already. By having a version that scans everything as default with the option to replace it with a version that does only scan the JAX-RS types (again, as per the original comment) these warnings can be prevented.

Indeed, it could have been 2 PRs, since it are two different kinds of warnings (not providing the recommended ID vs scanning too many types), but you could also say a single PR makes some sense as its about warnings altogether.

Yes, I will supply a new PR with the license and copyright fix.

My humble apologies for overlooking the wrong license/copyright. This is indeed my fault and there's no excuse for that.

jansupol:
@arjantijms Thank you for supplying the fix PR. We try to better incorporate the copyright check job to PRs to warn about the license/copyright issues.

Issue #3846, opened 2018-06-04
OsgiRegistry attempts to load classes from comment lines


jesse-gallagher:
https://github.com/eclipse-ee4j/jersey/blob/12a057315292a624ceb31b58bd04eec3cb19a819/core-common/src/main/java/org/glassfish/jersey/internal/OsgiRegistry.java#L209

I've run into a problem in my (admittedly bizarre) environment where I'm attempting to bind Jersey with Weld in an OSGi environment that's a bit too old to use the non-SE Weld container. Setting aside the ugliness of my project, I found that the `OsgiLoader` class attempts to load lines from factory files that start with "#", such as this one from the cdi2-se package:

https://github.com/eclipse-ee4j/jersey/blob/12a057315292a624ceb31b58bd04eec3cb19a819/inject/cdi2-se/src/main/resources/META-INF/services/org.glassfish.jersey.internal.inject.InjectionManagerFactory

PR #3847, opened 2018-06-04
Issue #3846: Skip comment lines in service files in OSGi

jesse-gallagher:
This avoids trouble with service files like https://github.com/eclipse-ee4j/jersey/blob/12a057315292a624ceb31b58bd04eec3cb19a819/inject/cdi2-se/src/main/resources/META-INF/services/org.glassfish.jersey.internal.inject.InjectionManagerFactory


jansupol:
This pull request should be put to EE4J_8 branch, too.
Could you please put the fix back there, too?

PR #3852, opened 2018-06-08
New branch for JAX-RS 2.2 development

spericas:
Includes changes related to dropping default runtime delegate and implementation of new SSE methods. Depends on snapshot version of JAX-RS 2.2 API which needs to be available in your local repo.

Signed-off-by: Santiago Pericas-Geertsen <Santiago.PericasGeertsen@oracle.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,9 @@
+sudo: required
+
 language: java
+
+services:
+  - docker
 
 jdk:
   - oraclejdk8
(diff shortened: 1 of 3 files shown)


core-common/src/main/java/org/glassfish/jersey/internal/jsr166/SubmissionPublisher.java
```diff
@@ -581,29 +581,45 @@ final int doOffer(long nanos, T item,
...
+     * onComplete} signals to current subscribers when {@code cascading} is
+     * {@code true}. If {@code cascading} is {@code false}, then only closes this
+     * publisher and disallows subsequent attempts to publish. Upon return, even if
+     * {@link Flow.Subscriber#onComplete() onComplete} is called, this method does
+     * <em>NOT</em> guarantee that all subscribers have yet completed.
+     *
+     * @param cascading Controls closing of registered subscribers.
      */
-    public void close() {
+    public void close(boolean cascading) {
```
jansupol: This is jdk 9 SubmissionPublisher. It does not have close(boolean). Should we add it?

PR #3858, opened 2018-06-17
Allow the same Jersey client to be used for both authenticated and non-authenticated URLs

agherardi:
This is a clone of https://github.com/jersey/jersey/pull/3563


core-client/src/main/java/org/glassfish/jersey/client/authentication/BasicAuthenticator.java
```diff
@@ -68,15 +68,13 @@ private String calculateAuthentication(HttpAuthenticationFilter.Credentials cred
...
      * @param request Request context.
-     * @throws RequestAuthenticationException in case that basic credentials missing or are in invalid format
      */
-    public void filterRequest(ClientRequestContext request) throws RequestAuthenticationException {
+    public void filterRequest(ClientRequestContext request) {
         HttpAuthenticationFilter.Credentials credentials = HttpAuthenticationFilter.getCredentials(request,
                 defaultCredentials, HttpAuthenticationFilter.Type.BASIC);
-        if (credentials == null) {
-            throw new RequestAuthenticationException(LocalizationMessages.AUTHENTICATION_CREDENTIALS_MISSING_BASIC());
+        if (credentials != null) {
```
senivam: I suppose this brakes the logic above:
`final String authenticate = response.getHeaders().getFirst(HttpHeaders.WWW_AUTHENTICATE);
        if (authenticate != null && authenticate.trim().toUpperCase().startsWith("BASIC"))`
because there is a check where credentials are really required. So if they are checked to be really required, they  shall be provided.

agherardi: Are you referring to the logic in method BasicAuthenticator.filterResponseAndAuthenticate?

If the server returns a WWW-Authenticate Basic challenge, filterResponseAndAuthenticate checks if the application provides credentials. If the application provides credentials, filterResponseAndAuthenticate repeats the request adding an Authorization header containing the credentials. If the application does NOT provide credentials, filterResponseAndAuthenticate throws a ResponseAuthenticationException. So either way, no logic should be OK.

The purpose of this change is to avoid pre-emptively, and incorrectly, throwing a RequestAuthenticationException when the server requires no credentials and the application provides no credentials.

senivam: OK, this sounds reasonable. Could you please add Logger message instead of removed exception of the Level.FINE that no credentials provided and empty credentials are being sent?

agherardi: Sure. Done.

agherardi:
Hi - Any updates?

PR #3860, opened 2018-06-17
Patch for client connection leak when using digest authentication

agherardi:
This is a clone of https://github.com/jersey/jersey/pull/3751

### DIFF
--- a/core-client/src/main/java/org/glassfish/jersey/client/authentication/HttpAuthenticationFilter.java
+++ b/core-client/src/main/java/org/glassfish/jersey/client/authentication/HttpAuthenticationFilter.java
@@ -25,9 +25,9 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 
+import javax.annotation.Priority;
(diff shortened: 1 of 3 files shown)


core-client/src/main/java/org/glassfish/jersey/client/authentication/HttpAuthenticationFilter.java
```diff
@@ -419,4 +431,23 @@ static Credentials getCredentials(ClientRequestContext request, Credentials defa
             return specificCredentials != null ? specificCredentials : defaultCredentials;
         }
     }
+
+    private static void discardInputAndClose(InputStream is) {
```
jansupol: This discardInputAndClose is once again in HttpAuthenticationFilter in PR3859. Can you put it to the AuthenticationUtil as discussed in PR 3862?

agherardi: Yes. Done.

PR #3862, opened 2018-06-17
Exclude query parameters when caching authentication info

agherardi:
This is a clone of https://github.com/jersey/jersey/pull/3760

### DIFF
--- a/core-client/src/main/java/org/glassfish/jersey/client/authentication/HttpAuthenticationFilter.java
+++ b/core-client/src/main/java/org/glassfish/jersey/client/authentication/HttpAuthenticationFilter.java
@@ -25,9 +25,9 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 
+import javax.annotation.Priority;
 import javax.ws.rs.Priorities;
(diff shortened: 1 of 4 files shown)


jansupol:
Fixes #3629

core-client/src/main/java/org/glassfish/jersey/client/authentication/HttpAuthenticationFilter.java
```diff
@@ -247,7 +246,22 @@ public void filter(ClientRequestContext request, ClientResponseContext response)
     }
 
     private String getCacheKey(ClientRequestContext request) {
-        return request.getUri().toString() + ":" + request.getMethod();
+        URI requestUri = request.getUri();
```
jansupol: This method duplicates code from DigestAuthenticator#getCacheKey. Does it make sense to put the code to a new static method to a new utility AuthenticationUtil class in the package?

agherardi: Yes, asbolutely. Done.

PR #3870, opened 2018-06-27
SSE tests fixed

senivam:
Fixes SSE tests by suspending main thread until required events are completed (ready for assertions) 

Reverts back 2 build profiles for better build performance.


senivam:
better clone of #3867 which is deleted

PR #3884, opened 2018-07-14
Travis uses Eclipse Repo to get latest SNAPSHOTs

mkarg:
To be able to continously build support for JAX-RS API's new features into Jersey, accessing the Eclipse Foundation's Repository is essential, because it is the central place the EF's admins want nightly builds to be stored.

This PR adds the repository using the profile `eclipse_repo`, which is defined in the POM and explicitly enabled by the Travis CI driver script.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1204,6 +1204,20 @@
                 <skip.tests>true</skip.tests>
             </properties>
         </profile>
+	<profile>
+            <!-- Use -Peclipse_repo to use SNAPSHOTs stored in Eclipse's Nexus instance ("Nightly Builds") -->
+            <id>eclipse_repo</id>
+	    <repositories>
+                <repository>
+                    <snapshots>
+                        <enabled>true</enabled>
+                    </snapshots>
+                    <id>repo.jaxrs-api.eclipse.org</id>
+                    <name>JAX-RS API Repository - Snapshots</name>
+                    <url>https://repo.eclipse.org/content/repositories/jax-rs-api-snapshots</url>
+                </repository>
+            </repositories>
+	</profile>
     </profiles>
 
     <reporting>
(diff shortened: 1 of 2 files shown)


mkarg:
(Just in case you didn't know, but I am only a JAX-RS committer, not a Jersey committer. So one of you two needs to merge this -- I simply cannot).

mkarg:
@arjantijms Thanks. BTW, don't you have a "two-weeks-review-rule" in Jersey like we have one in JAX-RS?

arjantijms:
@mkarg not that I know of. Definitely not having such rule in Mojarra (impl) or Soteria. That would slow development down too much. If there are issues it can always be reverted or amended. But would be good to establish here that there's no such rule and/or what rule we do abide by then.

spericas:
@mkarg Which is the commit that you are referring to? When @jansupol and I looked at some issues last week, we saw a problem related to the new JAXRS interface that you checked in.

mkarg:
@spericas Thanks for picking this up, but it seems this got fixed meanwhile. FYI: I double checked the [JAX-RS Travis CI Dashboard](https://travis-ci.org/eclipse-ee4j/jersey/branches) and noticed a single red cross in the right side (the one which says #69 when hovering over it). Clicking on the red cross opened the [Build Result #69](https://travis-ci.org/eclipse-ee4j/jersey/builds/394071068). Right in the middle of that it clicked on [Compare 45c6ac1..3db9077]() to see [what commits](https://github.com/eclipse-ee4j/jersey/compare/45c6ac10fe06...3db90778ebe5) caused the build to fail. Apparently it was your first change to [support JAX-RS 2.2](https://github.com/eclipse-ee4j/jersey/commit/10b22cbca153737d00c0b0e02f302712a00565cc). It seems it became auto-fixed by @arjantijms merging my PR.

PR #3888, opened 2018-07-19
Jersey creates multiple provider instances if a class implements more than one

senivam:
Fixes issue #3796

### DIFF
--- a/tests/integration/jersey-3796/src/test/java/org/glassfish/jersey/tests/integration/jersey3796/Jersey3796ITCase.java
+++ b/tests/integration/jersey-3796/src/test/java/org/glassfish/jersey/tests/integration/jersey3796/Jersey3796ITCase.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright (c) 2018 Oracle and/or its affiliates. All rights reserved.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0, which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception, which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ */
+
+package org.glassfish.jersey.tests.integration.jersey3796;
+
+import org.glassfish.jersey.test.JerseyTest;
+import org.glassfish.jersey.test.external.ExternalTestContainerFactory;
+import org.glassfish.jersey.test.grizzly.GrizzlyTestContainerFactory;
+import org.glassfish.jersey.test.grizzly.GrizzlyWebTestContainerFactory;
+import org.glassfish.jersey.test.spi.TestContainerException;
+import org.glassfish.jersey.test.spi.TestContainerFactory;
+import org.junit.Assert;
+import org.junit.Test;
+
+import javax.ws.rs.core.Application;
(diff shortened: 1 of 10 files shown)


tests/integration/jersey-3796/src/test/java/org/glassfish/jersey/tests/integration/jersey3796/Jersey3796ITCase.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    @Test
+    public void testSameInstanceForProviders() {
+        final Map response = target("/myresource").request(MediaType.APPLICATION_JSON_TYPE).get(Map.class);
+        //Map shall not be null
+        Assert.assertNotNull(response);
+        //Map shall contain exactly three elements
+        Assert.assertEquals(3, response.size());
+        //Map shall contain ony keys Feature, Request and Response
```
mkarg: `ony`? ;-)

tests/integration/jersey-3796/src/test/java/org/glassfish/jersey/tests/integration/jersey3796/Jersey3796ITCase.java
```diff
@@ -0,0 +1,61 @@
...
+        //Values of that keys shall be equals.
+        //Equality of all values indicates the class is only one per all tested providers
+        Assert.assertEquals(response.get("Feature"), response.get("Request"));
```
mkarg: Shouldn't it be really the same instance instead of simply equals?

senivam: the purpose here is to check if 1 class is initialized only once (thus it's the only instance). To check that a map of strings is used. It is initialized along with a class and each provider method puts a record inside. Class implements 3 interfaces each interface has 1 method, so there are 3 records at the end. All 3 records shall be in 1 map. This indicates the class is initialized only once. If map contains less instances (and/or NPE exception is thrown when a method tries to put a record into the map) that indicates the contract is broken and class is initialized more than once.

mkarg:
@senivam Thanks a lot for fixing this! :-)

Conrad-T-Pino:
I'm suggesting not fixed in Jersey 2.35 nor 2.44. I'm logging in no argument constructor and see four (4) log entries; one for each implemented Contract.class for the first request but not thereafter.

Ideally I wanted singleton state as class field in singleton instance. Logs suggest I must use static field to circumvent multiple instances.

```java
public class ClientCapture implements 
	ClientRequestFilter, ClientResponseFilter, ReaderInterceptor, WriterInterceptor {
	public ClientCapture() {
		super();
		logger.info("ClientCapture()");
	}
	...
}
``` 
```shell
25-Aug-2024 15:02:24.978 INFO [pool-6-thread-1] tmms.rest.ClientCapture.<init> ClientCapture()
25-Aug-2024 15:02:24.978 INFO [pool-6-thread-1] tmms.rest.ClientCapture.<init> ClientCapture()
25-Aug-2024 15:02:24.982 INFO [pool-6-thread-1] tmms.rest.ClientCapture.<init> ClientCapture()
25-Aug-2024 15:02:24.983 INFO [pool-6-thread-1] tmms.rest.ClientCapture.<init> ClientCapture()
25-Aug-2024 15:02:25.078 INFO [pool-6-thread-1] tmms.rest.ClientCapture.filter (ClientRequestContext)
25-Aug-2024 15:02:25.090 INFO [pool-6-thread-1] tmms.rest.ClientCapture.aroundWriteTo (WriterInterceptorContext)
25-Aug-2024 15:02:25.785 INFO [pool-6-thread-1] tmms.rest.ClientCapture.filter (ClientRequestContext, ClientResponseContext)
25-Aug-2024 15:02:25.790 INFO [pool-6-thread-1] tmms.rest.ClientCapture.aroundReadFrom (ReaderInterceptorContext)
```

Issue #3872, opened 2018-06-28
JerseySseBroadcaster#broadcast() returns null


jansupol:
Let it return expected CompletionStage.

PR #3897, opened 2018-07-25
(master branch) fix for issue 3872

senivam:
fix for issue #3872 

JerseySseBroadcaster#broadcast() returns null

### DIFF
--- a/media/sse/src/main/java/org/glassfish/jersey/media/sse/internal/JerseySseBroadcaster.java
+++ b/media/sse/src/main/java/org/glassfish/jersey/media/sse/internal/JerseySseBroadcaster.java
@@ -16,6 +16,7 @@
 
 package org.glassfish.jersey.media.sse.internal;
 
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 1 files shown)


media/sse/src/main/java/org/glassfish/jersey/media/sse/internal/JerseySseBroadcaster.java
```diff
@@ -102,10 +103,9 @@ public void onClose(Consumer<SseEventSink> onClose) {
         if (event == null) {
             throw new IllegalArgumentException(LocalizationMessages.PARAM_NULL("event"));
         }
-        publish(event);
 
         // TODO JAX-RS 2.1
```
Cousjava: Should this TODO be removed?

jansupol: @Cousjava We do not feel strongly about the TODO. Ideally, the `CompletableFuture#isDone()` would return true AFTER the messages are broadcasted (which possibly was the Spec intention), but it does not happen now and it is hard to accomplish using the JSR166 SubmissionPublisher. So we tend to leave it there as a mark of what have been made better.  The PR is mainly not to return null. Do you feel the TODO should be removed?

jansupol:
ok, we remove the TODO

Issue #3879, opened 2018-07-11
SSLConfigurator requires java.util.PropertyPermission for * with read and write


shaneargo:
Hi,

We are using Jersey Client in an environment which enforces certain Java Security Manager restrictions, which we have no control over, and we have noticed that SslConfigurator requires java.util.PropertyPermission for all properties with read and write.

This class doesn't appear to actually require writing any system properties or the ability to read most.

This is because, when constructing an SslConfigurator using system properties, every system property is loaded (using `PropertiesHelper.getSystemProperties()` which calls `System.getProperties()`), and then the individual properties are selected from the map:
https://github.com/eclipse-ee4j/jersey/blob/12a057315292a624ceb31b58bd04eec3cb19a819/core-common/src/main/java/org/glassfish/jersey/SslConfigurator.java#L310

If this refactored so that `PropertiesHelper.getSystemProperty()` (which calls `System.getProperty()`) instead, then we could grant read-only permissions to the specific system properties.

I'd be happy to do this refactor and send a pull request, I just wanted to run it by you before doing so.

Cheers,
Shane.

jansupol:
Hi Shane,
This sounds reasonable. Please submit the PR.
Thanks,
Jan

PR #3904, opened 2018-08-02
Modify the SSLConfigurator to get specific system properties.

shaneargo:
Resolves issue #3879.

Instead of reading all system properties in the SSLConfigurator, it will now
read only the specific properties required. This ensures that in an environment
utilising the Java Security Manager only the privileges for those system
properties are required.

### DIFF
--- a/core-common/src/main/java/org/glassfish/jersey/SslConfigurator.java
+++ b/core-common/src/main/java/org/glassfish/jersey/SslConfigurator.java
@@ -307,7 +307,7 @@ public final class SslConfigurator {
 
     private SslConfigurator(boolean readSystemProperties) {
         if (readSystemProperties) {
-            retrieve(AccessController.doPrivileged(PropertiesHelper.getSystemProperties()));
+            retrieve();
         }
     }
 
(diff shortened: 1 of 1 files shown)


core-common/src/main/java/org/glassfish/jersey/SslConfigurator.java
```diff
@@ -780,35 +780,42 @@ public SSLContext createSSLContext() {
     }
 
     /**
-     * Retrieve the SSL context configuration from the supplied properties.
+     * Retrieve the SSL context configuration from the system properties.
      *
-     * @param props properties containing the SSL context configuration.
      * @return updated SSL configurator instance.
      */
-    public SslConfigurator retrieve(Properties props) {
```
jansupol: This method is public. It should remain in the Class, should the users use it.

jansupol: Its fine to add just a new method.

jansupol:
The commit does not have sign-off, nor you have singed the OCA. Do you want to go through the process, or do you want us to create the PR on our own from scratch?

shaneargo:
Thanks @jansupol.

Sorry for my oversight with that public method. I'll fix it up.

I read the CONTRIBUTING.md file before-hand and signed the ECA? Is this distinct from the OCA?

If I add the 'Signed-off-by' field to the next commit, will this suffice, or do I need to kill this commit and go again?

If you'd prefer do this another way, please let me know. I am happy with whatever.

jansupol:
Unfortunatelly Eclipse Contribution Agreement is different from OCA, and OCA no longer applies. Both ECA and sign-off is needed. The ECA is easier to be signed, once you have the Eclipse account, you get access to the ECA form from your account. To me, adding sign-off to the commit should be sufficient, though the ip-validation check sometimes seems to think otherwise.

shaneargo:
I am confused. I am saying that I have signed the ECA, but you asked me to sign the OCA. What ever the case, I’ve signed the ECA, if you want me to sign something else, let me know. 

I’ll fix up my commit and add the sign off as soon as I can.

jansupol:
Sorry I was not clear. I thought you already have OCA. Unfortunately, people who already have OCA need a new ECA. OCA is of no use in Jersey now, you do not need to sign it. Fortunately, it is easier to get ECA than it used to be to get OCA. So now the Eclipse process requires two things: ECA and sign-off.

shaneargo:
I am sorry, I am partly to blame for the confusion around ECA. I had already signed it, but using a different email address.

I've also made the changes we discussed. 

Please let me know if there is anything else you'd like me to do.

jansupol:
The ip-validation does not seem to approve you. As I understand, you signed the commit with different email address than the email used for ECA. It is possible to add other email addresses to your eclipse account, maybe it would help. Can you give it a try? If that does not work, could you sign  off the commit with the email address used for ECA?

shaneargo:
I have already signed the ECA using the address I use for Github and signed the commit with.

Is it because the first commit was missing the "Signed-off-by" with it being added in the last commit?

jansupol:
I checked, Eclipse says: There is a valid ECA on file for shane@alltheducks.com.
It could complain because  the first two commits do not have sign-off. The tool is available at https://accounts.eclipse.org/users/<user>/eca after login. I think we are safe to merge this

Issue #3905, opened 2018-08-02
Don't log warnings when only checking if CDI bean is also JAX-RS resource


arjantijms:
Jersey checks whether CDI beans are also JAX-RS resources, and if so modifies their injection points. For this check it shouldn't validate a CDI bean according to JAX-RS rules, if it's not a JAX-RS resource.

Practically, this often results in useless warnings, but it can even result in exceptions preventing deployments.

spericas:
What are the details of the validation step done by Jersey?

PR #3906, opened 2018-08-02
#3905 Fix warnings when checking CDI bean is JAX-RS resource

arjantijms:
#3905 Check with validation disabled first. Only if it's indeed a JAX-RS bean, perform validation.

Signed-off-by: arjantijms <arjan.tijms@gmail.com>

### DIFF
--- a/core-server/src/main/java/org/glassfish/jersey/server/model/IntrospectionModeller.java
+++ b/core-server/src/main/java/org/glassfish/jersey/server/model/IntrospectionModeller.java
@@ -183,8 +183,10 @@ final class IntrospectionModeller {
                     method.getGenericParameterTypes()[0],
                     method.getAnnotations());
             if (null != p) {
-                ResourceMethodValidator.validateParameter(p, method.getMethod(), method.getMethod().toGenericString(), "1",
+                if (!disableValidation) {
+                    ResourceMethodValidator.validateParameter(p, method.getMethod(), method.getMethod().toGenericString(), "1",
                         InvocableValidator.isSingleton(handlerClass));
+                }
 
                 // we do not inject entity parameters into class instance fields and properties.
                 if (p.getSource() != Parameter.Source.ENTITY) {
@@ -208,8 +210,11 @@ final class IntrospectionModeller {
                         field.getGenericType(),
                         field.getAnnotations());
                 if (null != p) {
-                    ResourceMethodValidator.validateParameter(p, field, field.toGenericString(), field.getName(),
+                    if (!disableValidation) {
+                        ResourceMethodValidator.validateParameter(p, field, field.toGenericString(), field.getName(),
                             isInSingleton);
+                    }
+
                     // we do not inject entity and unknown parameters into class instance fields and properties.
                     if (p.getSource() != Parameter.Source.ENTITY) {
                         injectableParameters.add(p);
--- a/ext/cdi/jersey-cdi1x/src/main/java/org/glassfish/jersey/ext/cdi1x/internal/CdiComponentProvider.java
+++ b/ext/cdi/jersey-cdi1x/src/main/java/org/glassfish/jersey/ext/cdi1x/internal/CdiComponentProvider.java
@@ -144,7 +144,10 @@ public class CdiComponentProvider implements ComponentProvider, Extension {
         }
     });
 
-    private final Cache<Class<?>, Boolean> jaxRsResourceCache = new Cache<>(clazz -> Resource.from(clazz) != null);
+    // Check first if a class is a JAX-RS resource, and only if so check with validation.
+    // This prevents unnecessary warnings being logged for pure CDI beans.
+    private final Cache<Class<?>, Boolean> jaxRsResourceCache = new Cache<>(
+            clazz -> Resource.from(clazz, true) != null && Resource.from(clazz) != null);
 
     private final Hk2CustomBoundTypesProvider customHk2TypesProvider;
     private final InjectionManagerStore injectionManagerStore;


core-server/src/main/java/org/glassfish/jersey/server/model/IntrospectionModeller.java
```diff
@@ -208,8 +210,11 @@ private void checkResourceClassFields(final boolean encodedFlag,
                         field.getGenericType(),
                         field.getAnnotations());
                 if (null != p) {
```
senivam: why not to extend existing IF by adding another condition? (same question is valid for the first modification in that class)

jansupol:
@arjantijms Have you think of adding a test for the modified behaviour?

arjantijms:
@jansupol 
>Have you thought of adding a test for the modified behaviour?

I did, but it's about capturing the server log and checking that. Not 100% sure how to do that from Arquillian.

jansupol:
From the top of my head I am not sure either (not Arquillian in Jersey, right?). One way is to grab the logger something like:
`Logger logger = Logger.getLogger("org.glassfish.jersey.tracing");` and pass it the logging Handler that would be checked via `LogManager.getLogManager().readConfiguration`...
The other option would be use Jersey test framework logging mechanism, something like ` for (final LogRecord logRecord : getLoggedRecords()) {...}` but I would need to check whether either of it would actually help....

jansupol:
Do  you have some small test app that is fixed by the patch? I can try to think of some way of testing it...

arjantijms:
To reproduce I did the following.

Using the following JAX-RS app as a base: 

https://github.com/eclipse-ee4j/soteria/tree/master/test/app-jaxrs

Adding the following class to it:

```
package org.glassfish.soteria.test;

import static javax.interceptor.Interceptor.Priority.PLATFORM_BEFORE;

import java.io.Serializable;

import javax.annotation.Priority;
import javax.enterprise.inject.Intercepted;
import javax.enterprise.inject.spi.Bean;
import javax.inject.Inject;
import javax.interceptor.AroundInvoke;
import javax.interceptor.Interceptor;
import javax.interceptor.InvocationContext;
import javax.security.enterprise.authentication.mechanism.http.RememberMe;

@Interceptor
@RememberMe
@Priority(PLATFORM_BEFORE + 500)
public class FOOOOOO implements Serializable {

    private static final long serialVersionUID = 1L;

    @Inject
    @Intercepted
    private Bean<?> interceptedBean;

    @AroundInvoke
    public Object intercept(InvocationContext invocationContext) throws Exception {
        return invocationContext.proceed();
    }
}
```

Downloading Payara Server full 182 from:

https://payara.fish/downloads

Deploying the above application to it.

Then the log displays:

```
Warning: The following warnings have been detected: WARNING: Parameter interceptedBean of type javax.enterprise.inject.spi.Bean<?> from private javax.enterprise.inject.spi.Bean<?> org.glassfish.soteria.test.FOOOOOO.interceptedBean is not resolvable to a concrete type.
```

Updating Payara with the patched Jersey and deploying the same application again will not display that warning.

arjantijms:
@jansupol did you had any luck with creating a test, or looking for an approach for a test?

jansupol:
Sorry I have been occupied with other tasks this week. Fortunately, we still have some time before 2.29 goes out.

jansupol:
I was able to put together a test:
    `package org.glassfish.jersey.ext.cdi1x.internal.internal;

    import org.glassfish.jersey.server.ResourceConfig;
    import org.glassfish.jersey.server.ServerProperties;
    import org.glassfish.jersey.test.JerseyTest;
    import org.glassfish.jersey.test.TestProperties;
    import org.junit.Assert;
    import org.junit.Test;

    import javax.annotation.Priority;
    import javax.enterprise.inject.Intercepted;
    import javax.enterprise.inject.spi.Bean;
    import javax.inject.Inject;
    import javax.interceptor.AroundInvoke;
    import javax.interceptor.InvocationContext;
    import javax.ws.rs.GET;
    import javax.ws.rs.Path;
    import javax.ws.rs.core.Application;
    import java.io.Serializable;
    import java.util.logging.Level;
    import java.util.logging.LogRecord;

    import static javax.interceptor.Interceptor.Priority.PLATFORM_BEFORE;
    import static org.junit.Assert.assertEquals;
    import javax.interceptor.Interceptor;

    public class CdiComponentProviderWarningTest extends JerseyTest {

    @Path("resource")
    public static class CdiResource {
        public static final String OK = "OK";
        @GET
        public String get() {
            return OK;
        }
    }

    @Interceptor
    @Priority(PLATFORM_BEFORE + 500)
    public class FooBean implements Serializable {

        @Inject
        @Intercepted
        private Bean<?> interceptedBean;
        private static final long serialVersionUID = 1L;

        @AroundInvoke
        public Object intercept(InvocationContext invocationContext) throws Exception {
            return invocationContext.proceed();
        }
    }

    @Override
    protected Application configure() {
        set(TestProperties.RECORD_LOG_LEVEL, Level.WARNING.intValue());
        return new ResourceConfig(CdiResource.class, FooBean.class)
                .property(ServerProperties.WADL_FEATURE_DISABLE, true);
    }

    @Test
    public void testWarning() {
        String response = target("resource").request().get(String.class);
        assertEquals(response, CdiResource.OK);

        for (final LogRecord logRecord : getLoggedRecords()) {
            if (logRecord.getLoggerName().equals("org.glassfish.jersey.internal.Errors")
                && logRecord.getMessage().contains(FooBean.class.getName())) {
                Assert.fail("Checking CDI bean is a JAX-RS resource should not cast warnings");
            }
        }
    }
    }`

But the fix does not seem to work,as the check originates in `ResourceBagConfigurator#init`. @arjantijms Maybe you added the class some different way?

jansupol:
The test should have beans.xml and the FooBean does not need to be registered in ResourceConfig, right?

arjantijms:
>But the fix does not seem to work,as the check originates in ResourceBagConfigurator#init. @arjantijms Maybe you added the class some different way?

Sorry for the late reply, your reply somehow snowed under in my email box. There's indeed no need to add the class in a different way.

I just repeated 
https://github.com/eclipse-ee4j/jersey/pull/3906#issuecomment-410232589 and their it certainly does work.

Maybe it's the inner classes that make the difference?

jansupol:
@arjantijms I have added a PR with a test to [Payara w/s](https://github.com/payara/patched-src-jersey/pull/4). Please alter the test as you see fit, and update this PR.

jansupol:
Sorry it took so long, the preparation for Jakarta EE release is a bit hectic

PR #3908, opened 2018-08-03
Update to Bean Validation 2.0 (Hibernate 6)

### LINKED ISSUE #3871: Upgrade BV dependencies to latest
>         <javax.validation.version>2.0.1.Final</javax.validation.version>
>         <hibernate-validator.version>6.0.10.Final</hibernate-validator.version>

mszabo-wikia:
This is an attempt to update to version 2.0 of the [JSR 380](https://beanvalidation.org/2.0/) Bean Validation spec by updating the Hibernate version to 6.0, in order to resolve #3871.

Hibernate now [requires](https://developer.jboss.org/wiki/HibernateValidatorMigrationGuide#jive_content_id_60x) an implementation of version 3.0 of JSR 341 if a `ResourceBundleMessageInterpolator` is used, so the corresponding dependencies needed updating too. It seems that CDI API and yasson were pulling in an old JSR 341 API with a different group ID; ignoring them out fixed dependency issues in tests.

This also caused integration tests that used old Jetty plugin and depended on BV to break, since such old version (8.x) of Jetty doesn't offer a JSR 341 3.0 implementation. I switched those to use the new Jetty plugin which is already a project dependency. (It's probably something that should be eventually done for all integration tests anyways.)

According to the referenced issue a CQ will be needed for upgrading all these dependencies, but the system to do so is locked to committers only, so unfortunately I could not do anything about that. However since there was already an issue in the project I was tempted to go forward with this change.

>Signed-off-by: mszabo-wikia <mszabo@wikia-inc.com>

### DIFF
--- a/examples/declarative-linking/pom.xml
+++ b/examples/declarative-linking/pom.xml
@@ -50,7 +50,7 @@
         </dependency>
 
         <dependency>
-            <groupId>org.glassfish.web</groupId>
+            <groupId>org.glassfish</groupId>
             <artifactId>javax.el</artifactId>
         </dependency>
 
--- a/ext/cdi/jersey-cdi1x-validation/src/main/java/org/glassfish/jersey/ext/cdi1x/validation/internal/CdiInterceptorWrapperExtension.java
+++ b/ext/cdi/jersey-cdi1x-validation/src/main/java/org/glassfish/jersey/ext/cdi1x/validation/internal/CdiInterceptorWrapperExtension.java
@@ -43,7 +43,7 @@ import javax.interceptor.Interceptor;
 import org.glassfish.jersey.internal.util.collection.Cache;
 import org.glassfish.jersey.server.model.Resource;
 
-import org.hibernate.validator.internal.cdi.interceptor.ValidationInterceptor;
+import org.hibernate.validator.cdi.internal.interceptor.ValidationInterceptor;
 
 /**
  * CDI extension to register {@link CdiInterceptorWrapper}.
@@ -78,7 +78,7 @@ public class CdiInterceptorWrapperExtension implements Extension {
      * @param afterTypeDiscovery CDI bootstrap event.
      */
     private void afterTypeDiscovery(@Observes final AfterTypeDiscovery afterTypeDiscovery) {
-        afterTypeDiscovery.getInterceptors().remove(ValidationInterceptor.class);
+        afterTypeDiscovery.getInterceptors().removeIf(ValidationInterceptor.class::equals);
     }
 
     /**
--- a/pom.xml
+++ b/pom.xml
@@ -1656,7 +1656,7 @@
                 <version>${javax.el.version}</version>
             </dependency>
             <dependency>
-                <groupId>org.glassfish.web</groupId>
+                <groupId>org.glassfish</groupId>
                 <artifactId>javax.el</artifactId>
                 <version>${javax.el.impl.version}</version>
             </dependency>
@@ -1673,13 +1673,13 @@
             </dependency>
 
             <dependency>
-                <groupId>org.hibernate</groupId>
+                <groupId>org.hibernate.validator</groupId>
                 <artifactId>hibernate-validator</artifactId>
                 <version>${validation.impl.version}</version>
             </dependency>
 
             <dependency>
-                <groupId>org.hibernate</groupId>
+                <groupId>org.hibernate.validator</groupId>
                 <artifactId>hibernate-validator-cdi</artifactId>
                 <version>${validation.impl.version}</version>
             </dependency>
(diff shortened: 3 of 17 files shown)


arjantijms:
I've done similar work in the branch over here, but didn't do a PR yet:

https://github.com/payara/patched-src-jersey/tree/jersey-2.27.payara-maintenance

Might be worth it to compare for differences (if any).

mszabo-wikia:
Thanks! Looking at https://github.com/payara/patched-src-jersey/commit/bbb20bd3d3dacfcd628bd344fcff6c2299f701b5 it seems to indeed make some changes also made in this PR, but parts of it also appears to be a fix for a different issue. Is my interpretation correct?

arjantijms:
@mszabo-wikia That's indeed correct. I think @mulderbaba was about to do a PR as well for that issue.

arjantijms:
@jansupol What does CQ stand for? Is that something that's only needed for when Oracle employees do and signoff the PR?

mszabo-wikia:
according to [Eclipse Handbook](https://www.eclipse.org/projects/handbook/#ip-cq) it's required every time a dependency is updated or a new one is introduced

arjantijms:
Oh, "nice". TIL ;)

The engineer in me immediately wonders whether we can't automate that.

mszabo-wikia:
I'm inclined to agree on the automation part 😉 

On another note, I've looked at the JERSEY-2673 integration test failure, it seems it's because the test uses an ancient version of jetty-maven-plugin that has Jetty 8 and thus an old JSR 341 API. So probably the best way to solve that is to have all integration tests use the new plugin—it's already a project dependency and cleaning up the old one is a TODO so it should be fine.

ext/cdi/jersey-cdi1x-validation/src/main/java/org/glassfish/jersey/ext/cdi1x/validation/internal/CdiInterceptorWrapperExtension.java
```diff
@@ -78,7 +78,7 @@ private void beforeBeanDiscovery(@Observes BeforeBeanDiscovery beforeBeanDiscove
      * @param afterTypeDiscovery CDI bootstrap event.
      */
     private void afterTypeDiscovery(@Observes final AfterTypeDiscovery afterTypeDiscovery) {
-        afterTypeDiscovery.getInterceptors().remove(ValidationInterceptor.class);
+        afterTypeDiscovery.getInterceptors().removeIf(ValidationInterceptor.class::equals);
```
mszabo-wikia: It seems the `ValidationInterceptor` was registered more than once in a test, I'm not yet sure why.

examples/declarative-linking/pom.xml
```diff
@@ -50,7 +50,7 @@
         </dependency>
 
         <dependency>
-            <groupId>org.glassfish.web</groupId>
+            <groupId>org.glassfish</groupId>
```
jansupol: This should be updated in root pom, too. Anyway, since el was contributed to the EF, too, we need to have dependency on this new version of el, which has not been released, yet

jansupol: Sorry I missed the change the root pom...

pom.xml
```diff
@@ -1943,11 +1943,11 @@
         <javassist.version>3.22.0-CR2</javassist.version>
         <javax.annotation.version>1.2</javax.annotation.version>
         <javax.annotation.bundle.version>1.2</javax.annotation.bundle.version>
-        <javax.el.version>2.2.4</javax.el.version>
-        <javax.el.impl.version>2.2.4</javax.el.impl.version>
+        <javax.el.version>3.0.1-b06</javax.el.version>
+        <javax.el.impl.version>3.0.1-b09</javax.el.impl.version>
```
jansupol: latest is b-10. but at the end we would need to update to some EPL licensed version before release

jansupol: I do not mean it is as a PR merge blocker, just a note of needs to be done before the 2.28 release which is going to happen very soon.

tests/integration/jersey-2689/pom.xml
```diff
@@ -81,7 +81,7 @@
                 <artifactId>maven-failsafe-plugin</artifactId>
             </plugin>
             <plugin>
-                <groupId>org.mortbay.jetty</groupId>
```
jansupol: Why this has been updated in just these two tests? Jetty is used many places

jansupol:
@mszabo-wikia Thanks for this PR, this is going to help a lot with the 2.28/2.29 release. I can see you tried to get the latest jetty in the tests that fail for you. I think this dependency should be in the plugin management, too. Since Jetty is Eclipse project, we do not need to file a CQ for Jetty. I will file a CQ for hibernate validation, which we need to do. Possibly we can have both Jetty versions in the plugin management, so that we do not need to update jetty everywhere in the project?

jansupol:
I have added [CQ](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=17142). The latest BV is 6.0.11.Final though. Would need to upgrade if GF does.

jansupol:
The CQ has been approved

tests/integration/jersey-2689/pom.xml
mszabo-wikia: @jansupol yeah I agree it's good to be consistent there, just wasn't sure if making that upgrade should be done in the scope of this PR or separately :)

mszabo-wikia:
@jansupol It seems Hibernate have [released](https://developer.jboss.org/wiki/HibernateValidatorMigrationGuide#jive_content_id_60x) 6.0.12.Final to fix a performance regression. Should I bump the version in this PR?

jansupol:
@mszabo-wikia  please do not. It makes sense to be aligned with glassfish/nucleus I think for now. But if you can update the project pom with org.eclipse.jetty it would be great.

mszabo-wikia:
@jansupol sure, thanks :)

I'm not sure what you mean by updating the project POM though—Jetty, and the Jetty Maven Plugin are already listed there and I didn't change their version. It's just those tests still using the old Jetty Maven Plugin under the `org.mortbay.jetty` group ID had to be changed to use `org.eclipse.jetty` instead since the old plugin's Jetty version doesn't offer a recent enough EL implementation. But neither of the Jetty plugins were updated since the new one was already in the project too.

jansupol:
@mszabo-wikia  Indeed it is there, sorry I did not check it before. Thanks.

PR #3915, opened 2018-08-13
Fix WORKER_BY_TYPE_COMPARATOR to follow contract.

hatstand:
Given:
```
MessageBodyWriter<A>
MessageBodyWriter<B>
MessageBodyWriter<C>
```

where no media types are specified and `C` extends `A`.

When sorting these using the `WORKER_BY_TYPE_COMPARATOR`, the compare
contract is broken as:

    a == b

which implies that either:

    a < c && b < c

or:

    a > c && b > c

or:

    a == b == c

However, with this comparator:

    a < c && b == c

The fix, I think, is to no longer have two `MessageBodyWriters` compare as
equal if they provide a different type.


hatstand:
This fixes a heisenbug (#3387) for us, so it would be great if someone could review & merge this, thanks!

hatstand:
Hi, could we get this merged, please?

hatstand:
Ping?

PR #3918, opened 2018-08-14
#3700 - Fixed verification logic for component instances

serhiypal:
The PR is aimed to fix the issue #3700 by adding missing verification logic

### DIFF
--- a/core-server/src/main/java/org/glassfish/jersey/server/ResourceModelConfigurator.java
+++ b/core-server/src/main/java/org/glassfish/jersey/server/ResourceModelConfigurator.java
@@ -21,7 +21,6 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.IdentityHashMap;
 import java.util.Set;
-import java.util.function.BiPredicate;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
 
@@ -113,31 +112,9 @@ public class ResourceModelConfigurator implements BootstrapConfigurator {
 
         Set<Class<?>> registeredClasses = runtimeConfig.getRegisteredClasses();
 
-        /*
-         * Check the {@code component} whether it is correctly configured for client or server {@link RuntimeType runtime}.
-         */
-        java.util.function.Predicate<Class<?>> correctlyConfigured =
-                componentClass -> Providers.checkProviderRuntime(componentClass,
-                        componentBag.getModel(componentClass),
-                        RuntimeType.SERVER,
-                        !registeredClasses.contains(componentClass),
-                        resourceClasses.contains(componentClass));
-
-        /*
-         * Check the {@code resource class} whether it is correctly configured for client or server {@link RuntimeType runtime}.
-         */
-        BiPredicate<Class<?>, ContractProvider> correctlyConfiguredResource =
-                (resourceClass, model) -> Providers.checkProviderRuntime(
-                        resourceClass,
-                        model,
-                        RuntimeType.SERVER,
-                        !registeredClasses.contains(resourceClass),
-                        true);
-
-        Set<Class<?>> componentClasses =
-                componentBag.getClasses(ComponentBag.excludeMetaProviders(injectionManager)).stream()
-                        .filter(correctlyConfigured)
-                        .collect(Collectors.toSet());
+        Set<Class<?>> componentClasses = componentBag.getClasses(ComponentBag.excludeMetaProviders(injectionManager)).stream()
+                .filter(clazz -> isComponentClassConfiguredCorrectly(componentBag, registeredClasses, resourceClasses, clazz))
+                .collect(Collectors.toSet());
 
         // Merge programmatic resource classes with component classes.
         Set<Class<?>> classes = Collections.newSetFromMap(new IdentityHashMap<>());
@@ -157,7 +134,7 @@ public class ResourceModelConfigurator implements BootstrapConfigurator {
                     continue;
                 }
 
-                if (model != null && !correctlyConfiguredResource.test(componentClass, model)) {
+                if (isNotCorrectlyConfiguredResource(registeredClasses, componentClass, model)) {
                     model = null;
                 }
                 resourceContext.unsafeBindResource(componentClass, model);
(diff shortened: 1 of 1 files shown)


serhiypal:
The logic is similar to https://github.com/eclipse-ee4j/jersey/pull/3817 but in this PR there is a separation between component classes and component instances checks.

core-server/src/main/java/org/glassfish/jersey/server/ResourceModelConfigurator.java
```diff
@@ -113,31 +112,9 @@ private void bindProvidersAndResources(
 
         Set<Class<?>> registeredClasses = runtimeConfig.getRegisteredClasses();
 
-        /*
-         * Check the {@code component} whether it is correctly configured for client or server {@link RuntimeType runtime}.
-         */
-        java.util.function.Predicate<Class<?>> correctlyConfigured =
```
serhiypal: Extracted into separate methods for readability

serhiypal: `isComponentClassConfiguredCorrectly`

core-server/src/main/java/org/glassfish/jersey/server/ResourceModelConfigurator.java
```diff
@@ -113,31 +112,9 @@ private void bindProvidersAndResources(
...
-                componentClass -> Providers.checkProviderRuntime(componentClass,
-                        componentBag.getModel(componentClass),
-                        RuntimeType.SERVER,
-                        !registeredClasses.contains(componentClass),
-                        resourceClasses.contains(componentClass));
-
-        /*
-         * Check the {@code resource class} whether it is correctly configured for client or server {@link RuntimeType runtime}.
-         */
-        BiPredicate<Class<?>, ContractProvider> correctlyConfiguredResource =
```
serhiypal: `isNotCorrectlyConfiguredResource`

core-server/src/main/java/org/glassfish/jersey/server/ResourceModelConfigurator.java
```diff
@@ -197,4 +173,22 @@ private boolean bindWithComponentProvider(
...
+                !registeredClasses.contains(resourceClass), true);
+    }
+
+    private static boolean isComponentClassConfiguredCorrectly(ComponentBag componentBag, Collection<Class<?>> registeredClasses,
+            Collection<Class<?>> resourceClasses, Class<?> componentClass) {
+        return Providers.checkProviderRuntime(componentClass, componentBag.getModel(componentClass), RuntimeType.SERVER,
+                !registeredClasses.contains(componentClass), resourceClasses.contains(componentClass));
+    }
+
+    private static boolean isComponentInstanceConfiguredCorrectly(ComponentBag componentBag,
```
serhiypal: This is new method that checks instances

core-server/src/main/java/org/glassfish/jersey/server/ResourceModelConfigurator.java
```diff
@@ -167,17 +144,16 @@ private void bindProvidersAndResources(
         }
 
         // Merge programmatic resource instances with other component instances.
-        Set<Object> instances =
-                componentBag.getInstances(ComponentBag.excludeMetaProviders(injectionManager)).stream()
-                        .filter(instance -> correctlyConfigured.test(instance.getClass()))
-                        .collect(Collectors.toSet());
+        Set<Object> instances = componentBag.getInstances(ComponentBag.excludeMetaProviders(injectionManager)).stream().filter(
+                instance -> isComponentInstanceConfiguredCorrectly(componentBag, registeredClasses, resourceInstances, instance))
```
serhiypal: New method `isComponentInstanceConfiguredCorrectly` is used here

senivam:
Actually I do not agree with merge of this pull request (any way it's good work but the framework shall not be modified in this case). The behavior is actually expected. My arguments why it shall not be merged (I will copy that to the issue #3700 as well):

Jersey supports 2 ways of components registration - Class based registration and Instance based registration. When the class based registration is used the life-cycle of components is fully managed by the JAX-RS implementation or any underlying IoC container supported by the implementation. 

On the other hand when the instance based registration is used the life-cycle of providers registered using this instance-based register(...) is not managed by JAX-RS runtime.

That is the reason why in the issue #3700 the HelloWorldResource instance is not recognized as resource. It is expected to implement some Provider interface (which is noted in the WARNING). However it still works because the registration is valid even though it did not pass through JAX-RS IoC. 

But for resources which are not going to be providers nor features (not implement those interfaces) there is class based registration which shall be called like `register(HelloWorldResource.class)` (and not `register(new HelloWorldResource())` ). In that case the component is supposed to be a resource and no WARNING occurs.

senivam:
@serhiypal OK, will try to describe it in other words. 

Let's presume that the purpose of the class and instance components registrations is to distinguish between a basic resource and something more complex (Feature/Provider) but you still can use class registration to register just everything in Jersey. 

So, the instance is registered and now it shall be validated. The first step is to check if it compliances with Feature/Provider (which is not simple resource itself). So it shall implement some appropriate methods/interfaces. No, it does not implement - Jersey says no, it's not a provider so here is a WARNING about that and we do not use it as a provider. But the next step is to validate if the component could be used as a resource. And yes it can. So it's registered as a resource. The WARNING here only notifies that Jersey expects something more complex than a resource if the instance registration is used instead of class registration. Nothing critical but user shall know that if there is no specific purpose a more simple instrumentation shall be used.

 When a resource is registered via class registration the validation steps are the same. However in both cases Jersey presumes that it works with resources. So it generates no warnings and simply uses registered classes (if they passes all other validations for sure).  

What I see in the PR is that validation is modified in the way to allow instance to be treated as a resource. Which is not fully correct. Instance is expected to be something more complex. 

Even though the warning here does nothing with registered component. Just a notice that probably class registration could be used.

serhiypal:
@senivam gotcha, thanks.

serhiypal:
According to the comment above the behavior is expected so closing the PR

PR #3931, opened 2018-09-10
Fix license information

senivam:

### DIFF
--- a/NOTICE.md
+++ b/NOTICE.md
@@ -70,10 +70,6 @@ Javassist Version 3.22.0-CR2
 * Project: http://www.javassist.org/
 * Copyright (C) 1999- Shigeru Chiba. All Rights Reserved.
 
-Java(TM) EE Interceptors 1.1 API Version 1.0.0.Beta1
-* License: LGPL 2.1
-* Copyright 2005, JBoss Inc., and individual contributors as indicated by the @authors tag.
-
 Jackson JAX-RS Providers Version 2.8.10
 * License: Apache License, 2.0
 * Project: https://github.com/FasterXML/jackson-jaxrs-providers
--- a/core-client/src/test/java/org/glassfish/jersey/client/ClientConfigTest.java
+++ b/core-client/src/test/java/org/glassfish/jersey/client/ClientConfigTest.java
@@ -158,7 +158,7 @@ public class ClientConfigTest {
     }
 
     /**
-     * Copied from DefaultConfigTest.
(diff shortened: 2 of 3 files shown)


jansupol:
@mkarg It was made in my clone with my sign-off, @senivam created the PR for me while I was busy with some other stuff. Do you think it can bring some issues?

mkarg:
@jansupol I don't know actually. The EF is serious about that sign-off stuff so it might be better to be careful and correct it. On the other hand, really, I do not know if somebody or a tool would have noticed.

senivam:
Pull requests are not history units. In git's history only commiters and authors remains:
`commit ff7183d3ccce66b0c58ac99d3b0ee876123dd7a0
Author: Maxim Nesen <24524084+senivam@users.noreply.github.com>
Commit: jansupol <15908245+jansupol@users.noreply.github.com>

    JerseySseEventSource#register throws IllegalStateException (#3891)
    
`
sign-off always belongs to the author independent on the commiter.  Commiter is the person who merged a pull request. Author is the person who contributes the code. Role of a person who created a pull request is miserable and can be omitted. 

In situations when author did a commit with proper sign-off and left for long-term absence or is just not available for indefinite time it's probably the only solution to merge author's code when someone else does a pull request for that commit. Author remains the same. Person who performs pull request is not important.

mkarg:
@senivam That's all clear. I think the best solution is to not sign-off _PRs_ at all but solely sign-off _COMMITs_. Just as a proposal.

mkarg:
@jansupol Agreed completely. 

>The PR has *just a comment*...

The "signoff" feature of git (i. e. `-s` switch) effectively does nothing other than simply adding the Sign-Off line to the end of a commit's *comment*. It is *not* a separate information. The merge of a Github PR *produces a commit* ("merge-commit"); for this Github takes the comment of the PR as the comment of that merge commit. As it is a Sign-Off line, the result will be a signed commit. The effect is that in the end for git (and for Github) the merged PR looks like a simple list of commits (Jan's work) plus one commit ontop (Maxim's merge), where Jan's work is signed by Jan, and Maxim's merge-commit is signed by Jan (not Maxim). While this looks trivial, it is a feast for lawyers: The sign-off is enforced by Eclipse Legal because it is the digital way of *initialing a treaty*. It is like Maxim sign's a contract with Jan's name! So to sum up: We should not write "Sign-Off" in a PR comment unless it contains the PR-author's name. :-)

PR #3939, opened 2018-09-20
Replace CDI servlet uber jar with CDI servlet-core jar

jansupol:


jansupol:
CDI servlet uber jar contains CDI API with CDDL license, which would be hard to get CQ on. The new jar contains just the weld code under Apache license

PR #3944, opened 2018-09-26
a ReaderWriter utilities getWriter and getReader

tareqhs:


jansupol:
Can you please elaborate why this is useful?

PR #3949, opened 2018-09-28
PAYARA-3119 NullPointerException when starting Jersey/EJB Containers with <initialize-in-order> flag true in Ear

jGauravGupta:
Signed-off-by: Gaurav Gupta <gaurav.gupta.jc@gmail.com>

When multiple WARs are in an EAR with `initialize-in-order` set to true, and both contain both an EJB and a JAX-RS endpoint, a NullPointerException is thrown on startup. Actually during startup of first module, Jersey provider registers the interceptor for all modules (even other modules are not started in case of <initialize-in-order>). And during the startup of second module, again Jersey provider registers the interceptor for all modules.


jGauravGupta:
Hi @jansupol,

Can you trigger CI for this PR as test passing for OracleJDK.

Issue #3670, opened 2017-09-14
Broken ParamConverterProvider ordering in 2.26


jerseyrobot:
It looks like there are a few issues regarding provider ordering in 2.26. Although I think older versions are also affected.

I created a minimal sample app to reproduce the issue: https://github.com/chkal/jersey-prio

Basically I'm trying to register a custom `ParamConverter` which overrides the default converter for the data type `java.lang.Integer`.

**First issue:**

The first issue is that whether the custom provider or the default provider is used randomly changes between deployments. That's caused by the use of a `HashSet` instead of a `LinkedHashSet` in `ParamConverterFactory`. My pull request #3669 should fix this.

**Second issue**

The second problem is more difficult to reason about. After merging #3669 you will see that Jersey will always pick the default provider and ignores the custom provider deployed with the app. 

I'll try to explain what I debugged so far. Actually the code in the `ParamConverterFactory` tries to handle the case of custom providers overriding default providers correctly:

https://github.com/jersey/jersey/blob/4ecf20c7585abfe1bd2c312d890943e884a9eb3b/core-server/src/main/java/org/glassfish/jersey/server/internal/inject/ParamConverterFactory.java#L74-L81

The weird thing is that providers deployed with the app always end up in the `providers` set in the constructor and NOT in the `customProviders` set. The `ParamConverterFactory` is created here:

https://github.com/jersey/jersey/blob/4ecf20c7585abfe1bd2c312d890943e884a9eb3b/core-server/src/main/java/org/glassfish/jersey/server/internal/inject/ParamExtractorConfigurator.java#L68-L71

If you dig deeper into the `Provider.getProviders()` and `Provider.getCustomProviders()` code, you will see that `getCustomProviders()` will look for providers with a qualifier `Custom`. I'm not sure under which circumstances providers get this qualifier. 

Also noteworthy: Only `getCustomProviders()` seems to apply the ordering algorithm specified in JAX-RS 2.1 (see jax-rs/api#538). As custom providers aren't found by this method, the ordering algorithm isn't applied at all.

I hope this description helps to debug the problem. Let me know if there is anything I can do to help with this.

chkal:
Hi all. As open merge requests haven't been migrated from the jersey/jersey repository, I create a new pull request #3955 which contains the changes from jersey/jersey#3669. This pull request fixes the first issue described above.

It would be great to get some feedback from the Jersey team regarding the second issue.

jansupol:
[...]

PR #3925, opened 2018-09-04
PAYARA-3038/3077 Jersey JAXB Unmarshalling Error

jGauravGupta:
Fixes : https://github.com/payara/Payara/issues/3077

Signed-off-by: Gaurav Gupta <gaurav.gupta.jc@gmail.com>

PR #3955, opened 2018-10-03
ParamConverterFactory should use LinkedHashSet to preserve ordering

chkal:
This pull request has been migrated from jersey/jersey#3669. It fixes the first issue described in #3670.

In a nutshell: If you provide a custom `ParamConverterProvider` for a data type for which there is also a builtin one provided by Jersey, a random one wins and gets executed. This changes even between deployments. This issue is caused by a `HashSet` in `ParamConverterFactory` which should actually be a `LinkedHashSet` to preserve the ordering.


arjantijms:
Maybe it's unrelated, but this test is failing now:

```
Failed tests: 
  BroadcasterTest.test:193 Waiting for resultsB1 to be complete failed.
```

senivam:
shall be unrelated.. however the full PR duplicates 3925 or am I missing something?

chkal:
@senivam Yes, looks like #3925 is actually fixing the same issue.

senivam:
but this gained 2 approves

PR #3958, opened 2018-10-05
Update pom.xml

arjantijms:


arjantijms:
@bshannon 
>I don't know why we're wasting time on reviews of trivial changes such as this...

I don't know either, seems like a massive waste of resources. In JAX-RS project it's even worse. Even the most trivial change must have 2(!) reviews and a 2 week waiting time.

arjantijms:
@asoldano We still need a second review on this. Can you take a look as well?

arjantijms:
@bravehorsie What do you think about this PR?

arjantijms:
@senivam Care to approve this PR? Thx in advance!

chkal:
> I don't know either, seems like a massive waste of resources.

Well, it is up to the corresponding project to define rules for code reviews. Are there such rules for Jersey defined anywhere? My impression was that there are no strict rules. 

> In JAX-RS project it's even worse.

I don't think it is fair to criticize JAX-RS for the review rules. JAX-RS is one of the most active EE4J projects, already passed the first release review and implemented a first major feature for the next spec version. And we did all this even with a 2-week review period.

> Even the most trivial change must have 2(!) reviews and a 2 week waiting time.

That's not correct. Yes, the default review period is two weeks. Of course that has been defined with pull requests in mind which change the API of JAX-RS. And [you also voted +1](https://www.eclipse.org/lists/jaxrs-dev/msg00147.html) for these rules. However, we also defined in the [committer conventions](https://github.com/eclipse-ee4j/jaxrs-api/wiki/Committer-Conventions), that a PR assignee may reduce the review period if the actual API is unchanged. So for trivial changes the review period is definitely shorter than two weeks.

arjantijms:
>That's not correct. Yes, the default review period is two weeks. Of course that has been defined with pull requests in mind which change the API of JAX-RS.

Yes, indeed. So you're right, sorry for sounding to black and white.

Still, I had the idea the two weeks was for major new features, the kind of features that may be controversial. I should of course had mentioned it at the time, but I didn't realised so much then that even an API project may have tons of little changes. For example, small typos in Javadoc etc, or an update of the mavin plug-ins and what have you.

>So for trivial changes the review period is definitely shorter than two weeks.

You're right again, although it also says the following:

"but it must never be shorter than one week."

Maybe we should change that to accommodate truly trivial changes such as the ones mentioned above? It's source control, almost all of us are adults I think, so I may perhaps be better if for trivial changes there's no or at most one review?

chkal:
@arjantijms I'm open for discussing changes to the JAX-RS review process. Basically the current rules were a "first try" for defining formal rules and I think it is perfectly fine to adjust these rules if we think that they aren't optimal in all cases. But let's discuss this on the JAX-RS list.

PR #3968, opened 2018-10-14
Release http(content)/http2(frame) bytebuf to prevent memory leak

lcybo:
As the last handler that accessed the http data(bytebuf), Jersey handler has the responsibility to release them.
That require support for `releaseOnClose` in `ByteBufInputStream` since `4.1.7.Final`.
Drop by to construct http2 pipeline via builder style.([detail](https://github.com/netty/netty/commit/74f24a5c19f8f351e9c6a7a84bdd9fbcc7a07ada))
Bump the version of netty to `4.1.30.Final`.

### DIFF
--- a/connectors/netty-connector/src/main/java/org/glassfish/jersey/netty/connector/JerseyClientHandler.java
+++ b/connectors/netty-connector/src/main/java/org/glassfish/jersey/netty/connector/JerseyClientHandler.java
@@ -101,7 +101,7 @@ class JerseyClientHandler extends SimpleChannelInboundHandler<HttpObject> {
                     }
                 });
 
-                jerseyResponse.setEntityStream(new NettyInputStream(isList));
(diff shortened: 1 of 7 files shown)


jansupol:
Filed [CQ for Netty 4.1.31](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=19281)

010gvr:
@lcybo @jansupol Not sure if this PR is still actively being monitored. But adding `ChannelHandlerContext` to a data processing class (`NettyInputStream`) feels a bit out of scope. Also, further load testing of current implementation revealed even more problems with `JerseyChunkedInput` and built in buffers leading to timeout and other related overheads. I've addressed several of these in https://github.com/eclipse-ee4j/jersey/pull/4312 which may be a permanent fix to all of netty/jersey issues that has been tested at a large-scale production use.

PR #3976, opened 2018-10-31
Remove old JDK backwards compatibility classes

mszabo-wikia:
The `ConcurrentHashMapV8` class is simply the implementation of `java.util.concurrent.ConcurrentHashMap`, as found in Java 8 and above. It seems it was added to the source to allow older JDK versions to take advantage of its more efficient implementation. However, since version 2.26, Jersey only supports Java 8 and above, therefore this class should no longer be needed and `java.util.concurrent.ConcurrentHashMap` can be used instead. This also has the advantage of removing some `sun.misc.Unsafe` usages from Jersey proper.

Similarly there was a `LinkedTransferQueue` implementation for Java 6 and older. I've removed this too, and the `JdkVersion` utility class (which was also not ready to support Java 11 version strings) along with it.


mszabo-wikia:
Not sure if the test failure is related—it passed on the first commit as well as on OpenJDK 8.

jansupol:
While you are right that the jdk 8 is minimum, we definitely cannot remove public classes/method.

jansupol:
Also, I assume we will have some use of the JdkVersion class, in trasition to jdk 9, where we will get rid of the usage of Unsafe class.

jansupol:
I incline to keep the factory methods even when they return directly the new instance of a single class.

mszabo-wikia:
@jansupol thanks! I'll prepare a new PR that avoids breaking BC & extend `JdkVersion` to support modern Java version strings.

Issue #3672, opened 2017-09-18
JSON-Jackson does not handle resource returning CompletionStage<X>


jerseyrobot:
It seems like `jersey-media-json-jackson` is not up to date with the latest JAX-RS 2.1 features.
I tried creating the following resource method:

    @GET
    @Produces({"application/json" })
    @Path("async")
    public CompletionStage<Hello> asyncHello(@NotNull @PathParam(value = "lang") String lang,
                                             @NotNull @Size(min = 2) @QueryParam("yo") String greet){
            return CompletableFuture.supplyAsync(() -> new Hello(lang, greet);
    }

This fails with stack trace

```
java.lang.IllegalArgumentException: Class org.kantega.reststop.helloworld.jaxrs.Hello not subtype of [simple type, class java.util.concurrent.CompletionStage<org.kantega.reststop.helloworld.jaxrs.Hello>]
	at com.fasterxml.jackson.databind.type.TypeFactory.constructSpecializedType(TypeFactory.java:359)
	at org.glassfish.jersey.jackson.internal.jackson.jaxrs.base.ProviderBase.writeTo(ProviderBase.java:624)
	at org.glassfish.jersey.message.internal.WriterInterceptorExecutor$TerminalWriterInterceptor.invokeWriteTo(WriterInterceptorExecutor.java:266)
```

Returning a plain String with `text/plain` works as expected.

Just as a test I build my own version of Jersey with some additional code in `org.glassfish.jersey.jackson.internal.jackson.jaxrs.base.ProviderBase.writeTo(...)`:
    
    TypeFactory typeFactory = writer.getTypeFactory();

    String typeName = genericType.getTypeName();
    if(typeName.contains(CompletionStage.class.getName())) {
        Type[] actualTypeArguments = ((ParameterizedTypeImpl) genericType).getActualTypeArguments();
        genericType = actualTypeArguments[0];
    }

    JavaType baseType = typeFactory.constructType(genericType);

The middle part unwrapping the CompletionStage is new.
This works.

So, have I missed something, or is the `CompletionStage` support not complete?

dandew:
@computerlove Maybe worth trying to send a PR with your proposed fix? This bug is turning into a showstopper.

mageshwaranr:
Faced the exact same problem. Following lines of code helped fixed the issue. 

[...]

PR #3978, opened 2018-10-31
JSON Jackson does not handle resource returning CompletionStage<X> #3672

mageshwaranr:
A resource method like below fails with java.lang.IllegalArgumentException: Class *  not subtype of *. 

```
@GET
@Produces({"application/json" })
@Path("async")
public CompletionStage<Hello> asyncHello(@NotNull @PathParam(value = "lang") String lang,
                                         @NotNull @Size(min = 2) @QueryParam("yo") String greet){
        return CompletableFuture.supplyAsync(() -> new Hello(lang, greet);
}
```

This pull request attempts to fix this. Let me know whether the fix is appropriate.

### DIFF
--- a/media/json-jackson/src/main/java/org/glassfish/jersey/jackson/internal/jackson/jaxrs/base/ProviderBase.java
+++ b/media/json-jackson/src/main/java/org/glassfish/jersey/jackson/internal/jackson/jaxrs/base/ProviderBase.java
@@ -6,12 +6,14 @@ import java.io.OutputStream;
 import java.io.Reader;
 import java.io.Writer;
 import java.lang.annotation.Annotation;
+import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.concurrent.CompletionStage;
 import java.util.concurrent.atomic.AtomicReference;
 
 import javax.ws.rs.core.MediaType;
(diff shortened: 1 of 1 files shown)


jansupol:
#3672

jansupol:
There is a valid ECA on file for mageshwaran.r@gmail.com.

mageshwaranr:
Sorry for the delay. Added test cases. Is there a way to re-trigger build ?
Not sure whether test is failing due to code issue.

mageshwaranr:
@jansupol Anything expected from me to take this forward ?

dandew:
@jansupol Can we please have this in 2.29? This has been hurting for a while now 😅

jansupol:
Hm,...I was about to approve this...but that I realized (late, sorry)...

org/glassfish/jersey/jackson/internal/jackson/jaxrs/base/ProviderBase.java is actually repackaged from [Jackson 2.8.10](https://github.com/FasterXML/jackson-jaxrs-providers/blob/jackson-jaxrs-providers-2.8.10/base/src/main/java/com/fasterxml/jackson/jaxrs/base/ProviderBase.java). 

Should the PR be made against Jackson rather than against Jersey? If a new version gets repackaged, the code could be lost...The best way is to create a PR against Jackson (2.9.x) and then a PR/issue to Jersey to repackage the latest version

Sorry for that coming late, but it still is doable before 2.29.

jansupol:
Hm... If it gets merged to Jackson, we can merge this to Jersey, too. But since this is 3rd party code, we need to ask for a [CQ](https://github.com/eclipse-ee4j/jersey/wiki/About-CQs), first.

jansupol:
The last commit is missing sign-off-by (-s)...

examples/json-jackson/src/main/java/org/glassfish/jersey/examples/jackson/CompletableFutureResource.java
```diff
@@ -0,0 +1,44 @@
+package org.glassfish.jersey.examples.jackson;
```
jansupol: Copyright header is missing

jansupol:
@mageshwaranr The code in Jackson seems to be stable at least from 2.8.4. I think we could add this to Jersey. Can you please fix the reported issues?
- copyright header in the example resource class
- add sign off
- please force push so that git hub check does not see the commit with the missing sign-off

jansupol:
We have adopted [Jackson 2.9.9](https://github.com/eclipse-ee4j/jersey/pull/4144)

edinardo:
@mageshwaranr could you please add the last request from @jansupol so we can have it.
So far, this is the only impediment to use JAX-RS 2.1 async standard with the official build.

jpadre:
> 
> 
> @mageshwaranr could you please add the last request from @jansupol so we can have it.
> So far, this is the only impediment to use JAX-RS 2.1 async standard with the official build.

There's one more thing missing to complete this for me, which is to unwrap CompletionException properly so appropriate javax.ws.rs.ext.ExceptionMapper are called down the line.

If it helps anyone else, I am modifying the mapException method of org.glassfish.jersey.server.ServerRuntime using this code to workaround it, but unsure if there's a better way it can be integrated:

`
private Response mapException(final Throwable originalThrowable) throws Throwable {
            LOGGER.log(Level.FINER, LocalizationMessages.EXCEPTION_MAPPING_START(), originalThrowable);
            Throwable throwable = originalThrowable;
            boolean inMappable = false;
            boolean mappingNotFound = false;

            do {
                //---------------------------------- Begin Change -----------------------------------
                if (throwable instanceof CompletionException && throwable.getCause() != null) {
                    throwable = throwable.getCause();
                }
                //----------------------------------- End Change ------------------------------------   
                if (throwable instanceof MappableException) {`

Issue #3500, opened 2017-02-06
Buffer overflow due to timeout with Netty + Jersey


jerseyrobot:
When returning a large response using the Netty connector, a "Buffer overflow" error occurs. Switching from `jersey-container-netty-http` to `jersey-container-grizzly2-http` resolves the issue, so it appears to be specific to the netty connector.

Testing with a 521k response and Jersey 2.25.1, the buffer overflow occurs in org.glassfish.jersey.netty.connector.internal.JerseyChunkedInput#write at line 224\. The revelant code is:

```
private void write(Provider<ByteBuffer> bufferSupplier) throws IOException {

        checkClosed();

        try {
            boolean queued = queue.offer(bufferSupplier.get(), WRITE_TIMEOUT, TimeUnit.MILLISECONDS);
            if (!queued) {
throw new IOException("Buffer overflow.");
            }

        } catch (InterruptedException e) {
            throw new IOException(e);
        }
    }
```

The queue has a capacity of 8, and the offer timeout is 1000ms, which seems low.

Relevant portion of the stack trace:

```
Caused by: java.io.IOException: Buffer overflow.
	at org.glassfish.jersey.netty.connector.internal.JerseyChunkedInput.write(JerseyChunkedInput.java:224)
	at org.glassfish.jersey.netty.connector.internal.JerseyChunkedInput.write(JerseyChunkedInput.java:204)
	at org.glassfish.jersey.message.internal.CommittingOutputStream.write(CommittingOutputStream.java:229)
	at org.glassfish.jersey.message.internal.WriterInterceptorExecutor$UnCloseableOutputStream.write(WriterInterceptorExecutor.java:299)
	at sun.nio.cs.StreamEncoder.writeBytes(StreamEncoder.java:221)
```
#### Environment
OpenJDK 1.8.0_121
#### Affected Versions
[2.25.1]

rvowles:
This is now a blocking issue for me, is there a workaround? I am having to drop and filter data because I can't get the buffer to stop overflowing.

brettkail-wk:
Not as far as I know.  jersey/jersey#3791 would need to be applied.  I'm no longer pursuing a fix since we stopped using Netty with Jersey.

rvowles:
This affected us so badly we have now done the same and swapped to Grizzly for Jersey. I will keep watch on this ticket as swapping back (which is desirable as we also use gRPC) is fairly easy at the moment.

jansupol:
[...]

PR #3983, opened 2018-11-02
 Upgrade Netty and expose Netty Context

juur:
Bumped netty version to latest
Updated API usage to match
Enhanced returned SecurityContext to include Netty information


jansupol:
There is a valid ECA on file for eclipse@wut.to.

juur:
Does the CQ required label require me to do anything?

jansupol:
No, I think only committers can file CQs. Once Jersey goes through release review, we will deal with new CQs

mkarg:
@jansupol FYI I just tested it, and you are right: The Eclipse Committer Tools present a 404 to me when trying to file a CQ for Jersey, while it works for JAX-RS. So this is definitively a committers-only task.

jansupol:
Created [CQ for Netty 4.1.31.FINAL](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=19281)

jansupol:
CQ has been approved.

rvowles:
Netty is still unusable because of #3500

PR #3987, opened 2018-11-08
updated the pom to use yasson version 1.0.2 rather than 1.0.1

mattHoyDev:


mattHoyDev:
Not sure why I did not pass Eclipse validation, I've signed the ECA.

jansupol:
We need to use 1.0.3 once its released for 2.28

PR #4013, opened 2018-12-21
Pom fixes

romain-grecourt:
Pom fixes for things that break the GlassFish integration currently.

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -30,6 +30,7 @@
 
     <groupId>org.glassfish.jersey</groupId>
     <artifactId>jersey-bom</artifactId>
+    <version>2.28-SNAPSHOT</version>
     <packaging>pom</packaging>
     <name>jersey-bom</name>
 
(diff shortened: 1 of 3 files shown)


senivam:
@jansupol no, it was done because of release process - when a new version was set (a project version was modified generally) using maven versions plugin, Jersey failed to compile any more due to inconsistency in the project which appeared after bom/pom.xml project version was changed. So I've asked if it's possible to modify bom's parent (as soon as I was not sure what more it will break). After short 3 hours discussion I was allowed to do that so I've submitted a PR to Jersey which was approved and release was fixed. However as appears now this breaks GF integrasion, so I will try to solve inconsistency in some other way. BTW, I suppose after this PR is merged release will be for some time unavailable.

romain-grecourt:
One of the problems that I'm aware of is that the bom pom is not processed by the reactor ordering resolution.

I saw that jersey uses it in the tests and example that could be the underlying problem.
This assumption could be wrong though since it worked before.

senivam:
@romain-grecourt I've already solved that problem (bom is not in reactor) in the release job, I suppose finally release will work as it is supposed to be

PR #4019, opened 2018-12-28
Prepare 2.28 - rc4

jansupol:


jansupol:
Do not squash this PR

jansupol:
> The term Intellectual Property (IP) refers to any sort of creative work, be it literature, art, or software. In the realm of open source software, artifacts like source code, documentation, and images are considered intellectual property. Unless otherwise stated, intellectual property is the property of its creator, who may grant permission for others to use that intellectual property by providing a license.

@mkarg You need to explain this. You have created a single symbol change commit that broke the build. You pushed the PR, you left the build fix on us, you did not try to fix it. We have spent our time (significantly more time than one symbol change) in finding the correct fix. And you want to claim that all to be your IP and you say we are not fair to you?

We gladly helped you [the last time](https://github.com/eclipse-ee4j/jersey/pull/3991) with the failing tests, it took us quite some time to find the OSGi fix. But simply updating versions without addressing issues caused by the new versions and breaking builds is not the way how anyone should do the PRs.

jansupol:
@mkarg Happy new year to you too. I hoped the new year won't bind us with a new set of rules, especially about trivial PRs.  Really, we are glad for PRs, even or the ones that just upgrade a version of a dependency. But if the PR does not build, and does not contain any change that solves any issue, it just upgrades a version,  I feel free to ignore it. 

In general, PRs like this bring only additional complexity for others whoever wants to provide fixes, while the added value provided by the original trivial commit is arguably none. All that would really looks like the reason is not that the committer wants to help, it looks more like the reason is to have his own name in the log. I do not feel like we should support that. I do not suspect it's your case, I am sure you are capable of providing lot's of great PRs, but in general, we really should not support these trivial PRs.

I do not imply PRs should be 100% working.  As soon as they contain any added value, they originator should not be omitted and I am happy to spend some time with helping it. Having `WIP:` is a good idea, but it sounds more like the original committer is still working on it and just accepts comments. If the committer is stuck and needs help, it's better to add a comment. But please, let it be a meaningful PR.

jansupol:
Running `addLegalInfo` profile is rather slow. It extends build time from ~20 minutes to more than an hour. While having this in a profile used just once per release could be ok, it would be better to use a different plugin that does not slow the build that drastically. Experiments show the build-helper-maven-plugin could be replaced.

jansupol:
superseded by #4031

PR #3887, opened 2018-07-18
FIX #3849: Add vnd.api+json to LoggingFeature

arjvik:
Original Issue:

> It would be nice if the [vnd.api+json](http://www.iana.org/assignments/media-types/application/vnd.api%2Bjson) media type, used by the [JSON API](http://jsonapi.org/) standard, could be added to the media types logged by the LoggingFeature.Verbosity PAYLOAD_TEXT.

Signed-off-by: Arjun Vikram <arjvik@gmail.com>

PR #4030, opened 2019-01-17
Break from looping over FormDataBodyParts once first match is found.

MuffinTheMan:
Since this is supposed to return the first result found, it seems appropriate to break from the `for` loop once we have a `result`.

### DIFF
--- a/media/multipart/src/main/java/org/glassfish/jersey/media/multipart/FormDataMultiPart.java
+++ b/media/multipart/src/main/java/org/glassfish/jersey/media/multipart/FormDataMultiPart.java
@@ -107,6 +107,7 @@ public class FormDataMultiPart extends MultiPart {
             }
             if (name.equals(((FormDataBodyPart) bodyPart).getName())) {
                 result = (FormDataBodyPart) bodyPart;
+                break; // Break to return first result found.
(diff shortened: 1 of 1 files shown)


jansupol:
Please see [Missing-ECA,-Missing-Sign-off](https://github.com/eclipse-ee4j/jersey/wiki/Missing-ECA,-Missing-Sign-off)

MuffinTheMan:
I signed the ECA prior to submitting my PR. I used my primary GitHub email address for the ECA and also configured the email for my commit to the same (all prior to my commit and subsequent PR). I validated my email address just now with the ECA validator, and it says I have a signed agreement.

jansupol:
@MuffinTheMan that sounds all good. The only issue causing this that comes to my mind is that you did not commit with **-s**. At least it looks like that, when I look to your commit, I do not see the "Singned-off-by".  Compare for instance with https://github.com/eclipse-ee4j/jersey/pull/3887/commits.

MuffinTheMan:
I ammended my commit and signed it off (then ended up merging in master for some odd reason despite there being no changes...). Will this work, or do I need to start from scratch so every commit is signed off?

PR #4032, opened 2019-01-18
Updated versions of dependencies

jansupol:
Change-Id: I50f311ae10caa02e967a4fb3117eadc98b01afcb

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1992,7 +1992,7 @@
         <kryo.version>4.0.1</kryo.version>
         <mimepull.version>1.9.11</mimepull.version>
         <mockito.version>1.10.19</mockito.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -1992,7 +1992,7 @@
         <kryo.version>4.0.1</kryo.version>
         <mimepull.version>1.9.11</mimepull.version>
         <mockito.version>1.10.19</mockito.version>
-        <moxy.version>2.7.3</moxy.version>
+        <moxy.version>2.7.4-RC2</moxy.version>
```
lukasj: you should use 2.7.4 instead

jansupol: done

PR #4055, opened 2019-02-09
Enable @ConstrainedTo on Features

jansupol:

### DIFF
--- a/core-common/src/main/java/org/glassfish/jersey/model/internal/CommonConfig.java
+++ b/core-common/src/main/java/org/glassfish/jersey/model/internal/CommonConfig.java
@@ -669,6 +669,13 @@ public class CommonConfig implements FeatureContext, ExtendedConfig {
                 continue;
             }
 
+            final RuntimeType runtimeTypeConstraint = registration.getFeatureRuntimeType();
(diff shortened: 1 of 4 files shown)


core-common/src/main/java/org/glassfish/jersey/model/internal/CommonConfig.java
```diff
@@ -674,13 +690,17 @@ private void configureFeatures(InjectionManager injectionManager,
...
             }
-            final boolean success = feature.configure(featureContextWrapper);
 
-            if (success) {
-                processed.add(registration);
-                configureFeatures(injectionManager, processed, resetRegistrations(), managedObjectsFinalizer);
-                enabledFeatureClasses.add(registration.getFeatureClass());
-                enabledFeatures.add(feature);
+            final RuntimeType runtimeTypeConstraint = registration.getFeatureRuntimeType();
+            if (runtimeTypeConstraint == null || type.equals(runtimeTypeConstraint)) {
```
tomas-langer: This check should be performed before line 674, so we do not create an injected instance of a Feature that should not exist in this type of runtime.

PR #4057, opened 2019-02-11
Add dependency on jakarta.activation-api to core-common

mszabo-wikia:
The class `org.glassfish.jersey.message.internal.DataSourceProvider` in the `core-common` submodule references the `javax.activation.DataSource` interface. This results in a compilation error on Java 11, which no longer provides the activation API built-in. This PR adds `jakarta.activation-api` as a dependency to the module to resolve the issue.

(Sidenote: as of this change, `core-common` compilation still fails due to use of `sun.misc.Contended` in `org.glassfish.jersey.internal.jsr166.SubmissionPublisher`—there does not seem to be a backwards compatible migration path there.)

### DIFF
--- a/core-common/pom.xml
+++ b/core-common/pom.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
-    Copyright (c) 2010, 2018 Oracle and/or its affiliates. All rights reserved.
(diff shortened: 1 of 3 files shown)


mszabo-wikia:
👍 done in b64d80e58ec523e02ce218be2ed015428794fe60

PR #4059, opened 2019-02-12
Travis CI is caching local Maven Repository

mkarg:
Improves Travis CI performance by loading one big folder instead of running a lot of single dependency downloads (in my tests, performance was about doubled). :-)

Unfortunately EF has no similar solution at hand for Jenkins (at least right now off-the-box). :-(

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,8 @@ jdk:
 cache:
   directories:
     - .autoconf
+    - $HOME/.m2
+
 env:
   matrix:
     - TEST_SET="-Ptravis_e2e_skip"


senivam:
I was not giving .m2 repository to cache on purpose, because this ensures the full clear build from scratch, including re-download of *jakarta* and *org.glassfish* dependencies and cleaning of the *org.glassfish.jersey* artifacts before build itself. However if you exclude *jakarta*, *org.glassfish* and *org.eclipse* artifacts from caching that should result in (almost) the same clear build as if no cache is used.  

And looking at timings for cached and non-cached builds we have those results:

**copyright-check** profile on JDK8 and OpenJDK8 are identical (slight difference in seconds)
**travis_e2e_skip** profile on JDK8 and OpenJDK8 are identical (slight difference in seconds)
**travis_e2e** profile on OpenJDK8 is one minute longer for the cached version and three minutes faster for the same cached version on JDK8

mkarg:
That timing results are really strange. I assume that measuring timing on a cloud system is rather hard. Anyways, if it is not really beneficial to Jersey, we can simply close this issue?

senivam:
Thank you for the feedback. OK, I will close this and it could be re-opened any time if .m2 cache considered to be required for any reason. Thank you for your efforts.

PR #4064, opened 2019-02-19
Fix usage of sun.misc.Contended in core-common

mszabo-wikia:
The `sun.misc.Contended` annotation has been moved to `jdk.internal.vm.annotation` package for Java 11, which breaks compilation as `org.glassfish.jersey.internal.jsr166.SubmissionPublisher.BufferedSubscription` uses it. This PR adds two dummy `@Contended` annotations that allow the project to be compiled on Java >8; at runtime, the JDK will pick up whichever of the `@Contended` annotations is provided by the JDK.

An alternative may be to simply remove the `@Contended` usage from `org.glassfish.jersey.internal.jsr166.SubmissionPublisher.BufferedSubscription`.  Because the class is not a system class, the annotation will only ever take effect if Jersey users explicitly opt-in by running their JVM with the `-XX:-RestrictContended` flag.

### DIFF
--- a/core-common/src/main/java/jdk/internal/vm/annotation/Contended.java
+++ b/core-common/src/main/java/jdk/internal/vm/annotation/Contended.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright (c) 2019 Oracle and/or its affiliates. All rights reserved.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0, which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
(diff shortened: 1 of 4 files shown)


senivam:
Thank you, @mszabo-wikia, for such an awesome activity in helping as to make Jersey buildable on JDK11, it is really appreciated. Your previous PRs on this issue are really brilliant and helpful.  

For now we are actively working on Jersey's ability to be built on JDK9+ and it will soon be implemented (+ Travis CI job will be modified to build Jersey on JDK9+).

The current issue with Contended annotation you are solving is being actually solved by us as well, but actually in very different way. Further more the change in this PR introduces OSGI error (which results in tests failures). 

So I would prefer to close this PR (because of described above) and we will introduce fixes for JDK9+ compatibility very soon.

Thank you.

mszabo-wikia:
Thanks for the update @senivam ! Looking forward to it :)

PR #4067, opened 2019-02-22
Build core-common on JDK 11

jansupol:
Possible to create a multi-release jar when first built with JDK 8 and then with JDK11 (without clean)


jansupol:
Created CQ 19042

jansupol:
Created CQ 19042 has been approved

core-common/src/main/jsr166/org/glassfish/jersey/internal/jsr166/Flow.java
```diff
@@ -6,6 +6,8 @@
 
 package org.glassfish.jersey.internal.jsr166;
 
+import java.util.concurrent.Executor;
```
tomas-langer: Probably unused import

PR #4072, opened 2019-03-01
Add module-info to core-common

jansupol:

### DIFF
--- a/core-common/pom.xml
+++ b/core-common/pom.xml
@@ -292,15 +292,23 @@
         </profile>
 
         <profile>
-            <id>jdk9+</id>
+            <id>jdk11+</id>
+            <!--
+                JDK 9 & 10 is unsupported (as well as <release>9</release>)
+                module-info for java.xml.bind is taken from JDK (lib/ct.sym/9-modules)
+                and it depends on java.activation which clashes with javax.activation
+            -->
             <activation>
-                <jdk>[9,12)</jdk>
+                <jdk>[11,)</jdk>
             </activation>
             <dependencies>
                 <dependency>
+                    <groupId>jakarta.xml.bind</groupId>
+                    <artifactId>jakarta.xml.bind-api</artifactId>
+                </dependency>
+                <dependency>
                     <groupId>com.sun.activation</groupId>
                     <artifactId>jakarta.activation</artifactId>
-                    <version>${jakarta.activation.version}</version>
                 </dependency>
             </dependencies>
             <build>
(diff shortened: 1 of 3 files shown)


core-common/pom.xml
```diff
@@ -564,6 +528,21 @@
                                     </resources>
                                 </configuration>
                             </execution>
+                            <execution>
+                                <id>copy-jdk9-classes-to-meta-inf</id>
+                                <phase>prepare-package</phase>
+                                <goals>
+                                    <goal>copy-resources</goal>
```
lukasj: Are you sure that sources jar contains these jdk "overrides"?
Did you consider using buildhelper's add-resources goal together with its targetfolder configuration option?

core-common/src/main/java/module-info.java
```diff
@@ -0,0 +1,56 @@
...
+  // requires osgi.resource.locator;
+
+  // Exports rather all, which corresponds to previous state without module-info
+  exports org.glassfish.jersey.internal.guava;
+  exports org.glassfish.jersey.internal.inject;
+  exports org.glassfish.jersey.internal.l10n;
+  exports org.glassfish.jersey.internal.sonar;
+  exports org.glassfish.jersey.internal.spi;
+  exports org.glassfish.jersey.internal.util;
+  exports org.glassfish.jersey.internal;
```
lukasj: it is usually not good idea to export internal packages. Wouldn't it be better to selectively export them on the case by case basis if there is no option to use standard way to access required functionality?

core-common/pom.xml
jansupol: Affirmative. Sources are copied by copy-jdkX-sources targets.

core-common/src/main/java/module-info.java
jansupol: +1 leaving module.info for next release

brandon1024:
I would really love to see these changes make it in at some point. Jersey does not play nicely with some IDEs on version 11 of the JDK.

core-common/src/main/java/module-info.java
```diff
@@ -0,0 +1,56 @@
...
+  exports org.glassfish.jersey.message;
+  exports org.glassfish.jersey.message.internal;
+  exports org.glassfish.jersey.model;
+  exports org.glassfish.jersey.model.internal;
+  exports org.glassfish.jersey.process;
+  exports org.glassfish.jersey.process.internal;
+  exports org.glassfish.jersey.spi;
+  exports org.glassfish.jersey.uri;
+  exports org.glassfish.jersey.uri.internal;
+
```
brandon1024, line 49: I believe `org.glassfish.jersey.server` may be missing?

senivam: server package is a part of the core-server module. core-common does not have it.

PR #4062, opened 2019-02-18
Support for Spring 5

ankurpathak:
Fixes: gh-3736

PR #4080, opened 2019-03-20
Change SpringComponentProvider to look for beans in current and parent application contexts

radarsh:
Signed-off-by: Adarsh Ramamurthy <adarsh.ramamurthy@gmail.com>


radarsh:
I still don't understand how `LocalizationMessages` is resolved as in my IDE, it results in a compilation error. If there is a set up step that needs to be followed, please include that in the contribution guide.

jansupol:
Relates to #4062

ankurpathak:
> I still don't understand how `LocalizationMessages` is resolved as in my IDE, it results in a compilation error. If there is a set up step that needs to be followed, please include that in the contribution guide.

Even I was getting the same error. Its auto generated source. You run mvn clean package inside spring4 artifact and you will find that file inside build directory target. You can repeat same for spring5 artifact also.

Issue #4082, opened 2019-03-21
JerseyClientBuilder modifies Map content during provider registration


Verdent:
When you try to register new provider with Map of contracts, it will get modified when there is non supported provider class present in the map.

Example:
Map<Class<?>, Integer> contracts = new HashMap<>();
contract.put(ClientRequestFilter.class, 500);
contract.put(SomeCustomClass.class, 501);

contracts.size();// -> returns 2
jerseyClientBuilder.register(MyProvider.class, contracts);
contracts.size();// -> returns 1

senivam:
That is documented behavior:

 /**
     * Create a contract provider for a given component class.
     *
     * @param componentClass  component class to create contract provider model for.
     * @param defaultPriority default component priority. If {@value ContractProvider#NO_PRIORITY},
     *                        the value from the component class {@link javax.annotation.Priority} annotation will be used
     *                        (if any).
     * @param contractMap     map of contracts and their binding priorities. If {@code null}, the contracts will
     *                        gathered by introspecting the component class. Content of the contract map
     *                        may be modified during the registration processing.
     * @param modelEnhancer   custom contract provider model enhancer.
     * @return contract provider model for the class.
     */
    private static ContractProvider modelFor(final Class<?> componentClass,
                                             final int defaultPriority,
                                             final Map<Class<?>, Integer> contractMap,
                                             final Inflector<ContractProvider.Builder, ContractProvider> modelEnhancer)

senivam:
Even though the behavior is documented it does not seem to be correct. Thus fix for it is introduced.

senivam:
merged, fixed

PR #4083, opened 2019-03-21
fix for #4082

senivam:
Fixes reported issue #4082 


senivam:
I was thinking about that and I'm not sure that change is really needed. However it seems to be really weird behavior when framework modifies input from client which is not expected to be modified. And the only reason Jersey does this which I've found is that it tries to avoid NPE when cloning NULL map from input. As a result I've handled possible NULL map in some different way.  

I've added tests which demonstrates 3 use-cases which can hapend - irrelevant input map (checked and map is still the same). Irrelevant immutable Map (same as previous but the map is explicitly immutable so if there is an attempt to modify it an exception is thrown), NULL input map - works as before (original functionality is kept).

PR #4086, opened 2019-03-26
MP rest client 1.2.1 implementation

Verdent:
MP rest client implementation


smillidge:
This PR depends on Helidon that should be removed I think only MicroProfile dependencies should be added.

Verdent:
Hi @smillidge would it be acceptable from you point of view to move all helidon dependecies to the test scope?

Verdent:
Helidon removed from this implementation.

ext/mp-rest-client/src/main/java/org/glassfish/jersey/restclient/BeanClassModel.java
```diff
@@ -0,0 +1,299 @@
...
+        /**
+         * Parses all {@link QueryParam} annotated fields from bean class.
+         *
+         * @return updated builder instance
+         */
+        Builder processQueryFields() {
+            Stream.of(beanClass.getDeclaredFields())
+                    .filter(field -> field.isAnnotationPresent(QueryParam.class))
+                    .forEach(field -> parameterModels.add(ParamModel.from(interfaceModel, field.getType(), field, -1)));
+            return this;
```
senivam: make private method 
Builder processFieldsByParameterClass(Class<?> parameterClass) {
            Stream.of(beanClass.getDeclaredFields())
                    .filter(field -> field.isAnnotationPresent(parameterClass))
                    .forEach(field -> parameterModels.add(ParamModel.from(interfaceModel, field.getType(), field, -1)));
            return this;
        }

and call that private method from those 5 process*Fields() with appropriate parameter

ext/mp-rest-client/src/main/java/org/glassfish/jersey/restclient/BeanParamModel.java
```diff
@@ -0,0 +1,89 @@
...
+        } else if (CookieParam.class.equals(annotationClass)) {
+            return beanClassModel.resolveCookies((Map<String, String>) requestPart, instance);
+        } else if (QueryParam.class.equals(annotationClass)) {
+            return beanClassModel.resolveQuery((Map<String, Object[]>) requestPart, instance);
+        } else if (MatrixParam.class.equals(annotationClass)) {
+            return beanClassModel.resolveMatrix((WebTarget) requestPart, instance);
+        } else if (FormParam.class.equals(annotationClass)) {
+            return beanClassModel.resolveForm((Form) requestPart, instance);
+        }
+        throw new UnsupportedOperationException(annotationClass.getName() + " is not supported!");
```
senivam: I would somehow optimize the  whole method... not looks very good.

ext/mp-rest-client/src/main/java/org/glassfish/jersey/restclient/ParamModel.java
```diff
@@ -0,0 +1,303 @@
...
+                return new BeanParamModel(this);
+            } else if (cookieParamName != null) {
+                return new CookieParamModel(this);
+            } else if (queryParamName != null) {
+                return new QueryParamModel(this);
+            } else if (matrixParamName != null) {
+                return new MatrixParamModel(this);
+            } else if (formParamName != null) {
+                return new FormParamModel(this);
+            }
```
senivam: very nice structure, don't you think?

ext/mp-rest-client/pom.xml
```diff
@@ -0,0 +1,175 @@
...
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>project</artifactId>
+        <groupId>org.glassfish.jersey.ext</groupId>
+        <version>2.29-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>mp-rest-client</artifactId>
```
pdudits: The convention in the repo appears to be to start artifact names with `jersey-`

ext/mp-rest-client/pom.xml
```diff
@@ -0,0 +1,175 @@
...
+        </dependency>
+        <dependency>
+            <groupId>javax.enterprise</groupId>
+            <artifactId>cdi-api</artifactId>
+            <version>2.0</version>
+        </dependency>
+        <dependency>
+            <groupId>org.jboss.weld.se</groupId>
+            <artifactId>weld-se-core</artifactId>
+            <version>3.0.3.Final</version>
```
pdudits: This shouldn't be a compile-time dependency -- rather test or provided

ext/mp-rest-client/pom.xml
```diff
@@ -0,0 +1,175 @@
...
+            <groupId>io.opentracing</groupId>
+            <artifactId>opentracing-noop</artifactId>
+            <version>0.30.0</version>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
```
pdudits: The artifact should also be an OSGi bundle, simply adding `maven-bundle-plugin` like in others will work fine.

ext/mp-rest-client/src/main/java/org/glassfish/jersey/restclient/InterfaceUtil.java
```diff
@@ -0,0 +1,135 @@
...
+        }
+        return getComputeMethod(iClass, methodName);
+    }
+
+    private static List<Method> getStaticComputeMethod(String methodName) {
+        int lastIndex = methodName.lastIndexOf(".");
+        String className = methodName.substring(0, lastIndex);
+        String staticMethodName = methodName.substring(lastIndex + 1);
+        try {
+            Class<?> classWithStaticMethod = Class.forName(className);
```
pdudits: This assumes, that application code is reachable from REST client's classloader and that is usually not true in an application server.

Usual idiom I found is
```
Class<?> clazz = AccessController.doPrivileged(ReflectionHelper.classForNamePA(className));
```
That uses thread's context class loader and returns `null` (rather than throwing) when class is not found.

ext/mp-rest-client/src/main/java/org/glassfish/jersey/restclient/RestClientBuilderImpl.java
```diff
@@ -0,0 +1,331 @@
...
+        return (T) Proxy.newProxyInstance(interfaceClass.getClassLoader(),
+                                          new Class[] {interfaceClass},
+                                          new ProxyInvocationHandler(webTarget, restClientModel)
+        );
+    }
+
+    private void processConfigProviders(Class<?> restClientInterface, String[] providerArray) {
+        for (String provider : providerArray) {
+            try {
+                Class<?> providerClass = Class.forName(provider);
```
pdudits: Same as above

ext/mp-rest-client/pom.xml
Verdent: maven-bundle-plugin added to pom.

ext/mp-rest-client/pom.xml
Verdent: this dependency is now in "test" scope

ext/mp-rest-client/pom.xml
Verdent: Name updated.

ext/mp-rest-client/src/main/java/org/glassfish/jersey/restclient/InterfaceUtil.java
Verdent: OK I have used suggested doPrivileged method and added some null checks to validate returned value.

ext/mp-rest-client/src/main/java/org/glassfish/jersey/restclient/RestClientBuilderImpl.java
Verdent: See above.

ext/mp-rest-client/src/main/java/org/glassfish/jersey/restclient/BeanClassModel.java
Verdent: I have updated this builder as you suggested.

ext/mp-rest-client/src/main/java/org/glassfish/jersey/restclient/RestClientBuilderImpl.java
```diff
@@ -0,0 +1,332 @@
...
+        jerseyClientBuilder.executorService(new ExecutorServiceWrapper(executorService.get(), asyncInterceptors));
+
+        RestClientModel restClientModel = RestClientModel.from(interfaceClass,
+                                                               responseExceptionMappers,
+                                                               paramConverterProviders,
+                                                               asyncInterceptors);
+
+
+        JerseyClient client = jerseyClientBuilder.build();
+        client.preInitialize();
```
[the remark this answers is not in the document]
Verdent: Well yes, that might be possible approach, but the problem is, what happens if the implementation is not `Initializable`. I would just stick with this approach for the time being before the better solution is found :-) 

The other thing is that I had to make a change in `JerseyWebTarget` to be able to obtain `InjectionManager` which is used for parameter converter invocation. To be fair I do not know about any better way how to get it. Due to that I have another implementation-specific part. 

Any suggestions? :-)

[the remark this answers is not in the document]
pdudits: For implementation-specifc part I was only referring to Builder and direct instantiation of it, and I think it is safe not to initialize it if it doesn't offer the interface, but let environment choose ClientBuilder customizations.

As far as `InjectionManager` go, I don't see enough into Jersey internals yet, but available APIs suggest (`InjectionManagerClientProvider`), that you should be in process of executing the request to obtain it. But as I understand it, that's too late for this usage, as it serves for constructing the request, so using `Supplier<InjectionManager>` to lazily load at request time  would not help. 

You might need to seek help with core committers whether they see other solution than opening up the encapsulation in web target. In Payara's implementation of client 1.1, we ended up exposing the injection manager from `ClientRuntime`, so very likely we're missing some API here for this use case.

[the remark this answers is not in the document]
pdudits: For our usage it is quite important to be able to use `ServiceLoader` for building the client. So I managed to find this workaround for obtaining injection manager, without resorting to opening up classes and `preInitialize`:

````java
public class JerseyClientInjectionManagerExposingTest {
    static class InjectionManagerExposer implements Feature {
        InjectionManager injectionManager;

        @Override
        public boolean configure(FeatureContext context) {
            if (context instanceof InjectionManagerSupplier) {
                this.injectionManager = ((InjectionManagerSupplier) context).getInjectionManager();
                return true;
            } else {
                return false;
            }
        }
    }

    @Test
    public void injectionManagerObtainable() {
        InjectionManagerExposer exposer = new InjectionManagerExposer();
        Client client = ClientBuilder.newBuilder().build();
        WebTarget target = client.target("http://localhost").register(exposer);

        // building an invocation also forces initialization without resorting to preInitialize
        target.request().buildGet();

        assertNotNull(exposer.injectionManager);
    }
}
````

File: ext/mp-rest-client/src/main/java/org/glassfish/jersey/restclient/BeanClassModel.java [hunk 10]
```diff
@@ -0,0 +1,300 @@
...
+                    parameterModel.handleParameter(form,
+                                                   FormParam.class,
+                                                   resolveValueFromField(field, instance));
+                });
+        return form;
+    }
+
+    private Object resolveValueFromField(Field field, Object instance) {
+        try {
+            Object toReturn;
```
jansupol: 1. Do not force field.setAccessible() if field.getAccessible()
2. Use new PrivilegedAction<Object>() {
            @Override
            public Object run() {
                return field.get(instance);
            }
        };

ext/mp-rest-client/src/test/java/org/glassfish/jersey/restclient/RestClientModelTest.java
```diff
@@ -0,0 +1,46 @@
...
+        enable(TestProperties.LOG_TRAFFIC);
+        return new ResourceConfig(ApplicationResourceImpl.class);
+    }
+
+    @Test
+    public void testGetIt() throws URISyntaxException {
+        ApplicationResource app = RestClientBuilder.newBuilder()
+                .baseUri(new URI("http://localhost:9998"))
+                .build(ApplicationResource.class);
+        System.out.println(app.getValue());
```
jansupol: An asseriont is supposed to be here?

core-client/src/main/java/org/glassfish/jersey/client/JerseyWebTarget.java
```diff
@@ -26,6 +26,7 @@
 import javax.ws.rs.core.UriBuilder;
```
jansupol: Update copyright year

ext/mp-rest-client/src/main/java/org/glassfish/jersey/restclient/ParamModel.java
```diff
@@ -0,0 +1,310 @@
...
+        }
+
+        /**
+         * Header parameter name.
+         *
+         * @param headerParam {@link HeaderParam} annotation
+         * @return updated Builder instance
+         */
+        Builder headerParamName(HeaderParam headerParam) {
+            this.headerParamName = headerParam == null ? null : headerParam.value();
```
jansupol: Why `headerParamName`, `pathParamName`, but `beanParam`, `cookieParam`....?

ext/mp-rest-client/src/main/java/org/glassfish/jersey/restclient/ReflectionUtil.java
```diff
@@ -0,0 +1,53 @@
...
+package org.glassfish.jersey.restclient;
+
+import java.lang.invoke.MethodHandles;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Proxy;
+
+/**
+ * Created by David Kral.
+ */
+class ReflectionUtil {
```
jansupol: AccessController.doPrivileged...

jansupol:
Please add this module to the `ext/microprofile/restclient` folder. 
Please add this to `org.glassfish.jersey.microprofile.restclient` package.
It seems we are getting more microprofile specific code.

ext/microprofile/mp-rest-client/src/main/java/org/glassfish/jersey/microprofile/restclient/RestClientExtension.java
```diff
@@ -0,0 +1,134 @@
...
+        }
+    }
+
+    /**
+     * Iterates over all {@link ProcessInjectionPoint} to find only those annotated
+     * with {@link RestClient} and configures their proper injection.
+     *
+     * @param pip processed injection point
+     */
+    public void collectClientProducer(@Observes ProcessInjectionPoint<?, ?> pip) {
```
pdudits: I don't understand the need for this transformation here.

In my opinion, without it, the spec is also satisfied (TCK pass), the code is simpler, and also CDI error messages are more reasonable when user doesn't annotate the interface `@RegisterRestClient` or produces conflicting `@RestClient` qualified bean.

pdudits: I solved it in the meantime and will provide PR on your PR (let's see how that works).

Thanks for incorporating my suggestions, I was about to post those as well ;)

jansupol:
[CQ #19756](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=19756) for SmallRye seems to have licensing issues, as it incorrectly refers to LGPL. Possibly, we need to use some other MP config implementation till SmallRye is released with the fixed license to be able to put this to 2.29

ext/microprofile/mp-rest-client/src/main/java/org/glassfish/jersey/microprofile/restclient/ConfigWrapper.java
```diff
@@ -0,0 +1,91 @@
+package org.glassfish.jersey.microprofile.restclient;
```
senivam: what about license header here?

ext/microprofile/mp-rest-client/src/main/java/org/glassfish/jersey/microprofile/restclient/ExecutorServiceWrapper.java
```diff
@@ -0,0 +1,118 @@
+package org.glassfish.jersey.microprofile.restclient;
```
senivam: license header?

ext/microprofile/mp-rest-client/src/main/java/org/glassfish/jersey/microprofile/restclient/MethodModel.java
```diff
@@ -0,0 +1,668 @@
...
+
+        /**
+         * Return type of the method.
+         *
+         * @param returnType Method return type
+         * @return updated Builder instance
+         */
+        Builder returnType(Type returnType) {
+            if (returnType instanceof ParameterizedType) {
+                this.returnType = (Class<?>) ((ParameterizedType) returnType).getActualTypeArguments()[0];
```
pdudits: You were probably thinking of `CompletionStage<T>` here.
But this logic is incorrect for anything else, especially lists and maps. I think TCK misses few very essential use cases ;)

core-client/src/main/java/org/glassfish/jersey/client/JerseyWebTarget.java
```diff
@@ -26,6 +26,7 @@
...
 
 import org.glassfish.jersey.internal.guava.Preconditions;
+import org.glassfish.jersey.internal.inject.InjectionManager;
```
jansupol: This does not seem to be needed any longer

ext/microprofile/mp-rest-client/src/test/resources/server.policy
```diff
@@ -0,0 +1,78 @@
...
+ * Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception, which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ */
+
+grant {
+  permission java.util.PropertyPermission "idea.launcher.bin.path", "read";
+  permission java.lang.RuntimePermission "loadLibrary.C:\\Program Files\\JetBrains\\IntelliJ IDEA 2019.1\\bin\\breakgen64.dll";
```
jansupol: Not this one

ext/microprofile/mp-rest-client/src/test/resources/server.policy
```diff
@@ -0,0 +1,78 @@
...
+  permission java.lang.RuntimePermission "accessDeclaredMembers";
+  permission java.lang.RuntimePermission "getClassLoader";
+  permission java.lang.RuntimePermission "createClassLoader";
+  permission java.lang.RuntimePermission "setContextClassLoader";
+  permission java.lang.RuntimePermission "getProtectionDomain";
+  permission java.lang.RuntimePermission "getenv.*";
+  permission java.lang.RuntimePermission "setIO";
+  permission java.io.FilePermission "C:\\Program Files\\JetBrains\\IntelliJ IDEA 2019.1\\bin\\breakgen64.dll", "read";
```
jansupol: Nor this one

ext/microprofile/mp-rest-client/pom.xml
```diff
@@ -0,0 +1,202 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.glassfish.jersey.media</groupId>
+            <artifactId>jersey-media-json-binding</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>jakarta.json</groupId>
+            <artifactId>jakarta.json-api</artifactId>
+            <version>1.1.5</version>
```
jansupol: Please use ${jsonp.ri.version}

ext/microprofile/mp-rest-client/pom.xml
```diff
@@ -0,0 +1,202 @@
...
+            <version>2.0</version>
+        </dependency>
+        <dependency>
+            <groupId>org.glassfish</groupId>
+            <artifactId>jsonp-jaxrs</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.jboss.weld.se</groupId>
+            <artifactId>weld-se-core</artifactId>
+            <version>3.0.3.Final</version>
```
jansupol: ${weld3.version} ?

ext/microprofile/mp-rest-client/src/main/java/org/glassfish/jersey/microprofile/restclient/MethodModel.java
Verdent: Oh... that was very good point :-) Thank you very much for this :-) It is fixed now.

ext/microprofile/mp-rest-client/src/main/java/org/glassfish/jersey/microprofile/restclient/RequestHeaderAutoDiscoverable.java
```diff
@@ -0,0 +1,36 @@
...
+import javax.ws.rs.RuntimeType;
+import javax.ws.rs.core.FeatureContext;
+
+import org.glassfish.jersey.internal.spi.AutoDiscoverable;
+
+/**
+ * Auto discoverable feature to bind into jersey runtime.
+ */
+@ConstrainedTo(RuntimeType.SERVER)
+public class RequestHeaderAutoDiscoverable implements AutoDiscoverable {
```
jansupol: ForcedAutoDiscoverable?

ext/microprofile/mp-rest-client/src/main/java/org/glassfish/jersey/microprofile/restclient/RestClientBuilderImpl.java
```diff
@@ -0,0 +1,389 @@
...
+import org.glassfish.jersey.internal.inject.InjectionManagerSupplier;
+import org.glassfish.jersey.internal.util.ReflectionHelper;
+
+/**
+ * Rest client builder implementation. Creates proxy instance of requested interface.
+ *
+ * @author David Kral
+ * @author Patrik Dudits
+ */
+public class RestClientBuilderImpl implements RestClientBuilder {
```
jansupol: package private?

ext/microprofile/mp-rest-client/src/main/java/org/glassfish/jersey/microprofile/restclient/DefaultResponseExceptionMapper.java
```diff
@@ -0,0 +1,39 @@
...
+import javax.ws.rs.core.Response;
+
+import org.eclipse.microprofile.rest.client.ext.ResponseExceptionMapper;
+
+/**
+ * Default {@link ResponseExceptionMapper} implementation
+ *
+ * @author David Kral
+ */
+public class DefaultResponseExceptionMapper implements ResponseExceptionMapper {
```
jansupol: package private?

ext/microprofile/mp-rest-client/src/main/java/org/glassfish/jersey/microprofile/restclient/HeadersContext.java
```diff
@@ -0,0 +1,103 @@
...
+import java.util.Optional;
+import java.util.function.Supplier;
+
+import javax.ws.rs.core.MultivaluedMap;
+
+/**
+ *
+ * @author David Kral
+ */
+public final class HeadersContext {
```
jansupol: package private?

ext/microprofile/mp-rest-client/src/main/java/org/glassfish/jersey/microprofile/restclient/InterfaceModel.java
```diff
@@ -0,0 +1,374 @@
...
+                               List<AsyncInvocationInterceptor> asyncInterceptors,
+                               InjectionManager injectionManager,
+                               BeanManager beanManager) {
+        return new Builder(restClientClass,
+                           responseExceptionMappers,
+                           paramConverterProviders,
+                           asyncInterceptors,
+                           injectionManager,
+                           beanManager)
+                .pathValue(restClientClass.getAnnotation(Path.class))
```
tomas-langer: Is this needed as separate builder methods? Shouldn't these be called within the `build` method and be private?

ext/microprofile/mp-rest-client/src/main/java/org/glassfish/jersey/microprofile/restclient/MethodModel.java
```diff
@@ -0,0 +1,667 @@
...
+    /**
+     * Processes interface method and creates new instance of the model.
+     *
+     * @param interfaceModel
+     * @param method
+     * @return
+     */
+    static MethodModel from(InterfaceModel interfaceModel, Method method) {
+        return new Builder(interfaceModel, method)
+                .returnType(method.getGenericReturnType())
```
tomas-langer: These methods again should be private and called from build

ext/microprofile/mp-rest-client/src/main/java/org/glassfish/jersey/microprofile/restclient/ParamModel.java
```diff
@@ -0,0 +1,310 @@
...
+     * @param interfaceModel model of the interface
+     * @param type annotated element type
+     * @param annotatedElement annotated element
+     * @param position position in method params
+     * @return new parameter instance
+     */
+    static ParamModel from(InterfaceModel interfaceModel, Type type, AnnotatedElement annotatedElement,
+                           Parameter parameter, int position) {
+        return new Builder(interfaceModel, type, annotatedElement, parameter)
+                .pathParamName(annotatedElement.getAnnotation(PathParam.class))
```
tomas-langer: And once again - methods can be internal to builder

ext/microprofile/mp-rest-client/src/main/java/org/glassfish/jersey/microprofile/restclient/ParamModel.java
```diff
@@ -0,0 +1,310 @@
...
+            return matrixParamName;
+        }
+
+        /**
+         * Creates new ParamModel instance.
+         *
+         * @return new instance
+         */
+        ParamModel build() {
+            if (pathParamName != null) {
```
tomas-langer: Could you create the param model when the parameter is set?
e.g.
```
Builder formParam(FormParam formParam) {
   if (null != formParam) {
    this.paramModel = new FormParamModel(formParam);
  } 
return this
 }
```
And then build method would just return the this.paramModel, and if null do the entity trick?

ext/microprofile/mp-rest-client/src/main/java/org/glassfish/jersey/microprofile/restclient/RestClientBuilderImpl.java
```diff
@@ -0,0 +1,389 @@
...
+
+        //Provider registration part
+        Object providersFromJerseyConfig = clientBuilder.getConfiguration()
+                .getProperty(interfaceClass.getName() + CONFIG_PROVIDERS);
+        if (providersFromJerseyConfig instanceof String && !((String) providersFromJerseyConfig).isEmpty()) {
+            String[] providerArray = ((String) providersFromJerseyConfig).split(PROVIDER_SEPARATOR);
+            processConfigProviders(interfaceClass, providerArray);
+        }
+        Optional<String> providersFromConfig = config.getOptionalValue(interfaceClass.getName() + CONFIG_PROVIDERS, String.class);
+        if (providersFromConfig.isPresent() && !providersFromConfig.get().isEmpty()) {
```
tomas-langer: This may be clearer if functional approach is used
```
providerFromConfig.ifPresent(providers -> {
   if (!providers.isEmpty()) {
   // do the magic
 }
}
```

tomas-langer: This method may benefit from separation into multiple methods

ext/microprofile/mp-rest-client/src/main/java/org/glassfish/jersey/microprofile/restclient/RestClientBuilderImpl.java
```diff
@@ -0,0 +1,389 @@
...
+
+
+        return (T) Proxy.newProxyInstance(interfaceClass.getClassLoader(),
+                                          new Class[] {interfaceClass},
+                                          new ProxyInvocationHandler(webTarget, restClientModel)
+        );
+    }
+
+    private BeanManager lookupBeanManager() {
+        Context initialContext = null;
```
tomas-langer: There is an existing `CdiUtil` in `jersey-cdi1x` module that has the same function

ext/microprofile/mp-rest-client/src/main/resources/META-INF/beans.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
```
tomas-langer: Remove this file - everything is done through CDI Extension. This may add beans not inteded to be added through discovery

ext/microprofile/mp-rest-client/pom.xml
```diff
@@ -0,0 +1,196 @@
...
+                            *
+                        </Import-Package>
+                    </instructions>
+                    <unpackBundle>true</unpackBundle>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+
+<!--    <properties>-->
```
tomas-langer: Should this not be uncommented? Otherwise the `server.policy` is unused.

jansupol:
Added CQs: 
 - ~~[Arquillian TestNG Container Version: 1.4.1.Final ](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=20097)            CQ 20097~~
 - [Arquillian Container SPI Version: 1.4.1.Final ](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=20098)                   CQ 20098
 - ~~[Arquillian Weld Embedded Version: 2.0.1.Final](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=20099)               CQ 20099~~
 - [weld-se-core Version: 3.0.3.Final  ](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=20104)                                   CQ 20104
 - ~~[wiremock Version: 2.21.0    ](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=20106)                                             CQ 20106~~

ext/microprofile/mp-rest-client/src/main/java/org/glassfish/jersey/microprofile/restclient/RestClientBuilderImpl.java
```diff
@@ -0,0 +1,389 @@
...
+        if (isSupportedCustomProvider(aClass)) {
+            register(ReflectionUtil.createInstance(aClass));
+        } else {
+            clientBuilder.register(aClass);
+        }
+        return this;
+    }
+
+    @Override
+    public RestClientBuilder register(Class<?> aClass, int i) {
```
jansupol: Why did you obscure the meaning of the method arguments? `register​(java.lang.Class<?> componentClass,
           int priority)`

ext/microprofile/mp-rest-client/src/main/java/org/glassfish/jersey/microprofile/restclient/RestClientBuilderImpl.java
```diff
@@ -0,0 +1,389 @@
...
+        if (isSupportedCustomProvider(aClass)) {
+            register(ReflectionUtil.createInstance(aClass), i);
+        } else {
+            clientBuilder.register(aClass, i);
+        }
+        return this;
+    }
+
+    @Override
+    public RestClientBuilder register(Class<?> aClass, Class<?>... classes) {
```
jansupol: `register​(java.lang.Class<?> componentClass,
           java.lang.Class<?>... contracts)` and so on

jansupol:
_Arquillian Container SPI Version: 1.4.1.Final_ contains LGPL licensed files for which we may have an issue to get a CQ for. Filed [a bug](https://github.com/arquillian/arquillian-core/issues/203) to make sure which license Arqullian has.

Issue #3873, opened 2018-06-29
Fine grained configuration of Jetty HttpClient in jersey-jetty-connector


cen1:
Jetty HttpClient exposes plenty of configuration options but (almost) none of these are configurable via `jersey-jetty-connector`. Specifically, I'd like to configure `requestBufferSize` to enable support for calling large URLs with the client.

Additionally, since API 2.5 `ClientConfig` now accepts `JettyConnectorProvider` instead of `JettyConnector` and I see no way to pass configuration to the `JettyConnector`.

Before:
```
 ClientConfig config = new ClientConfig();
 Connector connector = new JettyConnector(c); //you can pass some config here
 config.connector(connector);
 Client client = ClientBuilder.newClient(config);
```
API 2.5:
```
ClientConfig config = new ClientConfig();
config.connectorProvider(new JettyConnectorProvider()); //how to configure connector???
Client client = ClientBuilder.newClient(config);
```
Am I just missing something blatantly obvious here?

cen1:
Proposed patch: https://github.com/cen1/jersey/commit/34f43ce252ee17891f1a7ec82992e2edc060490c

I finally figured out I can set these properties via `client.propery()`. Are the other ways?

jansupol:
See [org.glassfish.jersey.jetty.connector.UnderlyingHttpClientAccessTest](https://github.com/eclipse-ee4j/jersey/blob/master/connectors/jetty-connector/src/test/java/org/glassfish/jersey/jetty/connector/UnderlyingHttpClientAccessTest.java) for a way to access Jetty `HttpClient`. Does it help?

jansupol:
Hm...`HttpClient` has `stop()` and `start()` start that can workaround the issue....but I agree its not a nice solution

zmanji:
I have run into a similar issue. I would like to configure the underlying Jetty client to use the [HttpClientTransportOverHTTP2](https://www.eclipse.org/jetty/documentation/9.3.x/http-client-transport.html#_http_2_transport) transport to support HTTP2. However the only way to set the transport is to pass in one via the constructor:
```
    public HttpClient(HttpClientTransport transport, SslContextFactory sslContextFactory)
    {
        this.transport = transport;
        this.sslContextFactory = sslContextFactory;
    }
```

Would it be possible to pass in an `HttpClient` instance into the `JettyConnectorProvider` constructor?

cen1:
[...]

PR #4089, opened 2019-04-01
Allow for providing Jetty HttpClient to Jetty Connector

jansupol:

### DIFF
--- a/connectors/jetty-connector/src/main/java/org/glassfish/jersey/jetty/connector/JettyConnector.java
+++ b/connectors/jetty-connector/src/main/java/org/glassfish/jersey/jetty/connector/JettyConnector.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2013, 2018 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2013, 2019 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
@@ -28,6 +28,7 @@ import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Future;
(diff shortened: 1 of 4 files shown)


jansupol:
Fixes #3873

ranarula:
Can something like this be done for Apache Connector as well. 
I currently need a functionality as described in #3586 

Access to `HttpClientBuilder` or a mechanism to provide a custom `org.apache.http.client.HttpClient` will be helpful

connectors/jetty-connector/src/main/java/org/glassfish/jersey/jetty/connector/JettyConnector.java
```diff
@@ -136,18 +137,28 @@
      * @param config client configuration.
      */
     JettyConnector(final Client jaxrsClient, final Configuration config) {
-        final SSLContext sslContext = jaxrsClient.getSslContext();
-        final SslContextFactory sslContextFactory = new SslContextFactory();
-        sslContextFactory.setSslContext(sslContext);
+        HttpClient httpClient = null;
+        if (config.isRegistered(JettyHttpClientSupplier.class)) {
```
jlorenzen: @jansupol Seems like you should be using the interface `JettyHttpClientContract` here instead of the implementation?

connectors/jetty-connector/src/main/java/org/glassfish/jersey/jetty/connector/JettyConnector.java
```diff
@@ -136,18 +137,28 @@
...
+            Optional<Object> contract = config.getInstances().stream()
+                    .filter(a-> JettyHttpClientSupplier.class.isInstance(a)).findFirst();
+            if (contract.isPresent()) {
+                httpClient = ((JettyHttpClientSupplier) contract.get()).getHttpClient();
```
jlorenzen: @jansupol How big of a deal is it that a user predefined `HttpClient` won't have access to set the `jaxrsClient.getSslContext`?

jlorenzen: For example, I'm looking to create similar changes for the Apache Http Client and it's currently doing the following:
```java
final SSLContext sslContext = client.getSslContext();
final HttpClientBuilder clientBuilder = HttpClientBuilder.create();
clientBuilder.setSslcontext(sslContext);
```

But it the user provides a predefined `HttpClient`, the way you are currently doing it in this PR, the user won't have access to the `sslContext`. I can think of ways we could provide that but we'd have to change some things up if we felt it was necessary. Not sure how important that is.

connectors/jetty-connector/src/main/java/org/glassfish/jersey/jetty/connector/JettyConnector.java
jansupol: No, it is JettyHttpClientSupplier that is being registered and not the JettyHttpClientContract.

connectors/jetty-connector/src/main/java/org/glassfish/jersey/jetty/connector/JettyConnector.java
jansupol: Well, if the Jetty HttpClient is provided, then the `sslContext` needs to be set on the provided HttpClient rather then on the JAX-RS `Client`. Of course, allowing the HttpClient to be provided is dangerous, since it can be misconfigured. But since the community seems to wish to have the possibility to configure the client as needed, I think Jersey should allow it.

jansupol: @jlorenzen I am not sure I did understand the comment about Apache client correctly, but would it help to provide the `HttpClientBuilder` rather than `ClosableHttpClient`?

PR #4090, opened 2019-04-02
Header utils header value update

Verdent:
Handling multiple header values updated to resolve single header value as simple String, otherwise surround all of the header values with [ ].


Verdent:
The problem was in my code. Closing PR.

PR #4091, opened 2019-04-04
Put Parameter class in core-common, to be used by both client and server

jansupol:


jansupol:
This is a proposal of how it would be possible to move the logic of Parameter to Core-Common. There is not a good solution, as Parameter ought to stay in core-server, as it is part of the public API. Since most methods are static and static methods are not virtual, it is hard to make the inheritance work.  The options are as follows:
1) Duplicate the code, maintain only the part in common, deprecate the server. 
2) Move package server (and the two classes - Parameter and AnnotatedMethod) to core-common. Not the cleanest either.
3) Some solution like this

PR #4094, opened 2019-04-09
Update JMockit version to 1.44 to support JDK 11

jansupol:
JMockit needs -javavagent since 1.42
JMockit does not have @RunWith(JMockit) since 1.43
JMockit does not support java.util mocks since 1.45


jansupol:
[CQ 16270](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=16270) covers this version of JMockit

jansupol:
This [bug] (https://github.com/jmockit/jmockit1/issues/598) could be similar to issue with `JarInputStream`  tests that are not mockable by JMockit since 1.45

Issue #4099, opened 2019-04-13
@Priority is not always picked up correctly for JAX-RS providers


chkal:
There are still some cases where ordering of JAX-RS providers via `@Priority` doesn't work correctly. In my case there are two `ExceptionMapper` instances for the same exception type but with different priorities. Jersey fails to pick up the `@Priority`, which results in a default priority of `Priorities.USER` for both mappers and therefore in random ordering at runtime.

IMO the problem actually occurs for all JAX-RS provider types. It occurs with Jersey 2.28. See the following analysis for details:

Custom providers are looked up in the `Providers` class like this:

https://github.com/eclipse-ee4j/jersey/blob/895c293e366afce166e508ae9db23f9e4b9b5f80/core-common/src/main/java/org/glassfish/jersey/internal/inject/Providers.java#L287-L290

The `getServiceHolders` method gets a `Comparator` which uses `Providers.getPriority()` to order the providers according to the `@Priority` annotation.

https://github.com/eclipse-ee4j/jersey/blob/895c293e366afce166e508ae9db23f9e4b9b5f80/core-common/src/main/java/org/glassfish/jersey/internal/inject/Providers.java#L363-L371

Everything is fine so far. 

The problem is that `Providers.getPriority()` isn't always called with the actual provider implementation class but sometimes gets an instance of `org.glassfish.jersey.inject.hk2.InstanceSupplierFactoryBridge`. This means that Jersey sometimes looks for the `@Priority` on the wrong class and is ignoring the `@Priority` on the actual implementation class.

![image](https://user-images.githubusercontent.com/159245/56075775-31b07100-5dc8-11e9-8f23-32560877e143.png)

The `Providers` class seems to use `ServiceHolder.getImplementationClass()` to get the type of the provider class, which doesn't work correctly in all cases.

https://github.com/eclipse-ee4j/jersey/blob/895c293e366afce166e508ae9db23f9e4b9b5f80/core-common/src/main/java/org/glassfish/jersey/internal/inject/Providers.java#L313-L315

<del>My current guess is that it should use `ServiceHolder.getImplementation**Type**()` instead. But I'll need to verify that. I'll try to create a failing integration tests for this issue.-</del>

chkal:
Please see #4100 for a possible fix. Although I'm not completely sure if it is the best place for a fix.

chkal:
I created #4101 which contains an integration test which reproduces this issue. I hope this helps.

jansupol:
[...]

PR #4100, opened 2019-04-13
Create ServiceHolderImpl with correct service implementation type

chkal:
As described in #4099, `ServiceHolder#getImplementationClass()` sometimes doesn't return the actual implementation class of the service, but `org.glassfish.jersey.inject.hk2.InstanceSupplierFactoryBridge` instead. This leads to the situation that Jersey looks for `@Priority` annotations on the wrong type (detailed description in #4099).

This change fixes the underlying problem and ordering via `@Priority` works fine now. Unforunately I wasn't able to reproduce the issue in an integration test. I guess that's because HK2 behaves differently when deploying a real app to Glassfish.


chkal:
Sorry, this first to fix the issue broke some tests. I guess it won't work this way. Sorry for the noise.

PR #4101, opened 2019-04-14
Added integration test to reproduce #4099

chkal:
This pull request contains a new integration test which reproduces #4099. Please see #4099 for a detailed description of the problem.

Steps to reproduce:
  * Remove the `@Ignored` annotation in `PriorityAnnotationOnExceptionMappersTest`
  * Run the test a few times
  * You will see that the test randomly passes or fails

Setting a breakpoint here will show the root cause of the problem:

https://github.com/eclipse-ee4j/jersey/blob/895c293e366afce166e508ae9db23f9e4b9b5f80/core-common/src/main/java/org/glassfish/jersey/internal/inject/Providers.java#L363-L371

The method scans `InstanceSupplierFactoryBridge` instead of the actual implementation class for the `@Priority` annotation.

### DIFF
--- a/tests/integration/jersey-4099/pom.xml
+++ b/tests/integration/jersey-4099/pom.xml
@@ -1,0 +1,79 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright (c) 2019 Christian Kaltepoth. All rights reserved.
+
+    This program and the accompanying materials are made available under the
+    terms of the Eclipse Public License v. 2.0, which is available at
+    http://www.eclipse.org/legal/epl-2.0.
+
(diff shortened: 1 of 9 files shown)


chkal:
I updated the pull request and removed the `@Ignore` annotation on the test.

alawrenc:
I know I'm coming in way too late, but doesn't this integration test demonstrate exactly the _wrong_ behavior for exception mappers?
Per the jax-rs spec, post-request filters (including exception mappers by my understanding) are sorted in reverse order. https://docs.oracle.com/javaee/7/api/javax/ws/rs/Priorities.html
So a _higher_ priority value should take precedence in this case

PR #4103, opened 2019-04-17
Fix priority ordering

jansupol:


jansupol:
Fixes #4099.

senivam:
@chkal I suppose, it should be vice versa. As soon as #4101 is ignored for now and will be valid only after this PR is merged. So this goes first and then #4101 without @Ignore

chkal:
Ok, sure! I'll update #4101 and remove the `@Ignore` as soon as this PR is merged! Thanks a lot for your help.

jansupol:
@chkal Please remove `@Ignore`. Thank you for the very nice test case.

chkal:
I just updated #4101 as requested. I hope the test case was useful for reproducing the issue.

PR #4104, opened 2019-04-19
Add Support to set useSystemProperties

jlorenzen:
I'd like to gauge interest in a change like this. This gives users of the `ApacheConnectorProvider` class the ability to set `useSystemProperties` on the `HttpClientBuilder` class in `ApacheConnector`. This is achieved like the existing hooks by adding a new property to `ApacheClientProperties`.

Let me know if there is interest and I can include some javadocs and any necessary tests you guys recommend.

### DIFF
--- a/connectors/apache-connector/src/main/java/org/glassfish/jersey/apache/connector/ApacheClientProperties.java
+++ b/connectors/apache-connector/src/main/java/org/glassfish/jersey/apache/connector/ApacheClientProperties.java
@@ -134,6 +134,8 @@ public final class ApacheClientProperties {
      */
     public static final String RETRY_HANDLER = "jersey.config.apache.client.retryHandler";
 
+    public static final String USE_SYSTEM_PROPERTIES = "jersey.config.apache.client.useSystemProperties";
+
     /**
      * Get the value of the specified property.
      *
--- a/connectors/apache-connector/src/main/java/org/glassfish/jersey/apache/connector/ApacheConnector.java
+++ b/connectors/apache-connector/src/main/java/org/glassfish/jersey/apache/connector/ApacheConnector.java
@@ -280,6 +280,13 @@ class ApacheConnector implements Connector {
             this.cookieStore = null;
         }
         clientBuilder.setDefaultRequestConfig(requestConfig);
+
+        final boolean useSystemProperties = PropertiesHelper
+                .isProperty(config.getProperties(), ApacheClientProperties.USE_SYSTEM_PROPERTIES);
+        if (useSystemProperties) {
+            clientBuilder.useSystemProperties();
+        }
+
         this.client = clientBuilder.build();
     }
 


jlorenzen:
Also, what about removing this `useSystemProperties` parameter on the private `createConnectionManager` method?
https://github.com/eclipse-ee4j/jersey/blob/master/connectors/apache-connector/src/main/java/org/glassfish/jersey/apache/connector/ApacheConnector.java#L318
The only place that calls it hardcodes it to `false` so it doesn't seem very useful.

jlorenzen:
@jansupol Thoughts on this change?

ranarula:
I was wondering if a more generic approach can be used here as adopted in PR #4089 
I added a comment in there as well if we can use the same approach to provide custom Http client it might help with setting configuration on the client which are not yet supported. 

I was looking for a way to set `ServiceUnavailableRetryStrategy` on the client and it's not yet supported. Its also an open issue https://github.com/eclipse-ee4j/jersey/issues/3586 

Adding multiple config properties doesn't seem too scalable.

jansupol:
@ranarula I agree with the general approach, we plan to add it to the next release. The general approach, however, is not in a contradiction to this PR, we can have both for the users to use what they prefer.

jansupol:
@jlorenzen Couple of thoughts:
1) I am fine with adding a property to the Apache Connector to use system properties if preferred by users. 
2) Regarding `useSystemProperties` parameter, rather than removing that, I would prefer to define another property that would set that parameter to `true`. It would be necessary to name these two properties distinctively, not to confuse users what property to use.
3) A copyright year needs to updated, and a test should be added.

jlorenzen:
@ranarula @jansupol Yeah I tend to like the approach taken in #4089 because my plan after this PR was to create another PR that allowed me to use [CachingHttpClientBuilder](https://hc.apache.org/httpcomponents-client-ga/httpclient-cache/apidocs/org/apache/http/impl/client/cache/CachingHttpClientBuilder.html). After thinking about it some more, an approach like the one taken in #4089 would work for both of my requirements. So I think I'll close out this PR and perhaps create another that is similar to #4089. How does that sound?

ranarula:
@jansupol - Thanks for your thoughts. Having both options for the user will surely be super. If we wanna take that approach @jlorenzen can you please add the support for `ServiceUnavailableRetryStrategy` similar to `useSystemProperties` you added in this PR.

Based on what @jansupol mentioned, having this PR adds value.

jlorenzen:
@ranarula @jansupol So after doing quite a bit of experimenting, I don't think the approach taken in #4089 is the right approach for Apache Http Client. The main reason is because I think it would be too easy for a user's `HttpClient` to be configured differently than the corresponding `ApacheConnector`. For example, let's say the user creates an `HttpClient` with a `cookieStore`, but then forgets to set `ApacheClientProperties.DISABLE_COOKIES`. In this scenario the `ApacheConnector.cookieStore` ends up being set to `null` whereas the user's local instance of `HttpClient` is configured with a cookie store. I think the same is possible in the reverse (ie user creates an `HttpClient` without a cookieStore but ApacheConnector gets configured with one). The same thing could happen for the `ApacheConnector.requestConfig` instance variable.

Given all that I don't think it's worth adding support like #4089 for Apache Http Client. It could probably be done but not without the potential to create a lot of confusion and heartache for end users. If Apache Http Client's class `CloseableHttpClient` had methods that let you get the cookieStore and the requestConfig, then it might be possible, but unfortunately it doesn't. Also, if Apache Http Client's class `HttpClientBuilder` had methods to get the cookieStore and requestConfig then it might be possible as well, but alias it doesn't.

I also don't think adding more properties to `ApacheClientProperties` is the right way to do things either. Like @ranarula said earlier it just doesn't scale well and I think it points to an underlying issue. Unless @jansupol has some other ideas I think I'm going to stop trying to make these changes.

Issue #4110, opened 2019-04-24
Create a way to better configure Jersey


jansupol:
While there is a set of properties that can be used to configure Jersey (ClientProperties, ServerProperties, CommonProperties), the configuration is performed inside a java code.

There are ways to configure Jersey application from outside:
- Using the environment properties. This is not anything that should be done by default, but it also should be able to configure the application without modifying it
- Using the property files. One way is to use Config JSR/ MP Config implementation to provide this functionality. The other way is to use an implementation of the user provided property provider.

Minimize impact on permissions needed by SecurityManager to read properties (such as from the environment).

PR #4116, opened 2019-04-25
SPI Config API for Jersey

senivam:
Relates to #4110 

### DIFF
--- a/core-common/src/main/java/org/glassfish/jersey/CommonProperties.java
+++ b/core-common/src/main/java/org/glassfish/jersey/CommonProperties.java
@@ -217,6 +217,17 @@ public final class CommonProperties {
     public static final String OUTBOUND_CONTENT_LENGTH_BUFFER_SERVER = "jersey.config.server.contentLength.buffer";
 
     /**
+     * Property which allows (if true) default System properties configuration provider.
+     *
+     * Effective if there are no any external properties providers
+     *
+     * Shall be set (if used) in system properties.
+     * @since 2.29
+     */
+
+    public static final String ALLOW_SYSTEM_PROPERTIES_PROVIDER = "jersey.config.allowSystemPropertiesProvider";
+
+    /**
      * Prevent instantiation.
      */
     private CommonProperties() {
--- a/core-common/src/main/java/org/glassfish/jersey/internal/config/ExternalPropertiesAutoDiscoverable.java
+++ b/core-common/src/main/java/org/glassfish/jersey/internal/config/ExternalPropertiesAutoDiscoverable.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright (c) 2019 Oracle and/or its affiliates. All rights reserved.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0, which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception, which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ */
+
+package org.glassfish.jersey.internal.config;
+
+import org.glassfish.jersey.internal.spi.AutoDiscoverable;
+
+import javax.annotation.Priority;
+import javax.ws.rs.ConstrainedTo;
+import javax.ws.rs.RuntimeType;
+import javax.ws.rs.core.FeatureContext;
+
+@ConstrainedTo(RuntimeType.CLIENT) //server is configured directly in ResourceConfig
+@Priority(AutoDiscoverable.DEFAULT_PRIORITY)
+public class ExternalPropertiesAutoDiscoverable implements AutoDiscoverable {
+    @Override
+    public void configure(FeatureContext context) {
+        if (!context.getConfiguration().isRegistered(ExternalPropertiesConfigurationFactoryFeature.class)) {
+            ExternalPropertiesConfigurationFactoryFeature.getFactory().configure(context);
+            context.register(ExternalPropertiesConfigurationFactoryFeature.getFactory());
+        }
+    }
+}
--- a/core-common/src/main/java/org/glassfish/jersey/internal/config/ExternalPropertiesAutoDiscoverable.java
+++ b/core-common/src/main/java/org/glassfish/jersey/internal/config/ExternalPropertiesAutoDiscoverable.java
@@ -28,9 +28,9 @@ import javax.ws.rs.core.FeatureContext;
 public class ExternalPropertiesAutoDiscoverable implements AutoDiscoverable {
     @Override
     public void configure(FeatureContext context) {
-        if (!context.getConfiguration().isRegistered(ExternalPropertiesConfigurationFactoryFeature.class)) {
-            ExternalPropertiesConfigurationFactoryFeature.getFactory().configure(context);
-            context.register(ExternalPropertiesConfigurationFactoryFeature.getFactory());
+        if (!context.getConfiguration().isRegistered(ExternalPropertiesConfigurationFeature.class)) {
+            ExternalPropertiesConfigurationFactory.configure(context);
+            context.register(ExternalPropertiesConfigurationFeature.class);
         }
     }
 }
--- a/core-common/src/main/java/org/glassfish/jersey/internal/config/ExternalPropertiesAutoDiscoverable.java
+++ b/core-common/src/main/java/org/glassfish/jersey/internal/config/ExternalPropertiesAutoDiscoverable.java
@@ -29,7 +29,6 @@ public class ExternalPropertiesAutoDiscoverable implements AutoDiscoverable {
     @Override
     public void configure(FeatureContext context) {
         if (!context.getConfiguration().isRegistered(ExternalPropertiesConfigurationFeature.class)) {
-            ExternalPropertiesConfigurationFactory.configure(context);
             context.register(ExternalPropertiesConfigurationFeature.class);
         }
     }
--- a/core-common/src/main/java/org/glassfish/jersey/internal/config/ExternalPropertiesConfigurationFactory.java
+++ b/core-common/src/main/java/org/glassfish/jersey/internal/config/ExternalPropertiesConfigurationFactory.java
@@ -49,7 +49,7 @@ public class ExternalPropertiesConfigurationFactory {
      *
      * @return map of merged properties from all found/plugged providers
      */
-    protected static Map<String, Object> readExternalPropertiesMap() {
+    static Map<String, Object> readExternalPropertiesMap() {
 
         final ExternalConfigurationProvider provider = mergeConfigs(EXTERNAL_CONFIGURATION_PROVIDERS);
         return provider == null ? Collections.emptyMap() : provider.getProperties();
@@ -81,7 +81,7 @@ public class ExternalPropertiesConfigurationFactory {
      *
      * @return merged Model object with all properties
      */
-    protected static ExternalConfigurationModel getConfig() {
+    static ExternalConfigurationModel getConfig() {
         final ExternalConfigurationProvider provider = mergeConfigs(getExternalConfigurations());
         return provider == null ? null : provider.getConfiguration();
     }
--- a/core-common/src/main/java/org/glassfish/jersey/internal/config/ExternalPropertiesConfigurationFactory.java
+++ b/core-common/src/main/java/org/glassfish/jersey/internal/config/ExternalPropertiesConfigurationFactory.java
@@ -21,11 +21,11 @@ import org.glassfish.jersey.spi.ExternalConfigurationModel;
 import org.glassfish.jersey.spi.ExternalConfigurationProvider;
 
 import javax.annotation.Priority;
+import javax.ws.rs.Priorities;
 import javax.ws.rs.core.Configurable;
-import javax.ws.rs.core.Feature;
-import javax.ws.rs.core.FeatureContext;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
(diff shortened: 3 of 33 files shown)


arjantijms:
Just wondering, but why are Helidon and SmallRye providers needed? Would it not be an option to only use the public MicroProfile Config APIs?

senivam:
Actually those are external modules, they are not needed, just optional. The aim is to demonstrate possibilities to plug different implementations of MP configurations. 

And in the same way Jersey provides more tooling prepared out of the box so particular user is not required to implement s/his own connector if existing fits requirements.

smillidge:
But can't you just use the vanilla MicroProfile API to get the Config without these provider? 

`
ConfigProvider.getConfig();
`

jansupol:
@smillidge  Using the MP config is just one possibility. All this should be able to
1) work with an internal environment variables provider
2) any Jersey specific config facade provider providing 
  - MP Config
  - Config JSR config (should it ever be finished)
  - any other user provided config provider, which could be able to provide properties from other sources (XML, DB, any other remote property service,...)

senivam:
OK, I will introduce some minor changes in a while (according to your comments)

smillidge:
@jansupol I like the whole config provider SPI. I was commenting more in the context of creating the MP Config provider as an implementation of the SPI there shouldn't be the need to create a dependency to the specific MP Config implementations as the MP ConfigProvider itself should provide that abstraction.

senivam:
@smillidge, @arjantijms, @jansupol I've updated the PR, build is passed OK. Could you please take a look?

ext/microprofile/config/src/main/java/org/glassfish/jersey/config/ConfigurationModel.java
```diff
@@ -0,0 +1,153 @@
...
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception, which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ */
+
+package org.glassfish.jersey.config;
```
jansupol: Please make this `org.glassfish.jersey.microprofile.config`.

senivam: I'll do my best

core-common/src/main/java/org/glassfish/jersey/CommonProperties.java
```diff
@@ -216,6 +216,17 @@
      */
     public static final String OUTBOUND_CONTENT_LENGTH_BUFFER_SERVER = "jersey.config.server.contentLength.buffer";
 
+    /**
+     * Property which allows (if true) default System properties configuration provider.
+     *
+     * Effective if there are no any external properties providers
+     *
+     * Shall be set (if used) in system properties.
+     * @since 2.9
```
jansupol: Since 2.29

core-common/src/main/java/org/glassfish/jersey/internal/config/SystemPropertiesConfigurationModel.java
```diff
@@ -0,0 +1,216 @@
...
+
+    @Override
+    public Map<String, Object> getProperties() {
+        final Map<String, Object> result = new HashMap<>();
+
+        final Boolean allowSystemPropertiesProvider = as(
+                CommonProperties.ALLOW_SYSTEM_PROPERTIES_PROVIDER, Boolean.class
+        );
+        if (!Boolean.TRUE.equals(allowSystemPropertiesProvider)) {
+            log.finer("System properties configuration provider not allowed");
```
jansupol: For Logging messages, please use `localization.properties`

core-common/src/main/java/org/glassfish/jersey/internal/config/SystemPropertiesConfigurationModel.java
```diff
@@ -0,0 +1,216 @@
...
+            return result;
+        }
+
+        try {
+            AccessController.doPrivileged(PropertiesHelper.getSystemProperties())
+                    .forEach((k, v) -> result.put(String.valueOf(k), v));
+        } catch (SecurityException se) {
+            log.warning(se.getLocalizedMessage());
```
jansupol: Please create a message that would make more clear in what case the warning arose. For instance "Could not read system properties:"

core-common/src/main/java/org/glassfish/jersey/internal/config/package-info.java
```diff
@@ -0,0 +1,23 @@
...
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception, which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ */
+
+/**
+ * Microprofile configuration factory to handle external properties from different
```
jansupol: Configuration factory to handle external properties from different config SPI implementations?

pom.xml
```diff
@@ -2043,6 +2043,9 @@
         <weld3.version>3.0.0.Final</weld3.version>
         <xerces.version>2.11.0</xerces.version>
         <yasson.version>1.0.3</yasson.version>
+        <helidon.version>1.0.2</helidon.version>
```
jansupol: CQ required for Helidon, MP Config API, smallrye

pom.xml
```diff
@@ -2043,6 +2043,9 @@
...
+        <microprofile-api.version>1.2.1</microprofile-api.version>
```
jansupol: We may end up supporting different version for config and rest client. Please adjust the property name by "config".

core-common/src/main/java/org/glassfish/jersey/internal/config/ExternalPropertiesConfigurationFactory.java
```diff
@@ -0,0 +1,142 @@
...
+
+    /**
+     * Map of merged properties from all found providers
+     *
+     * @return map of merged properties from all found/plugged providers
+     */
+    public Map<String, Object> readExternalPropertiesMap() {
+
+        final ExternalConfigurationProvider provider = mergeConfigs(getExternalConfigurations());
+        return provider == null ? new HashMap<>() : provider.getProperties();
```
jansupol: Collections.emptyMap?

pom.xml
senivam: [CQ#: 19752](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=19752)
[CQ#: 19756](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=19756)

jansupol:
Given the issue with smallrye licensing, we would probably need to split the commit into to, the test module with smallrye in a separate commit. I assume that one will not make it to 2.29

senivam:
Short summary on current state of the PR:

Config engine is modified to be plugged automatically for both client and server parts of Jersey. 

Tests are adjusted accordingly. 

Small-ray test/example is removed from the PR due to license header error in one of its sources. Will be added separately.

ext/microprofile/config/pom.xml
```diff
@@ -0,0 +1,95 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>project</artifactId>
+        <groupId>org.glassfish.jersey.ext</groupId>
+        <version>2.29-SNAPSHOT</version>
+        <relativePath>../../pom.xml</relativePath>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>config</artifactId>
```
jansupol: Please align the name with other ext projects: jersey-mp-config

ext/pom.xml
```diff
@@ -44,6 +44,7 @@
         <module>cdi</module>
         <module>entity-filtering</module>
         <module>metainf-services</module>
+	    <module>microprofile/config</module>
```
jansupol: just microprofile. Each sub module should be dealt with in the microprofile project. There would be at least 2 submodules

core-common/src/main/java/org/glassfish/jersey/internal/config/ExternalPropertiesConfigurationFactoryFeature.java
```diff
@@ -0,0 +1,153 @@
...
+        }
+
+        return firstConfig;
+    }
+
+    private Set<ExternalConfigurationProvider> orderConfigs(List<ExternalConfigurationProvider> configurations) {
+
+        final SortedSet<ExternalConfigurationProvider> sortedSet = new TreeSet<>((config1, config2) -> {
+
+            if (
```
jansupol: If there would a config without priority, adding a new config WITH a priority would not help. The solution is to check the priority and if not found set a default priority, e.g. Priorities.USER, so that a new one could win.

core-common/src/main/java/org/glassfish/jersey/internal/config/ExternalPropertiesAutoDiscoverable.java
```diff
@@ -0,0 +1,36 @@
...
+import javax.ws.rs.core.FeatureContext;
+
+@ConstrainedTo(RuntimeType.CLIENT) //server is configured directly in ResourceConfig
+@Priority(AutoDiscoverable.DEFAULT_PRIORITY)
+public class ExternalPropertiesAutoDiscoverable implements AutoDiscoverable {
+    @Override
+    public void configure(FeatureContext context) {
+        if (!context.getConfiguration().isRegistered(ExternalPropertiesConfigurationFactoryFeature.class)) {
+            ExternalPropertiesConfigurationFactoryFeature.getFactory().configure(context);
+            context.register(ExternalPropertiesConfigurationFactoryFeature.getFactory());
```
jansupol: This would register the twice. `Context.register(ExternalPropertiesConfigurationFactoryFeature.class)`?

core-common/src/main/java/org/glassfish/jersey/spi/ExternalConfigurationModel.java
```diff
@@ -0,0 +1,68 @@
...
+     * @return current instance of the model
+     */
+    ExternalConfigurationModel mergeProperties(Map<String, Object> inputProperties);
+
+    /**
+     * Obtain config object
+     *
+     * @return external config provider
+     */
+    CONFIG getConfig();
```
jansupol: Is this the proper way? If two configs would be configured, they would be merged in the mergeConfigs, the first would win. Then the user would be surprised when received config of some other unexpected type. Maybe return a list of all configs? Maybe return config of expected class?

tests/integration/pom.xml
```diff
@@ -132,6 +132,8 @@
         <!-- TODO: conflict in ASM version (too old) of jetty plugin -->
         <!--<module>spring4</module>-->
         <module>tracing-support</module>
+        <module>microprofile/config/helidon</module>
```
jansupol: Why is it here? Should it be in ext/microprofile?

tests/integration/microprofile/config/webapp/pom.xml
```diff
@@ -0,0 +1,61 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jansupol: Please move this to the microprofile/config module

core-common/src/main/java/org/glassfish/jersey/spi/ExternalConfigurationModel.java
senivam: it's not what you think. This method allows to obtain Eclipse's MP config implementation (like Helidon) and work with it directly if required. Merge etc has nothing to do with that. It happens on the layer above. And it's low level MP Config's IMPL wrapper.

ext/microprofile/config/pom.xml
senivam: wasn't that you who told me to avoid MP naming in modules/PRs?

ext/pom.xml
senivam: correct, but we already have a PR #4086 which does this with its module, so it would be less conflicts when both PR's will be merged

tests/integration/microprofile/config/webapp/pom.xml
senivam: it was previously discussed (25.4.2019) and the decision was to extract that from ext/microprofile/config to tests. In ext/microprofile/config we shall have only dependency to Eclipse's microprofile API. And all direct implementations (small rye, helidon) shall be put into tests.

tests/integration/pom.xml
senivam: it was previously discussed (25.4.2019) and the decision was to extract that from ext/microprofile/config to tests. In ext/microprofile/config we shall have only dependency to Eclipse's microprofile API. And all direct implementations (small rye, helidon) shall be put into tests.

core-common/src/main/java/org/glassfish/jersey/internal/config/ExternalPropertiesAutoDiscoverable.java
jansupol: :) I saw that. It's due to mixing the feature and the factory

tests/integration/microprofile/config/webapp/pom.xml
jansupol: Good point

core-common/src/main/java/org/glassfish/jersey/internal/config/ExternalPropertiesAutoDiscoverable.java
```diff
@@ -0,0 +1,36 @@
...
+import javax.ws.rs.RuntimeType;
+import javax.ws.rs.core.FeatureContext;
+
+@ConstrainedTo(RuntimeType.CLIENT) //server is configured directly in ResourceConfig
+@Priority(AutoDiscoverable.DEFAULT_PRIORITY)
+public class ExternalPropertiesAutoDiscoverable implements AutoDiscoverable {
+    @Override
+    public void configure(FeatureContext context) {
+        if (!context.getConfiguration().isRegistered(ExternalPropertiesConfigurationFeature.class)) {
+            ExternalPropertiesConfigurationFactory.configure(context);
```
jansupol: ExternalPropertiesConfigurationFeature.configure(context)?

senivam: actually you are right - it's enough just to register feature into context and it will configure client later on.

PR #4117, opened 2019-04-30
Support for JAX-RS 2.2's hasProperties method

mkarg:
This PR provides support for the new `#hasProperties(String)` method introduced by JAX-RS 2.2.

The code changes allow potential performance improvements by directly accessing the underlying key set's retrieval method (where applicable) instead of pulling the actual property value. While the effect typically is not dramatic, this might be at least slightly faster in some use cases.


mkarg:
@arjantijms @jansupol Review please. :-)

PR #4118, opened 2019-04-30
Support for JAX-RS Extensions

mkarg:
**This draft is a work in progress aiming at support for JAX-RS 2.2's new `javax.ws.rs.ext.Extension` interface in Jersey 2.30.**

The draft PR serves to host the discussion about *how* the implementation should be designed. Possible choices could be:
1. As a standalone module, as currently showcased by the new incubator module demonstrated by this draft PR. *This would be most simple as it is almost done, but it does not feel like making it a first class Jersey citizen.*
2. Replacing the existing `ForcedAutoDiscoverable` interface everywhere in the code by the new `javax.ws.rs.ext.Extension`.
3. Adding it *additionally* to `ForcedAutoDiscoverable` in the existing code as a third code path besides handling `AutoDiscoverable` and `ForcedAutoDiscoverable`.
4. ...others...

Please discuss the design of choice. I would be happy to provide an according implementation.

### DIFF
--- a/incubator/jaxrs-extension/pom.xml
+++ b/incubator/jaxrs-extension/pom.xml
@@ -1,0 +1,51 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright (c) 2019 Markus KARG. All rights reserved.
+
+    This program and the accompanying materials are made available under the
+    terms of the Eclipse Public License v. 2.0, which is available at
+    http://www.eclipse.org/legal/epl-2.0.
+
+    This Source Code may also be made available under the following Secondary
+    Licenses when the conditions for such availability set forth in the
+    Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
+    version 2 with the GNU Classpath Exception, which is available at
+    https://www.gnu.org/software/classpath/license.html.
+
+    SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+
+-->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.glassfish.jersey.incubator</groupId>
+        <artifactId>project</artifactId>
+        <version>2.29-SNAPSHOT</version>
+    </parent>
+
+    <groupId>org.glassfish.jersey.spi</groupId>
+    <artifactId>jersey-spi-jaxrs-extension</artifactId>
+    <name>jersey-spi-jaxrs-extension</name>
+
+    <description>
+        Automatically loads JAX-RS components implementing the JAX-RS Extension interface.
+    </description>
+
+    <dependencies>
+        <dependency>
+            <groupId>jakarta.ws.rs</groupId>
+            <artifactId>jakarta.ws.rs-api</artifactId>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.glassfish.jersey.core</groupId>
+            <artifactId>jersey-common</artifactId>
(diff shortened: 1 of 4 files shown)


jansupol:
Honestly, I do not see how the Extension should work. The javadoc mandates the extension needs to be registered, but I do not know what that means. One may call register on that Extension, but nothing will happen, as Jersey would not know what to do with the Extension.

mkarg:
@jansupol JAX-RS's `Extension` standardized interface simply is another word for Jersey's `AutoForcedDiscoverable` product-specific interface. We have different options to implement it, as I layed out above. One of them is the attached incubator module. It already works fine with Jersey 2.28.

jansupol:
No, it is actually very far from `ForcedAutoDiscoverable`. The "standard" does not say anything about what should happen upon registration. Can you provide a test that demonstrates what you think should happen if the feature is implemented?

mkarg:
Yes certainly, I will provide some tests to demonstrate it (a demo app was linked at the JAX-RS PR). Unfortunately I'm about to leave for a trading fair, so I assume I will not find the time to provide this before mid of May.

jansupol:
@mkarg I must be missing something. 

The Extension javadoc does not mandate the Extension is a service. It just needs to be extended. Neither module.info does contain the Extension service. So how you make the ServiceLoader load all implementations of the Extension interface? 

The Extension javadoc mandates all the JAX-RS applications need to be affected by the Extension implementation. How would you scan the Extension from one WAR and use it for the second WAR?

mkarg:
@jansupol The Javadocs try to give most freedom to vendors. You misunderstood the *if* in the Javadocs: https://github.com/eclipse-ee4j/jaxrs-api/commit/fd8b006a7aa80e14571a122698979ef904c0474f#diff-f9bdf5a4627a02dc2992a4327baa1897R32. The `Extensions` interface was designed to provide a cross-vendor standard for the functionality already existing in the various products for auto-loading, with particular support for *the mechanics of* service loaders (i. e. `ServiceLoaders`'s *mechanics* -not particularly the *ServiceLoader* class itself- *MUST* be supported *IF* the extension *is* a service -like Jersey currently does which is *not* using the Service Loader class but works *like* one- as a least-common-denominator *IF* the extension *shall* be portable, but products *MAY* support *additional* proprietary loading mechanics, making the extension non-portable). Actually the intension when authoring the Javadocs was that `Extension` is implemented in Jersey *exactly* like ForcedAutoDiscoverable, or to replace it, so I tried to follow the JavaDocs found at https://github.com/eclipse-ee4j/jersey/blob/JAXRS_2_2/core-common/src/main/java/org/glassfish/jersey/internal/spi/ForcedAutoDiscoverable.java#L20. The incubator module was just a vehicle to demonstrate a proof-of-concept, not to actually implement it that way. Or in short: Think of it as a different name for `ForcedAutoDiscoverable`. It does not imply anything else than that, it just is a common cross-vendor name for auto-loading.

mkarg:
@jansupol Regarding WARs: There is a misunderstanding. The intention is not that one application can infliltrate extensions into *another* extension. Extensions are *not intended* as being part of an application. They are intendend as being seen as part of *the JAX-RS product*. So their power depends on *how* the extension actually is provided. Certainly, *when* being part of a WAR instead (one *could* do that), their power would be limited by any existing constraints.

jansupol:
Ah, I misunderstood the "in" in the javadoc. Like should the class be already in a configuration, it should be registered:
> automatically register a class in {@link javax.ws.rs.core.Configuration runtime * configurations}
Maybe register by, or register to would be more clear.

jansupol:
The `Extension` sound more like `AutoDiscoverable` than `ForcedAutodiscoverable`. The feature is so strong if someone puts the `Extension` on the application server classpath, it inevitably can break someone's application. The user would want to protect himself from the `Extension` and he should have an option to disable the `Extension` functionality. Note that allowing this does not break the specification compliance as the default is that the `Extension` is found and used.

jansupol:
The 'incubator` is used for features that are nice, but they are not heavily tested, or there is some potential issue, security for instance, and the incubator feature can be dropped in the next version. The incubator is definitely not where JAX-RS feature implementation should reside. Unless the JAX-RS feature itself is marked as incubation feature.

Issue #4111, opened 2019-04-24
Return a possibility to use HK2 AbstractBinder in Jersey


jansupol:
HK2 AbstractBinder has been replaced by Jersey AbstractBinder in 2.26. From that point, the older applications need to be updated to use Jersey AbstractBinder, which is breaking backward compatibility. The goal is to provide a possibility to use HK2 AbstractBinder again.

It is imperative that core-common must stay independent on HK2.

jansupol:
Relates to
#4092
#4026

PR #4122, opened 2019-05-02
Allow for registering HK2 AbstractBinder

jansupol:
This is to retain backward compatibility with 2.25.


jansupol:
Fixes #4111.

Issue #4119, opened 2019-05-01
Path annotation of different methods are interfering


a13u:
I am using jersey 2.28 via maven and have two methods in my service class (i am using a jetty server). 

```
@GET
@Path("{Prefix}{p:/?}{id: ((\\d+)?)}/abc{p2:/?}{number: (([A-Za-z0-9]*)?)}")
public Response doAnything2() {
    return Response.ok("sdf").build();

}
```
```
/*with that code below the GET method does not work, if I remove it, it works*/
@POST
@Path("{Prefix}{p:/?}{id: ((\\d+)?)}/abc/{yeah}")
public Response doAnything3() {
    return Response.ok("sdf").build();

}
```

With Postman I try to access 
`GET **P/abc/MyNumber**`

If I remove the method `doAnything3` I can access the GET endpoint. If I leave the method I can't access the GET endpoint (Method not found error).
Please, inform me if you have troubles reproducing the error.

jansupol:
Uri `P/abc/MyNumber` matches both `@Path("{Prefix}{p:/?}{id: ((\\d+)?)}/abc/{yeah}")` and `@Path("{Prefix}{p:/?}{id: ((\\d+)?)}/abc{p2:/?}{number: (([A-Za-z0-9]*)?)}")`. In your case, the method with `@POST` is matched first and a correct method designator (i.e. `@GET`) is not found. 

You could find a more simple example:

        @GET        
        @Path("x")
        public Response doAnything2() {
            return Response.ok("sdf").build();
        }

        @POST
        @Path("{name:[a-zA-Z][a-zA-Z_0-9]*}")
        public Response doAnything3() {
            return Response.ok("sdfa").build();
        }
In this case `POST /x` will not get matched.

The issue here is that you have two different `@PATH`s each matching the request, but the paths are not equal. 

Unfortunately, you could have a resource with many resource locators, each would lead to another locator only to eventually find the method designator differs from the request type. Then, backtracking would need to be implemented to fallback to the previous resource locator and to try a different path.

For various performance reasons, it was decided the backtracking is not to be implemented and your example is (not) working as designed.

a13u:
So even if the method (get,Post) is different, the Path Matches?

Von meinem iPhone gesendet

jansupol:
[...]

PR #4132, opened 2019-05-18
Allow for matching multiple resources with equal paths as in 4119

jansupol:

### DIFF
--- a/core-server/src/main/java/org/glassfish/jersey/server/internal/routing/PathMatchingRouter.java
+++ b/core-server/src/main/java/org/glassfish/jersey/server/internal/routing/PathMatchingRouter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2018 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2010, 2019 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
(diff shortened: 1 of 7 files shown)


jansupol:
fixes #4119

core-server/src/main/java/org/glassfish/jersey/server/internal/routing/PathMatchingRouter.java
```diff
@@ -80,11 +92,42 @@
...
+
+    /**
+     * Return true iff the method is a sub-resource and not a locator and http method designator does not match
+     * the request http method designator
+     * @param route current route representing resource method / locator
+     * @param context Contains Request to check the http method
+     * @return false if method designator matches
+     */
+    private boolean designatorMismatch(final Route route, final RequestProcessingContext context) {
+        final String httpMethod = context.request().getMethod();
```
tomas-langer: I would negate the method, so the if statement does not have to use `!`.
Also we can be more "clever" with the `HEAD` method - e.g. first try it for `HEAD` and then for `GET`, as `GET` can also respond to `HEAD` requests.

core-server/src/main/java/org/glassfish/jersey/server/model/RuntimeResource.java
```diff
@@ -128,7 +128,16 @@ public RuntimeResource build(RuntimeResource parent) {
...
+            final int cmp = PathPattern.COMPARATOR.compare(o1.getPathPattern(), o2.getPathPattern());
+            int locatorCmp = 0;
+            return cmp != 0
+                    ? cmp
+                    // quaternary key sorting those derived from
+                    // sub-resource methods ahead of those derived from sub-resource locators
+                    : (locatorCmp = o1.resourceLocators.size() - o2.resourceLocators.size()) != 0
+                        ? locatorCmp
+                        // compare the regexes then
+                        : o2.regex.compareTo(o1.regex);
```
tomas-langer: Nested ternary operators create quite unreadable code. I would refactor this to use an if statement instead.
```
final int cmp = PathPattern.COMPARATOR.compare(o1.getPathPattern(), o2.getPathPattern());
        if (cmp == 0) {
            final int locatorCmp = Integer.compare(o1.resourceLocators.size(), o2.resourceLocators.size());
            // quaternary key sorting those derived from
            // sub-resource methods ahead of those derived from sub-resource locators

            // compare the regexes if still equal
            return (locatorCmp == 0) ? o2.regex.compareTo(o1.regex) : locatorCmp;
        } else {
            return cmp;
        }
```

tests/e2e-server/src/test/java/org/glassfish/jersey/tests/e2e/server/routing/RegularExpressionsTest.java
```diff
@@ -0,0 +1,133 @@
...
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+public class RegularExpressionsTest extends JerseyTest {
+    public static String GET_VALUE = "get value";
+    public static String POST_VALUE = "post value";
+    public static String PUT_VALUE = "put value";
```
senivam: final, possibly private?

core-server/src/main/java/org/glassfish/jersey/server/internal/routing/PathMatchingRouter.java
```diff
@@ -80,11 +92,42 @@
...
+        final Router.Continuation result = Router.Continuation.of(context, acceptedRoute.next());
+
+        // tracing
+        tracingLogger.log(ServerTraceEvent.MATCH_PATH_SELECTED, acceptedRoute.routingPattern().getRegex());
+
+        return result;
+    }
+
+    /**
+     * Return true iff the method is a sub-resource and not a locator and http method designator does not match
```
senivam: typo... (iff)

jansupol: iff stands for [if and only if](https://en.wikipedia.org/wiki/If_and_only_if)

PR #4136, opened 2019-05-24
Fix ValidationErrorMessageBodyWriter

### LINKED ISSUE #4134: Fix ValidationErrorMessageBodyWriter
See 539195

senivam:
Fixes #4134 

### DIFF
--- a/ext/bean-validation/src/main/java/org/glassfish/jersey/server/validation/internal/ValidationErrorMessageBodyWriter.java
+++ b/ext/bean-validation/src/main/java/org/glassfish/jersey/server/validation/internal/ValidationErrorMessageBodyWriter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2012, 2018 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2012, 2019 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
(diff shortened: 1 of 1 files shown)


davewichers:
I just reviewed this 'fix' and implementing your own HTML encoder is TERRIBLE. There are MANY ways this custom encoder can be bypassed. I strongly suggest you replace use of this custom HTML encoder with a strong encoder provided by a standard security library.  Like the OWASP Java Encoder or something from Apache, or whatever.

senivam:
Hi @davewichers, thanks for the comment. Regarding the custom HTML encoder - for now, I do not recall exactly why this solution was taken, but most probably just because of avoiding another dependency in the project (due to CQ bureaucracy). The method probably was just copied from an open library. However you are right, it's always preferable to use some maintained solution instead of some custom hacks. So, I'm opening another PR for this: #5053, and filling a CQ about the usage of Apache's commons-text (using your original suggestion from bugs.eclipse.org).

jansupol:
@davewichers In general, we understand your concern, which is twofold: 

First, it is the message which can be displayed wrongly because of the HTML tags. That is what we tried to fix with the custom code. 

Second, it is the security that can be at stake here. But the error message comes from a `ConstraintViolationException` which is a user code, provided by a class that is part of the application. It is not a custom message that an attacker can pass in. For that, we see no urge to utilize a 3rd party library that a user would need to deploy along with Jersey to an application server/microservices runtime.  

You accuse us of implementing our own HTML encoder which you claim to be TERRIBLE without proof or an example of how this can cause a bypass as you suggest. Can you provide us with such an example?

PR #4150, opened 2019-06-08
prepare jersey (2.30) to be ready to work as JAXRS impl without default 

senivam:
JAXRS 2.2 removes Jersey as the default impl. This PR aims to provide possibility for Jersey to work without being default JAXRS impl


jansupol:
The work is done in 3.x branch. Closing.

PR #4155, opened 2019-06-11
adding Automatic-Module-Name header to manifest.mf for each Jersey's module 

senivam:


jansupol:
Did not check it completely, but not to forget - mp config does not have defined name

Issue #4157, opened 2019-06-11
CDI Providers in dependent scope used twice per request


pdudits:
On current master, a component that is CDI Dependent bean is found and instantiated twice by logic in `Providers#getAllServiceHolders`.

The problem was introduced with aa8020a3, after which all CDI beans are now looked up with qualifier `@Custom`, as well as without it. Being in dependent scope, two instances of them are created, and are correctly treated as non-equal by `ServiceHolderImpl`. This is disturbing especially with container filters that are executed twice.

The fix is straighforward, rolling back changes in `CdiComponentProvider` from mentioned commit, as original prioritization issue was fixed in the meantime (I believe in a62d1ff).

PR #4160, opened 2019-06-11
Prevent duplicate lookup of CDI components (#4157)

pdudits:
Even though I initially thought, that no changes in `Providers` are needed, `jersey-4099` proved me wrong.
After sorting unqualified providers, all tests pass.

This left me wondering about the purpose of `CustomAnnotationLiteral` in first place. Wasn't it intended to prefer application classes to container provided ones before support for `@Priority` was introduced? If yes, then lookups in `Providers` could be rewritten to single lookups, provided that all container classes are properly prioritized. But that would surely be scope of different issue.


jansupol:
I do prefer fix #4174 over this one.

jansupol:
Closing in favor of #4174

PR #4174, opened 2019-06-20
 CDI duplicate lookup fix

### LINKED ISSUE #4157: CDI Providers in dependent scope used twice per request

senivam:
fixes #4157 and improves components distinguishing  - custom VS jersey's internal - in case of CDI component provider.

### DIFF
--- a/core-common/src/main/java/org/glassfish/jersey/internal/inject/Providers.java
+++ b/core-common/src/main/java/org/glassfish/jersey/internal/inject/Providers.java
@@ -179,15 +179,16 @@ public final class Providers {
      * @return iterable of all available ranked service providers for the contract. Return value is never {@code null}.
      */
     public static <T> Iterable<RankedProvider<T>> getAllRankedProviders(InjectionManager injectionManager, Class<T> contract) {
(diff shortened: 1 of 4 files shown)


pdudits:
If I understand the code correctly, this would still cause the components to be instantiated and initialized twice. That could still cause some surprises. 

I'll yet test if this is the case.

jansupol:
@pdudits I believe this fix returns the behaviour to the state it was in 2.25.1. If you find any issue with how it works, please feel free to open a new issue against Jersey. For now, we will go with this fix.

PR #4175, opened 2019-06-20
Consumes and produces value should be now properly changed

### LINKED ISSUE #4173: MP REST Client MethodModel Builder incorrectly changes consumes value
I'm starting to use the Microprofile REST Client and came across the following at the bottom of the [Sample Definitions section of the documentation](https://download.eclipse.org/microprofile/microprofile-rest-client-1.2.0/microprofile-rest-client-1.2.0.html#_sample_definitions).

> Users may specify the media (MIME) type of the outbound request using the @Consumes annotation (this determine the Content-Type HTTP header), and the expected media type(s) of the response by using the @Produces annotation (the Accept HTTP header). This indicates that the client interface expects that the remote service consumes/produces the specified types.

Given the following interface for an authorization server endpoint:

```java
@Path("auth")
@RegisterRestClient(baseUri = "http://localhost")
public interface AuthService {
  @POST
  @Path("token")
  @Consumes(MediaType.APPLICATION_FORM_URLENCODED)
  @Produces(MediaType.APPLICATION_JSON)
  JsonObject getToken(@HeaderParam(HttpHeaders.AUTHORIZATION) String authHeader);
}
```

The request headers **should** be set as follows:

```
POST /auth/token.HTTP/1.1
Authorization: <authHeader value>
Accept: application/json
Content-Type: application/x-www-form-urlencoded
User-Agent: Jersey/2.29.payara-p2.(HttpUrlConnection.1.8.0_201)
Host: localhost
Connection: keep-alive
```

Unfortunately, the [MethodModel Builder's `build` method](https://github.com/eclipse-ee4j/jersey/blob/master/ext/microprofile/mp-rest-client/src/main/java/org/glassfish/jersey/microprofile/restclient/MethodModel.java#L597) examines the method's return type, sees that it is a `JsonValue` and changes the `consumes` value to `application/json`. The test is doing this if the return type **OR** *the entity* are JSON.

In this case it's incorrectly replacing the value set on the interface by setting the `consumes` value based on the return type. If any value were to be  for the return type, it should be the `produces` value. I question if the values should be overriden if they are set via annotation and the defaults.

As you can see from the headers above, I'm experiencing this in Payara's patched version of 2.29 as packaged in 5.192.

Verdent:
Fixes #4173

### DIFF
--- a/tests/integration/microprofile/rest-client/src/test/java/org/glassfish/jersey/restclient/ConsumesAndProducesTest.java
+++ b/tests/integration/microprofile/rest-client/src/test/java/org/glassfish/jersey/restclient/ConsumesAndProducesTest.java
@@ -1,0 +1,82 @@
+package org.glassfish.jersey.restclient;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.List;
+
+import javax.json.Json;
+import javax.ws.rs.client.ClientRequestContext;
+import javax.ws.rs.client.ClientRequestFilter;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+
+import org.eclipse.microprofile.rest.client.RestClientBuilder;
+import org.glassfish.jersey.server.ResourceConfig;
+import org.glassfish.jersey.test.JerseyTest;
+import org.glassfish.jersey.test.TestProperties;
+import org.junit.Test;
+
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertTrue;
+import static org.testng.Assert.fail;
+
+/**
+ * Created by David Kral.
+ */
+public class ConsumesAndProducesTest extends JerseyTest {
+    @Override
+    protected ResourceConfig configure() {
+        enable(TestProperties.LOG_TRAFFIC);
+        return new ResourceConfig(ApplicationResourceImpl.class);
+    }
+
+    @Test
+    public void testWithEntity() throws URISyntaxException {
+        ApplicationResource app = RestClientBuilder.newBuilder()
+                .baseUri(new URI("http://localhost:9998"))
+                .register(new TestClientRequestFilter(MediaType.APPLICATION_JSON, MediaType.APPLICATION_JSON))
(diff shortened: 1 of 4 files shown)


tests/integration/microprofile/rest-client/src/test/java/org/glassfish/jersey/restclient/ConsumesAndProducesTest.java
```diff
@@ -0,0 +1,82 @@
...
+        TestClientRequestFilter(String expectedAccept, String expectedContentType) {
+            this.expectedAccept = expectedAccept;
+            this.expectedContentType = expectedContentType;
+        }
+
+        @Override
+        public void filter(ClientRequestContext requestContext) {
+            assertTrue(requestContext.getHeaders().containsKey("Accept"));
+            List<Object> accept = requestContext.getHeaders().get("Accept");
+            if (!accept.contains(expectedAccept) && !accept.contains(MediaType.valueOf(expectedAccept))) {
```
senivam: isn't it better to use 
```
assertTrue(accept.contains(expectedAccept) && accept.contains(MediaType.valueOf(expectedAccept)));
```
?

tests/integration/microprofile/rest-client/src/test/java/org/glassfish/jersey/restclient/ConsumesAndProducesTest.java
```diff
@@ -0,0 +1,82 @@
...
+                fail();
+            }
+
+            assertTrue(requestContext.getHeaders().containsKey("Content-Type"));
+            List<Object> contentType = requestContext.getHeaders().get("Content-Type");
+            assertEquals(contentType.size(), 1);
+            if (!contentType.contains(expectedContentType) && !contentType.contains(MediaType.valueOf(expectedContentType))) {
```
senivam: the same thing here for:
```
assertTrue(contentType.contains(expectedContentType) && contentType.contains(MediaType.valueOf(expectedContentType)));
```

senivam:
LGTM

PR #4176, opened 2019-06-20
Unify microprofile modules structure

senivam:


ext/microprofile/pom.xml
```diff
@@ -15,7 +15,8 @@
 
     <modules>
         <module>mp-rest-client</module>
+	    <module>mp-config</module>
```
ielatif: leading spaces

senivam: fixed

PR #4199, opened 2019-07-09
Added client support for ConnectionKeepAliveStrategy and ConnectionReuseStrategy

rsandidfc:
Added client support for ConnectionKeepAliveStrategy and ConnectionReuseStrategy.

Signed-off-by: Richard Sand <rsand@idfconnect.com>

### DIFF
--- a/connectors/apache-connector/src/main/java/org/glassfish/jersey/apache/connector/ApacheClientProperties.java
+++ b/connectors/apache-connector/src/main/java/org/glassfish/jersey/apache/connector/ApacheClientProperties.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2013, 2018 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2013, 2019 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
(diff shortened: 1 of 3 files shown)


jansupol:
Unfortunately, every single commit need to pass the Eclipse Check for the Sign-off-by footer. Only the last commit has the brackets around the email, though.

senivam:
@rsandidfc actually it is not required to close a PR in order to modify something. Just continue committing and pushing to the same branch from which the PR is originated and your new commits will be added to existing PR. 

And to fix already existing commit to pass ECA check I would advise to use git command line at your local repository:
`git commit --amend --signoff`
but if that commit is already in github, afterward you shall push it with force
`git push -f origin <branch_name>`
to overwrite existing commit. 

And furthermore, if required, you can squash several commits into one using interactive rebasing, like
`git rebase -i HEAD~3`
this will open a text editor with listed commits (3 in our case) and you will be able to do all you require with them (squash, delete, modify commit message etc).
But that requires push with force afterwards (because all 3 commits are already pushed into a branch).

rsandidfc:
I'm going to close this one too, I submitted a new pull request that I *think* should pass everythingm thanks guys!

PR #4202, opened 2019-07-10
Added support for Apache HTTP Client ConnectionKeepAliveStrategy and ConnectionReuseStrategy

rsandidfc:
Added support for Apache HTTP Client ConnectionKeepAliveStrategy and ConnectionReuseStrategy

3rd time's the charm? Added the support, checkstyle conforms, copywrite years updated, and signed-off-by is below. Fingers crossed..........

Signed-off-by: Richard A Sand <rsand@idfconnect.com>

### DIFF
--- a/connectors/apache-connector/src/main/java/org/glassfish/jersey/apache/connector/ApacheConnector.java
+++ b/connectors/apache-connector/src/main/java/org/glassfish/jersey/apache/connector/ApacheConnector.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2018 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2010, 2019 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
@@ -56,7 +56,7 @@ import org.glassfish.jersey.message.internal.HeaderUtils;
 import org.glassfish.jersey.message.internal.OutboundMessageContext;
 import org.glassfish.jersey.message.internal.ReaderWriter;
 import org.glassfish.jersey.message.internal.Statuses;
-
+import org.apache.http.ConnectionReuseStrategy;
 import org.apache.http.Header;
 import org.apache.http.HttpEntity;
 import org.apache.http.HttpHost;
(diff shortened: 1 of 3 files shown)


connectors/apache-connector/src/main/java/org/glassfish/jersey/apache/connector/ApacheConnector.java
```diff
@@ -197,6 +200,32 @@
             }
         }
 
+        final Object keepAliveStrategy = config.getProperties().get(ApacheClientProperties.KEEPALIVE_STRATEGY);
+        if (keepAliveStrategy != null) {
+            if (!(keepAliveStrategy instanceof ConnectionKeepAliveStrategy)) {
+                LOGGER.log(
+                        Level.WARNING,
+                        LocalizationMessages.IGNORING_VALUE_OF_PROPERTY(
```
senivam: the code actually does not ignore value of the property, it only logs intention to do so. And the if inside another if could be rewritten as if (A && !B).

connectors/apache-connector/src/main/java/org/glassfish/jersey/apache/connector/ApacheConnector.java
```diff
@@ -197,6 +200,32 @@
...
+                );
+            }
+        }
+
+        final Object reuseStrategy = config.getProperties().get(ApacheClientProperties.REUSE_STRATEGY);
+        if (reuseStrategy != null) {
+            if (!(reuseStrategy instanceof ConnectionReuseStrategy)) {
+                LOGGER.log(
+                        Level.WARNING,
+                        LocalizationMessages.IGNORING_VALUE_OF_PROPERTY(
```
senivam: The same thing is valid here: 
the code actually does not ignore value of the property, it only logs intention to do so. And the if inside another if could be rewritten as if (A && !B).

rsandidfc: Whoops, how silly, I just copied the connectionManager block and didn't change it to null the values if invalid. I'll fix now

rsandidfc: Ok I've committed the update to the pull

Issue #4189, opened 2019-07-05
Jersey Entity Filter Threads Racing issue leads to Corrupted Entity Graph and Object Graph  


shuangxipei:
We are using Jersey version 2.27
The Jersey configuration has the  SelectableEntityFilteringFeature configured.
The server environment is Felix server with Jetty.
When we send two concurrent same Web Service API calls to the server, the entity in the response got corrupted, i.e., some attributes in the entity got lost. 
This is intermittent, but if it happens, all subsequent calls to that corrupted web service API will got back the same corrupted entity with same attributes missed.

Correct response entity example

[
  {
    "jobInstanceId": "1",
    "jobId": "1",
    "startTime": "2019-06-26T01:52:54",
    "instanceState": {
      "state": "FINISHED",
      "detail": "Data discovery job finished successfully",
      "reportedTime": "2019-06-26T01:53:04",
      "stateSeq": 15
    },
    "schedule": "NOW",
  },
  {
    "jobInstanceId": "2",
    "jobId": "2",
    "startTime": "2019-06-26T01:55:24",
    "instanceState": {
      "state": "FINISHED",
      "detail": "Data discovery job finished successfully",
      "reportedTime": "2019-06-26T01:55:38",
      "stateSeq": 30
    },
    "schedule": "NOW",
  },
  {
    "jobInstanceId": "3",
    "jobId": "3",
    "startTime": "2019-06-26T02:06:52",
    "instanceState": {
      "state": "FINISHED",
      "detail": "Data discovery job finished successfully",
      "reportedTime": "2019-06-26T02:07:06",
      "stateSeq": 45
    },
    "schedule": "NOW",
  }
]

Corrupted response entity example

[
  {
    "jobInstanceId": "1"
  },
  {
    "jobInstanceId": "2"
  },
  {
    "jobInstanceId": "3"
  }
]

How to reproduce:
   1) Restart the JVM, sent two concurrent Web Service calls for the same URL (i.e, invoke the same web service at the same time).
    2) check if the response entity corrupted or not
    3) if not corrupted, repeat from step 1). You may need many times retries.

Observed Jersey Exception stack trace (occurred in both Moxy provider and Jackson provide.)

With Moxy Provider

	|Caused by: 
	|java.util.ConcurrentModificationException
	|	at java.util.HashMap$HashIterator.nextNode(HashMap.java:1442)
	|	at java.util.HashMap$KeyIterator.next(HashMap.java:1466)
[...]

Issue #4187, opened 2019-07-03
Race condition in EntityFiltering


jansupol:
The symptom is that when an entity filtering is used, only some fields are returned, not all selected.

The log can contain an exception:
> Caused by:
        |java.util.ConcurrentModificationException
        |       at java.util.HashMap$HashIterator.nextNode(HashMap.java:1437)
        |       at java.util.HashMap$EntryIterator.next(HashMap.java:1471)
        |       at java.util.HashMap$EntryIterator.next(HashMap.java:1469)
        |       at org.glassfish.jersey.internal.guava.AbstractMapBasedMultimap$AsMap$AsMapIterator.next(AbstractMapBasedMultimap.java:1348)
        |       at org.glassfish.jersey.internal.guava.AbstractMapBasedMultimap$AsMap$AsMapIterator.next(AbstractMapBasedMultimap.java:1336)
        |       at org.glassfish.jersey.internal.guava.TransformedIterator.next(TransformedIterator.java:45)
        |       at org.glassfish.jersey.internal.guava.AbstractMultimap.containsValue(AbstractMultimap.java:41)
        |       at org.glassfish.jersey.internal.guava.HashMultimap.containsValue(HashMultimap.java:44)
        |       at org.glassfish.jersey.message.filtering.EntityGraphImpl.presentInScopes(EntityGraphImpl.java:205)
        |       at org.glassfish.jersey.message.filtering.DefaultEntityProcessor.process(DefaultEntityProcessor.java:95)
        |       at org.glassfish.jersey.message.filtering.DefaultEntityProcessor.process(DefaultEntityProcessor.java:79)
        |       at org.glassfish.jersey.message.filtering.EntityInspectorImpl.inspectEntityProperties(EntityInspectorImpl.java:173)
        |       at org.glassfish.jersey.message.filtering.EntityInspectorImpl.inspect(EntityInspectorImpl.java:104)
        |       at org.glassfish.jersey.message.filtering.spi.AbstractObjectProvider.getFilteringObject (AbstractObjectProvider.java:90)

jansupol:
When `EntityInspectorImpl.inspect` is called, two threads can start to populate the EntityGraph leading to the exception.

Also, if a first thread is still populating the EntityGraph, another thread can invoke the method, 
`graphProvider.containsEntityGraph` will return `true`, the thread can create an `ObjectGraph` from the not fully populated `EntityGraph` in `AbstractObjectProvider#getFilteringObject`.

PR #4204, opened 2019-07-11
Prevent race condition in entity filtering

jansupol:


jansupol:
Unfortunately, I was not able to come up with a solid reproducer/test. 

However, when one thread stored a graph `graphProvider.getOrCreateEntityGraph(entityClass, forWriter)`, another thread used the graph before the graph was populated by the first thread. It resulted in a situation where everything was filtered out in an HTTP response, as the unpopulated graph did not contain the entities that should not have been filtered out.

jansupol:
Fixes #4189 #4187

PR #4213, opened 2019-07-31
Remove openjdk8 option from Travis build matrix 

jansupol:
Replace oraclejdk11 with openjdk11

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,8 +9,7 @@ services:
 
 jdk:
   - oraclejdk8
-  - oraclejdk11
-  - openjdk8
+  - openjdk11
 
 cache:
   directories:


jansupol:
- Oracle JDK11 is no longer available on Travis, and Open JDK 11 is used instead.
- Open JDK 8 seems to be no longer available on Trusty as well, and Oracle JDK 8 is used. See for instance [job](https://travis-ci.org/eclipse-ee4j/jersey/jobs/566008947) for open JDK 8. While the job description says:
   >  JDK: openjdk8 Java

   At line 372, it says:
   > Java home: /usr/lib/jvm/java-8-oracle/jre

   Removing open jdk8 for not running the tests on Oracle JDK 8 twice, then.

PR #4220, opened 2019-08-08
OutboundJaxrsResponse.from(): Don't propagate hop-by-hop headers

xkr47:
Without this PR, when you have jersey-server forwarding a request with jersey-client to a remote service and there is an error response from the remote service with `Transfer-Encoding: chunked` in the response headers, jersey-server propagates the Transfer-Encoding header back to its response. But jersey-server does not actually encode the response using chunked encoding. Also it adds a Content-Length header, which is also forbidden when a Transfer-Encoding header is present. So you end up with a response similar to this:

    HTTP/1.1 500 Internal Server Error
    Server: Foo
    Transfer-Encoding: chunked
    Content-Length: 25

    { "error": "Some error" }

A client seeing such a response can easily break down, for example with jersey-client itself (using Apache HttpComponents for transport):

```java
javax.ws.rs.client.ResponseProcessingException: Failed to convert a response into an exception.
	at org.glassfish.jersey.client.JerseyInvocation.convertToException(JerseyInvocation.java:1111)
	at org.glassfish.jersey.client.JerseyInvocation.translate(JerseyInvocation.java:948)
	at org.glassfish.jersey.client.JerseyInvocation.lambda$invoke$2(JerseyInvocation.java:787)
	at org.glassfish.jersey.internal.Errors.process(Errors.java:316)
	at org.glassfish.jersey.internal.Errors.process(Errors.java:298)
	at org.glassfish.jersey.internal.Errors.process(Errors.java:229)
	at org.glassfish.jersey.process.internal.RequestScope.runInScope(RequestScope.java:414)
	at org.glassfish.jersey.client.JerseyInvocation.invoke(JerseyInvocation.java:785)
	at org.glassfish.jersey.client.JerseyInvocation$Builder.method(JerseyInvocation.java:438)
	at org.glassfish.jersey.client.JerseyInvocation$Builder.get(JerseyInvocation.java:329)
	at Foo.main(Foo.java:43)
Caused by: javax.ws.rs.ProcessingException: Failed to buffer the message content input stream.
	at org.glassfish.jersey.message.internal.InboundMessageContext.bufferEntity(InboundMessageContext.java:931)
	at org.glassfish.jersey.client.InboundJaxrsResponse.bufferEntity(InboundJaxrsResponse.java:161)
	at org.glassfish.jersey.client.JerseyInvocation.convertToException(JerseyInvocation.java:1065)
	... 41 more
Caused by: org.apache.http.MalformedChunkCodingException: Bad chunk header: {
	at org.apache.http.impl.io.ChunkedInputStream.getChunkSize(ChunkedInputStream.java:277)
	at org.apache.http.impl.io.ChunkedInputStream.nextChunk(ChunkedInputStream.java:225)
	at org.apache.http.impl.io.ChunkedInputStream.read(ChunkedInputStream.java:184)
	at org.apache.http.conn.EofSensorInputStream.read(EofSensorInputStream.java:135)
	at java.io.BufferedInputStream.read1(BufferedInputStream.java:284)
	at java.io.BufferedInputStream.read(BufferedInputStream.java:345)
	at java.io.FilterInputStream.read(FilterInputStream.java:133)
	at java.io.FilterInputStream.read(FilterInputStream.java:107)
	at org.glassfish.jersey.message.internal.ReaderWriter.writeTo(ReaderWriter.java:116)
	at org.glassfish.jersey.message.internal.InboundMessageContext.bufferEntity(InboundMessageContext.java:918)
	... 43 more
```

The HTTP/1.1 spec says that connection-specific aka hop-by-hop headers (including Transfer-Encoding) must not be forwarded, and the code doing the forwarding in this case is `OutboundJaxrsResponse.from()`. It just copies all headers as-is, and the headers have not been filtered earlier in the process either.

This PR implements a helper in HeaderUtils (+ unit test) that copies only non-hop-by-hop headers and takes it into use in `OutboundJaxrsResponse.from()` (which imo is the correct place to do it).


core-common/src/main/java/org/glassfish/jersey/message/internal/OutboundJaxrsResponse.java
```diff
@@ -71,7 +72,7 @@ public static OutboundJaxrsResponse from(javax.ws.rs.core.Response response) {
         } else {
             final StatusType status = response.getStatusInfo();
             final OutboundMessageContext context = new OutboundMessageContext();
-            context.getHeaders().putAll(response.getMetadata());
+            copyEndToEndHeaders(response.getHeaders(), context.getHeaders());
```
jansupol: Should this be ` copyEndToEndHeaders(response.getMetadata(), context.getHeaders());`?

core-common/src/main/java/org/glassfish/jersey/message/internal/HeaderUtils.java
```diff
@@ -47,6 +53,20 @@
...
+    private static final Set<String> HOP_BY_HOP_HEADERS = new HashSet<>(32);
+    private static final Pattern CONNECTION_VALUE_SPLITTER = compile("\\s*+,[\\s,]*+");
+
+    static {
+        HOP_BY_HOP_HEADERS.add("connection");
+        HOP_BY_HOP_HEADERS.add("keep-alive");
+        HOP_BY_HOP_HEADERS.add("proxy-authenticate");
+        HOP_BY_HOP_HEADERS.add("proxy-authorization");
+        HOP_BY_HOP_HEADERS.add("te");
+        HOP_BY_HOP_HEADERS.add("trailer");
```
jansupol: `Trailers` is the header name

jansupol:
Hop by hop headers are defined in [RFC 2616](https://tools.ietf.org/html/rfc2616#page-92). [RFC 7230](https://tools.ietf.org/html/rfc7230#section-6.1) has made it obsolete and hop by hop headers should be defined in Connection header.

For compatibility with RFC 2616 this makes sense, but right now I'd say this should rather be optional. I would assume the list of headers should be configurable. [RFC 2068](https://www.ietf.org/rfc/rfc2068.txt) for instance defines additional hop by hop headers.

core-common/src/main/java/org/glassfish/jersey/message/internal/OutboundJaxrsResponse.java
xkr47: I was following the instructions found in the `getMetadata()` javadoc :
```java
    /**
     * See {@link #getHeaders()}.
     *
     * This method is considered deprecated. Users are encouraged to switch their
     * code to use the {@code getHeaders()} method instead. The method may be annotated
     * as {@link Deprecated &#64;Deprecated} in a future release of JAX-RS API.
     *
     * @return response headers as a multivalued map.
     */
    public abstract MultivaluedMap<String, Object> getMetadata();
```

core-common/src/main/java/org/glassfish/jersey/message/internal/HeaderUtils.java
xkr47: Have I missed something?
* [RFC2616](https://tools.ietf.org/html/rfc2616#section-14.40)
* [RFC7230](https://tools.ietf.org/html/rfc7230#section-4.4)
* [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Connection) documentation talk about the `Trailer` header.

xkr47:
> For compatibility with RFC 2616 this makes sense, but right now I'd say this should rather be optional.

My feeling is RFC723x were maybe catering for continuity of the HTTP/1.x series, but now we instead have HTTP/2 and HTTP/3 which have diverged heavily from this direction. HTTP/1.1 will remain strong in the enterprise environments for at least a decade but I don't expect any further revisions of HTTP/1 and therefore I believe RFC2616 compatibility will remain a necessity indefinitely.

Regarding current status; no HTTP/1.1 server I have seen specifies "transfer-encoding" in their Connection header. So making this optional will just keep jersey incompatible with *all* HTTP/1.1 servers currently implemented. Figuring out that this needs to be turned on for **correct operation** when getting some strange protocol errors might be tricky. Please don't do that.

Based on these thoughts I would assume everybody implementing HTTP/1.1 in the future will assume & retain RFC2616 compatibility and therefore refrain from adding "transfer-encoding" into the "Connection" header to avoid wasting bandwidth.

> I would assume the list of headers should be configurable. [RFC 2068](https://www.ietf.org/rfc/rfc2068.txt) for instance defines additional hop by hop headers.

Good point about 2068, will add the `Public` header to the list.

Regarding the request for configurable additional headers, do you have a proposal as for what method should be used to configure these? I am not familiar enough with Jersey in this respect. A system property?

core-common/src/main/java/org/glassfish/jersey/message/internal/HeaderUtils.java
jansupol: My bad. [RFC2616 Section 13.5.1](https://tools.ietf.org/html/rfc2616#page-92) talks about "trailers". I have not checked the [errata](https://www.rfc-editor.org/errata/eid4522).

jansupol:
> Regarding current status; no HTTP/1.1 server I have seen specifies "transfer-encoding" in their Connection header. So making this optional will just keep jersey incompatible with all HTTP/1.1 servers currently implemented. Figuring out that this needs to be turned on for correct operation when getting some strange protocol errors might be tricky. Please don't do that.

Agreed.

> Regarding the request for configurable additional headers, do you have a proposal as for what method should be used to configure these? I am not familiar enough with Jersey in this respect. A system property?

I still can imagine use-cases where the headers would be demanded.  So I would stick with configurable settings. `OutboundJaxrsResponse` is used in two places:
 - [ClientResponse](https://github.com/eclipse-ee4j/jersey/blob/068b3b23aaaa4eed5a02dfe210f8e09922074e76/core-client/src/main/java/org/glassfish/jersey/client/ClientResponse.java#L66). This is good, as `ClientResponse` has an instance of `ClientRequest`/`ClientRequestContext` that has `#getProperty` method. 
- [ContainerResponse](https://github.com/eclipse-ee4j/jersey/blob/068b3b23aaaa4eed5a02dfe210f8e09922074e76/core-server/src/main/java/org/glassfish/jersey/server/ContainerResponse.java#L65). This is good, as `ContainerResponse` has an instance of `ContainerRequest`/`ContainerRequestContext` that has `#getProperty` method. 

So it is possible to create a property in [ServerProperties](https://github.com/eclipse-ee4j/jersey/blob/068b3b23aaaa4eed5a02dfe210f8e09922074e76/core-server/src/main/java/org/glassfish/jersey/server/ServerProperties.java#L178) (assuming the filtering is not needed on a client-side) with a value of `List<String>` or `String[]` which would contain the headers you want to filter. You could pass the property to the `OutboundJaxrsResponse#from` then (by adding an argument) for filtering.

It is possible to create a Feature that would have a method (for possible overriding) that would return the list of headers and it would be called from `#configure` while registering the property.

xkr47:
Ok great;
Did I understand correctly that you suggest the following:
1. defaulting to the headers currently hardcoded in the PR?
2. implementing both the ServerProperties-based and the Feature-based approach for overriding the default?

jansupol:
I would suggest the `ServerProperties` would contain just the property name. The Feature would name (hardcode) the list of header names, something like:
```
public class HttpHeaderFilteringFeature implements Feature { 
   public boolean configure(FeatureContext context) {
       context.property(ServerProperties.FILTERED_HTTP_HEADERS, getHttpHeaderNames())
   }
  public List<String>getHttpHeaderNames() { ... }
}
```
It would be needed to register the feature for the effect. (I wanted to say I agree with not relying on `Connection` header not that I agree with the default filtering).  I understand your 
> So making this optional will just keep jersey incompatible with all HTTP/1.1 servers

but this is a breaking change. If the headers are by default filtered, it is hard to say how many applications it could break. I have not registered a similar filtering request on Jersey yet, so currently, I am inclined to not register it by default.

Note that with Jersey being able to use [Microprofile Config](https://github.com/eclipse-ee4j/jersey/tree/master/ext/microprofile/mp-config), it is possible to pass the header names from a config file, too.

Issue #4184, opened 2019-07-02
Spring5 missing in BOM


dreis2211:
Hi,

it seems that the spring5 module seems to be missing in the BOM. Could this be added, please? Or is there a reason why it isn't included? Duplicated classes maybe (in which case I'd favor replacing spring4 with spring5)?

Cheers,
Christoph

PR #4221, opened 2019-08-09
Better specify HK2 and Spring dependencies

jansupol:


jansupol:
Fixes #4184

ext/spring4/pom.xml
```diff
@@ -150,10 +150,6 @@
```
senivam: copyright year?

PR #4222, opened 2019-08-10
Fix NettyInputStream ByteBuf leak

Gbillou:
If not closed here `take` goes unreferenced leading to a leak of the underlying ByteBuf.
`LEAK: ByteBuf.release() was not called before it's garbage-collected. See http://netty.io/wiki/reference-counted-objects.html for more information.`

### DIFF
--- a/connectors/netty-connector/src/main/java/org/glassfish/jersey/netty/connector/internal/NettyInputStream.java
+++ b/connectors/netty-connector/src/main/java/org/glassfish/jersey/netty/connector/internal/NettyInputStream.java
@@ -68,9 +68,11 @@ public class NettyInputStream extends InputStream {
         this.isList = isList;
     }
 
-    @Override
-    public int read(byte[] b, int off, int len) throws IOException {
+    private interface ISReader {
+        int readFrom(InputStream take) throws IOException;
+    }
 
+    private int readInternal(ISReader takeRead) throws IOException {
         if (end) {
             return -1;
         }
@@ -83,7 +85,7 @@ public class NettyInputStream extends InputStream {
                 return -1;
             }
 
-            int read = take.read(b, off, len);
+            int read = takeRead.readFrom(take);
 
             if (take.available() > 0) {
                 isList.addFirst(take);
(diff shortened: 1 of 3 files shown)


jansupol:
Please follow the [WIKI ](https://github.com/eclipse-ee4j/jersey/wiki/Missing-ECA,-Missing-Sign-off)

connectors/netty-connector/src/main/java/org/glassfish/jersey/netty/connector/internal/NettyInputStream.java
```diff
@@ -87,6 +87,8 @@ public int read(byte[] b, int off, int len) throws IOException {
```
senivam: could you please update copyright year in both files (to 2019) - copyright header at the top of each file?

senivam: thank you for changes, and a question - you are changing method ```read(byte[] b, int off, int len)``` but immediately under this method there is method ```read()``` without any parameters, but inner logic looks very similar, except the very reading. So, the question - shall  ```InputStream take``` be closed as well in the ```read()``` method?

Gbillou: Indeed, I refactored the code a bit to avoid this mistake in the future.

connectors/netty-connector/src/main/java/org/glassfish/jersey/netty/connector/internal/NettyInputStream.java
```diff
@@ -68,9 +68,11 @@ public NettyInputStream(LinkedBlockingDeque<InputStream> isList) {
...
 
-    @Override
-    public int read(byte[] b, int off, int len) throws IOException {
+    private interface ISReader {
+        int readFrom(InputStream take) throws IOException;
+    }
 
+    private int readInternal(ISReader isReader) throws IOException {
         if (end) {
             return -1;
```
rafaelcoelho: sdada

rafaelcoelho: test

connectors/netty-connector/src/main/java/org/glassfish/jersey/netty/connector/internal/NettyInputStream.java
```diff
@@ -68,9 +68,11 @@ public NettyInputStream(LinkedBlockingDeque<InputStream> isList) {
         this.isList = isList;
     }
 
-    @Override
-    public int read(byte[] b, int off, int len) throws IOException {
+    private interface ISReader {
```
rafaelcoelho: testing comment

connectors/netty-connector/src/main/java/org/glassfish/jersey/netty/connector/internal/NettyInputStream.java
```diff
@@ -68,9 +68,11 @@ public NettyInputStream(LinkedBlockingDeque<InputStream> isList) {
...
+        int readFrom(InputStream take) throws IOException;
```
rafaelcoelho: test

Issue #4214, opened 2019-08-01
Jersey with Jackson exposes that fact to a potential attacker sending misformed  JSON data


jansupol:
FIX [CWE](https://bugs.eclipse.org/bugs/show_bug.cgi?id=540012)

PR #4225, opened 2019-08-14
Add an option to not register the Jackson's ExceptionMappers by JacksonFeature

jansupol:

### DIFF
--- a/tests/e2e-entity/src/test/java/org/glassfish/jersey/tests/e2e/json/JsonWithoutExceptionMapperTest.java
+++ b/tests/e2e-entity/src/test/java/org/glassfish/jersey/tests/e2e/json/JsonWithoutExceptionMapperTest.java
@@ -64,7 +64,7 @@ public class JsonWithoutExceptionMapperTest extends JerseyTest {
     public void testZullBooleanValue() {
         String response = target().request(MediaType.APPLICATION_JSON)
                 .buildPost(Entity.entity("zull", MediaType.APPLICATION_JSON_TYPE)).invoke().readEntity(String.class);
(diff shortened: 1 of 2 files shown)


jansupol:
Fixes #4214

tests/e2e-entity/src/test/java/org/glassfish/jersey/tests/e2e/json/JsonWithoutExceptionMapperTest.java
```diff
@@ -0,0 +1,70 @@
...
+    protected Application configure() {
+        return new ResourceConfig(Resource.class, LowPriorityExceptionMapper.class)
+                .register(JacksonFeature.withoutExceptionMappers());
+    }
+
+    @Test
+    public void testZullBooleanValue() {
+        String response = target().request(MediaType.APPLICATION_JSON)
+                .buildPost(Entity.entity("zull", MediaType.APPLICATION_JSON_TYPE)).invoke().readEntity(String.class);
+        Assert.assertFalse(response.contains("Unrecognized token 'zull'"));
```
tomas-langer: I would just check that it does not contain `zull` string. In case the next version of Jackson slightly modifies the message and the mapper would get registered, you would not catch it here.

jansupol: `Unrecognized token` is the message the issue complains about so it makes sense to have it there , and for the case you describe there is the second assertion, but ok, I made the assertion to check `zull` only.

PR #4226, opened 2019-08-14
Use configured executor service for RX.

tomas-langer:
Signed-off-by: Tomas Langer <tomas.langer@oracle.com>


tomas-langer:
Closing this PR, as the behavior is defined by the spec and this is not as simple a fix as it seemed.

Issue #4177, opened 2019-06-21
OsgiRegistry classToBundleMapping does not store classes for packages containing .class correctly


WouterBanckenACA:
We are running Jersey in an OSGi environment and we are experiencing the following error at startup:

```
Could not start the servlet context for context path [] - username= - trace-id=
javax.servlet.ServletException: org.ops4j.pax.web.service.spi.model.ServletModel-106@84178cbf==org.glassfish.jersey.servlet.ServletContainer,0,false
	at org.eclipse.jetty.servlet.ServletHolder.initServlet(ServletHolder.java:661) ~[214:org.ops4j.pax.web.pax-web-jetty-bundle:6.0.7]
...
Caused by: java.lang.RuntimeException: A class file of the class name, org.example.classification.rest.internal.MyClassificationResource, is identified but the class could not be found.
	at org.glassfish.jersey.server.internal.scanning.AnnotationAcceptingListener$AnnotatedClassVisitor.getClassForName(AnnotationAcceptingListener.java:265) ~[?:?]

...
Caused by: java.lang.ClassNotFoundException: org.example.classification.rest.internal.MyClassificationResource
	at org.glassfish.jersey.internal.OsgiRegistry.classForNameWithException(OsgiRegistry.java:477) ~[?:?]
```

When looking into this issue the OsgiRegistry seems to have stored the following class in its classToBundleMapping:
org.exampleification.rest.internal.MyClassificationResource

This is caused by the incorrect replacement at https://github.com/eclipse-ee4j/jersey/blob/2.29-RC1-BRANCH/core-common/src/main/java/org/glassfish/jersey/internal/OsgiRegistry.java#L421

**Affected versions**
2.27+ (and possibly earlier versions)

jansupol:
That particular line is there since [2.0](https://github.com/jersey/jersey/blob/2.0/core-common/src/main/java/org/glassfish/jersey/internal/OsgiRegistry.java#L327), and it was working well so far. Can you provide us your test case?

WouterBanckenACA:
@jansupol I time-boxed tried to provide a test case but the tests for OSGi seem to be in a non-working state for me? (And ignored in Java 11)

I'm expecting the error to appear in the test suit if you rename for example org.glassfish.jersey.examples.osgi.helloworld.additional.resource to org.glassfish.jersey.examples.osgi.helloworld.classification.additional.resource.

[...]

PR #4235, opened 2019-08-28
Fix issue with OSGi when having package name starting with "class"

jansupol:


jansupol:
Fixes #4177

PR #4236, opened 2019-08-28
Issue 4208 - Fails to inject SecurityContext into Helloworld-CDI2-SE example

pa314159:


jansupol:
Please update copyright year in the next push

pa314159:
@jansupol, rebase or new commit?

inject/cdi2-se/src/main/java/org/glassfish/jersey/inject/cdi/se/bean/SupplierBeanBridge.java
```diff
@@ -143,4 +143,9 @@ public void destroy(Object instance, CreationalContext context) {
     public Class<? extends Annotation> getScope() {
         return binding.getScope() == null ? Dependent.class : transformScope(binding.getScope());
     }
+
+    @Override
+    public Class<?> getBeanClass() {
```
jansupol: if (binding.getContracts().size() == 1) {
    return binding.getContracts().iterator().next();
} else {
   return super.getBeanClass();
}

jansupol:
rebase is fine, small PR

inject/cdi2-se/src/main/java/org/glassfish/jersey/inject/cdi/se/bean/SupplierBeanBridge.java
pa314159: OK, but how is this different than SupplierThreadScopeBean that has the same expression, i.e.

    return (Class<?>) this.binding.getContracts().iterator().next()

jansupol: Right, but better safe than NPE

PR #4241, opened 2019-09-04
"A HTTP" -> "An HTTP"

oliver-gramberg:
Signed-off-by: Oliver Gramberg <oliver.gramberg@gmx.de>


jansupol:
While you have a good point of changing "A" to "An", we must not change the license of the file.

jansupol:
A copyright year should be updated, too

oliver-gramberg:
Oh my, what an embarrassing copy-paste error...

PR #4243, opened 2019-09-04
Fixes #4239 MediaType in method parameter not overridden by annotation

### LINKED ISSUE #4239: MP REST Client MethodModel overrides header set by method parameter
Consider the following interface definition:
```
@RegisterRestClient(baseUri = "http://tika:9998")
public interface TikaClient {

  @Produces(MediaType.TEXT_PLAIN)
  @Path("tika")
  @PUT
  String getText(
      @HeaderParam(HttpHeaders.CONTENT_TYPE) String contentType, StreamingOutput content);
}
```
This interface implements a call to Apache Tika to extract content from data streamed to the service. The content type is not known at compile time and cannot be set via annotation.

The following is observed when invoking the client:
```
PUT /tika HTTP/1.1
Accept: text/plain
Content-Type: */*
User-Agent: Jersey/2.29.payara-p2 (HttpUrlConnection 1.8.0_222)
Host: localhost:9998
Connection: keep-alive
Content-Length: 117736
```
Note the content type is **"\*/\*"** even though the application set the correct value of "application/pdf" in the method parameter of the interface. 

Initially I was going to point to `MethodModel.addCustomHeaders` where the next to last line adds a second value to the `Content-Type` header set in the method parameter. After following the call further, the call to `synchronousCall` (or `asynchronousCall`) creates the entity for the `PUT` request and sets the media type to `consumes[0]`. This media type is ultimately used to set the actual `Content-Type` header later in the request processing, overriding the custom header set by the method parameter.

The value in `consumes[0]` in this instance is the default value of `MediaType.WILDCARD` set at `InterfaceModel.Builder.consumes` when no annotation is present on the interface or method. This is the value observed in the request above. However, this is not the value the specification prescribes in the absence of the annotation:

> If no @ Consumes or @ Produces annotation is specified for a given request, it will default to javax.ws.rs.core.MediaType.APPLICATION_JSON ("application/json").

While the default value doesn't match the specification, the implementation requires that the value of the `Content-Type` header come from an annotation rather than being set using a header parameter in the method invocation, preventing dynamic content type setting.

Verdent:
Fixes #4239

### DIFF
--- a/ext/microprofile/mp-rest-client/src/main/java/org/glassfish/jersey/microprofile/restclient/MethodModel.java
+++ b/ext/microprofile/mp-rest-client/src/main/java/org/glassfish/jersey/microprofile/restclient/MethodModel.java
@@ -220,8 +220,7 @@ class MethodModel {
         if (entity != null
                 && !httpMethod.equals(GET.class.getSimpleName())
                 && !httpMethod.equals(DELETE.class.getSimpleName())) {
-            String contentType = (String) customHeaders.get(HttpHeaders.CONTENT_TYPE).get(0);
-            response = builder.method(httpMethod, Entity.entity(entity, contentType));
+            response = builder.method(httpMethod, Entity.entity(entity, getContentType(customHeaders)));
         } else {
             response = builder.method(httpMethod);
         }
@@ -245,8 +244,7 @@ class MethodModel {
         if (entity != null
                 && !httpMethod.equals(GET.class.getSimpleName())
                 && !httpMethod.equals(DELETE.class.getSimpleName())) {
-            String contentType = (String) customHeaders.get(HttpHeaders.CONTENT_TYPE).get(0);
-            theFuture = builder.async().method(httpMethod, Entity.entity(entity, contentType));
+            theFuture = builder.async().method(httpMethod, Entity.entity(entity, getContentType(customHeaders)));
         } else {
             theFuture = builder.async().method(httpMethod);
         }
@@ -273,6 +271,10 @@ class MethodModel {
         });
 
         return result;
+    }
+
+    private String getContentType(MultivaluedMap<String, Object> customHeaders) {
+        return (String) customHeaders.getFirst(HttpHeaders.CONTENT_TYPE);
     }
 
     @SuppressWarnings("unchecked")
@@ -329,13 +331,13 @@ class MethodModel {
 
     private MultivaluedMap<String, Object> addCustomHeaders(Object[] args) {
         MultivaluedMap<String, Object> result = new MultivaluedHashMap<>();
+        for (Map.Entry<String, List<String>> entry : resolveCustomHeaders(args).entrySet()) {
+            entry.getValue().forEach(val -> result.add(entry.getKey(), val));
+        }
         for (String produce : produces) {
             result.add(HttpHeaders.ACCEPT, produce);
         }
-        result.putSingle(HttpHeaders.CONTENT_TYPE, consumes[0]);
-        for (Map.Entry<String, List<String>> entry : resolveCustomHeaders(args).entrySet()) {
-            entry.getValue().forEach(val -> result.putSingle(entry.getKey(), val));
-        }
+        result.putIfAbsent(HttpHeaders.CONTENT_TYPE, Collections.singletonList(consumes[0]));
         return result;
     }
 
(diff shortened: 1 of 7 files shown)


ext/microprofile/mp-rest-client/src/main/java/org/glassfish/jersey/microprofile/restclient/MethodModel.java
```diff
@@ -210,13 +211,17 @@ private Form handleForm(Object[] args) {
...
+                                   Object entity,
+                                   Method method,
+                                   MultivaluedMap<String, Object> customHeaders) {
         Response response;
 
         if (entity != null
                 && !httpMethod.equals(GET.class.getSimpleName())
                 && !httpMethod.equals(DELETE.class.getSimpleName())) {
-            response = builder.method(httpMethod, Entity.entity(entity, consumes[0]));
+            String contentType = (String) customHeaders.get(HttpHeaders.CONTENT_TYPE).get(0);
```
tomas-langer: This should be:
`String contentType = (String) customHeaders.getFirst(HttpHeaders.CONTENT_TYPE);`

ext/microprofile/mp-rest-client/src/main/java/org/glassfish/jersey/microprofile/restclient/MethodModel.java
```diff
@@ -231,13 +236,17 @@ private Object synchronousCall(Invocation.Builder builder, Object entity, Method
...
+                                               Object entity,
+                                               Method method,
+                                               MultivaluedMap<String, Object> customHeaders) {
         CompletableFuture<Object> result = new CompletableFuture<>();
         Future<Response> theFuture;
         if (entity != null
                 && !httpMethod.equals(GET.class.getSimpleName())
                 && !httpMethod.equals(DELETE.class.getSimpleName())) {
-            theFuture = builder.async().method(httpMethod, Entity.entity(entity, consumes[0]));
+            String contentType = (String) customHeaders.get(HttpHeaders.CONTENT_TYPE).get(0);
```
tomas-langer: Duplication of code - maybe a new method to get `contentType` from the customHeaders map?

ext/microprofile/mp-rest-client/src/main/java/org/glassfish/jersey/microprofile/restclient/MethodModel.java
```diff
@@ -320,13 +329,13 @@ private WebTarget addMatrixParams(WebTarget webTarget, Object[] args) {
...
-        for (Map.Entry<String, List<String>> entry : resolveCustomHeaders(args).entrySet()) {
-            entry.getValue().forEach(val -> result.add(entry.getKey(), val));
-        }
         for (String produce : produces) {
             result.add(HttpHeaders.ACCEPT, produce);
         }
-        result.add(HttpHeaders.CONTENT_TYPE, consumes[0]);
+        result.putSingle(HttpHeaders.CONTENT_TYPE, consumes[0]);
+        for (Map.Entry<String, List<String>> entry : resolveCustomHeaders(args).entrySet()) {
+            entry.getValue().forEach(val -> result.putSingle(entry.getKey(), val));
```
tomas-langer: Unless your intention is to only use the last header value, you should do:
`entry.getValue().forEach(val -> result.add(entry.getKey(), val));`

tests/integration/microprofile/rest-client/pom.xml
```diff
@@ -89,7 +94,20 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-surefire-plugin</artifactId>
+                <dependencies>
+                    <dependency>
+                        <groupId>org.apache.maven.surefire</groupId>
+                        <artifactId>surefire-junit47</artifactId>
+                        <version>${surefire.version}</version>
```
tomas-langer: This should probably be in `dependencyManagement` section?

tests/integration/microprofile/rest-client/pom.xml
```diff
@@ -89,7 +94,20 @@
...
+                    </dependency>
+                    <dependency>
+                        <groupId>org.apache.maven.surefire</groupId>
+                        <artifactId>surefire-testng</artifactId>
+                        <version>${surefire.version}</version>
```
tomas-langer: dtto

tests/integration/microprofile/rest-client/pom.xml
jansupol: Are you sure? You would force every module using the surefire plugin to have these dependencies, too.

tomas-langer: No, I mean the version management, not plugin management

jansupol: I see. That's a good point

PR #4258, opened 2019-09-12
Loading keystore resource if location starts with /

Verdent:
Signed-off-by: David Kral <david.k.kral@oracle.com>


ext/microprofile/mp-rest-client/src/main/java/org/glassfish/jersey/microprofile/restclient/RestClientProducer.java
```diff
@@ -334,6 +334,9 @@ private InputStream locationToStream(String location) throws IOException {
             if (null == result) {
```
jansupol: Can you fix `from teh same classloader` at 332?

Verdent: Fixed

PR #4268, opened 2019-09-20
Use locale insensitive case changes to ensure user code doesn't break…

mjameson-se:
… in other locales (i.e. Turkey)

### DIFF
--- a/connectors/jdk-connector/src/main/java/org/glassfish/jersey/jdk/connector/internal/ProxyDigestAuthenticator.java
+++ b/connectors/jdk-connector/src/main/java/org/glassfish/jersey/jdk/connector/internal/ProxyDigestAuthenticator.java
@@ -22,6 +22,7 @@ import java.nio.charset.Charset;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.security.SecureRandom;
+import java.util.Locale;
 import java.util.logging.Logger;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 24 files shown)


connectors/jdk-connector/src/main/java/org/glassfish/jersey/jdk/connector/internal/ProxyDigestAuthenticator.java
```diff
@@ -326,7 +327,7 @@ public static Algorithm parse(String val) {
                 return Algorithm.UNSPECIFIED;
             }
             val = val.trim();
-            if (val.contains(MD5_SESS.md) || val.contains(MD5_SESS.md.toLowerCase())) {
+            if (val.contains(MD5_SESS.md) || val.contains(MD5_SESS.md.toLowerCase(Locale.ROOT))) {
```
jansupol: MD5_SESS.md.toLowerCase(Locale.ROOT) should be equal to MD5_SESS.md.toLowerCase() even in Turkish, right? "MD5-sess" is "MD5-sess" in any locales, or not?

mjameson-se: Frankly, I don't know. Even if its true for Turkish, is it true globally? We want this to work literally anywhere in the world. The Javadoc for toLowerCase does not have the same emphatic warning as does toUpperCase but it does give an example that is not intuitive (in Turkish I does not become i, just as i does not become I in the toUpper case). As a best-practice, we should use locale-independent case changes where it is appropriate whether or not we know for sure it could bite us.

jansupol: Wow, who would have guessed so...

mjameson-se:
> This looks good, but please update the copyright year to 2019

I'm not certain what you mean but I took a guess based on other files' copyright headers

PR #4272, opened 2019-09-24
AsyncInvocationinverceptors not properly created for each request

### LINKED ISSUE #4264: Microprofile Rest Client's AsyncInvocationInterceptor has wrong lifecycle
`AsyncInvocationInterceptors` should intercept single asynchronous invocation with following lifecycle:

- new interceptor is created for every async invocation
- `prepareContext()` is invoked just before the invocation leaves the main thread
- `applyContext()` is invoked on executing thread before invocation proceeds
- `removeContext()` is invoked on executing thread after invocation finishes

Current lifecycle is as follows

- one interceptor per client instance is created
- `prepareContext()` is invoked at the time of creation

The other methods are invoked as above. Effectively it means, that any interceptors have correct context only at first use, and only if client is already created in correct context (i. e. not on startup).

Verdent:
Fixes #4264 

Signed-off-by: David Kral <david.k.kral@oracle.com>

### DIFF
--- a/ext/microprofile/mp-rest-client/src/main/java/org/glassfish/jersey/microprofile/restclient/ExecutorServiceWrapper.java
+++ b/ext/microprofile/mp-rest-client/src/main/java/org/glassfish/jersey/microprofile/restclient/ExecutorServiceWrapper.java
@@ -113,9 +113,7 @@ class ExecutorServiceWrapper implements ExecutorService {
     private <T> Callable<T> wrap(Callable<T> task) {
         List<AsyncInvocationInterceptor> asyncInvocationInterceptors = asyncInterceptors.get();
         return () -> {
-            if (asyncInvocationInterceptors != null) {
-                asyncInvocationInterceptors.forEach(AsyncInvocationInterceptor::applyContext);
-            }
+            applyContextOnInterceptors(asyncInvocationInterceptors);
             return task.call();
         };
     }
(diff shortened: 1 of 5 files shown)


ext/microprofile/mp-rest-client/src/main/java/org/glassfish/jersey/microprofile/restclient/ExecutorServiceWrapper.java
```diff
@@ -112,15 +111,21 @@ public void execute(Runnable command) {
     }
 
     private <T> Callable<T> wrap(Callable<T> task) {
+        List<AsyncInvocationInterceptor> asyncInvocationInterceptors = asyncInterceptors.get();
         return () -> {
-            asyncInterceptors.forEach(AsyncInvocationInterceptor::applyContext);
+            if (asyncInvocationInterceptors != null) {
+                asyncInvocationInterceptors.forEach(AsyncInvocationInterceptor::applyContext);
```
pdudits: [Javadoc for the factory](https://github.com/eclipse/microprofile-rest-client/blob/master/api/src/main/java/org/eclipse/microprofile/rest/client/ext/AsyncInvocationInterceptorFactory.java#L31) says, that applyContext is called in order of descending priorities, so you might want to iterate in reverse order in the wrappers

Verdent: Thank you. Now it should be in proper order.

PR #4273, opened 2019-09-26
DocBook fixes

senivam:


docs/src/main/docbook/uris-and-links.xml
```diff
@@ -63,7 +63,7 @@ public class UsersResource {
     @GET
     @Produces("application/json")
     public JSONArray getUsersAsJsonArray() {
-        JSONArray uriArray = new JSONArray();
+        JSONArray uriArray = new JSONArray  ();
```
jansupol: This is a checkstyle violation:  '(' is preceded with whitespace.

senivam: reverted

PR #4275, opened 2019-09-26
Throwing NoContentException when InputStream is empty

mkarg:
Since JAX-RS 2.1 an MBR must react upon an empty stream either with an empty instance or with NoContentException, but Jersey's JSON-B MBR does neither - it simply throws ProcessingException:

_"In case the entity input stream is empty, the reader is expected to either return a Java representation of a zero-length entity or throw a javax.ws.rs.core.NoContentException in case no zero-length entity representation is defined for the supported Java type."_

### DIFF
--- a/media/json-binding/src/main/java/org/glassfish/jersey/jsonb/internal/JsonBindingProvider.java
+++ b/media/json-binding/src/main/java/org/glassfish/jersey/jsonb/internal/JsonBindingProvider.java
@@ -19,7 +19,6 @@ package org.glassfish.jersey.jsonb.internal;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
-import java.io.PushbackInputStream;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Type;
 
@@ -41,6 +40,7 @@ import javax.json.bind.JsonbException;
 
 import org.glassfish.jersey.jsonb.LocalizationMessages;
 import org.glassfish.jersey.message.internal.AbstractMessageReaderWriterProvider;
+import org.glassfish.jersey.message.internal.EntityInputStream;
 
 /**
  * Entity provider (reader and writer) for JSONB.
@@ -72,20 +72,10 @@ public class JsonBindingProvider extends AbstractMessageReaderWriterProvider<Obj
                            MediaType mediaType,
                            MultivaluedMap<String, String> httpHeaders,
                            InputStream entityStream) throws IOException, WebApplicationException {
-        if (entityStream.markSupported()) {
-            entityStream.mark(1);
-            if (entityStream.read() == -1) {
-                throw new NoContentException(LocalizationMessages.ERROR_JSONB_EMPTYSTREAM());
-            }
-            entityStream.reset();
-        } else {
-            final PushbackInputStream buffer = new PushbackInputStream(entityStream);
-            final int firstByte = buffer.read();
-            if (firstByte == -1) {
-                throw new NoContentException(LocalizationMessages.ERROR_JSONB_EMPTYSTREAM());
-            }
-            buffer.unread(firstByte);
-            entityStream = buffer;
+        final EntityInputStream entityInputStream =  new EntityInputStream(entityStream);
+        entityStream = entityInputStream;
+        if (entityInputStream.isEmpty()) {
+            throw new NoContentException(LocalizationMessages.ERROR_JSONB_EMPTYSTREAM());
         }
 
         Jsonb jsonb = getJsonb(type);
--- a/media/json-binding/src/main/java/org/glassfish/jersey/jsonb/internal/JsonBindingProvider.java
+++ b/media/json-binding/src/main/java/org/glassfish/jersey/jsonb/internal/JsonBindingProvider.java
@@ -19,6 +19,7 @@ package org.glassfish.jersey.jsonb.internal;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.io.PushbackInputStream;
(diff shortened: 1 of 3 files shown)


mkarg:
This PR proposes a solution which reconstitutes compliance with JAX-RS 2.1 with a two-step approach to detect an empty input stream:
* First it tries to use mark/reset as this is very fast, but not guaranteed to be supported by all kinds of input streams.
* If this is not possible, then it will actively buffer the first byte, which is rather slow, but will work with all kinds of input stream.

As long as we allow this MBR to be used with other JAX-RS implementations (not just Jersey), we need this PR, and we need this two-step approach. An alternative solution could be to clearly restrict this MBR to work on Jersey only (e. g. by documentation or even technical checks), in which case we could leave out one of the two steps as we then know for sure the type of input stream Jersey will run this MBR with.

mkarg:
Please see also https://github.com/eclipse-ee4j/jaxrs-api/issues/795.

mkarg:
@spericas @chkal I assume this is the code location where you want the topic discussed in https://github.com/eclipse-ee4j/jaxrs-api/issues/795 to be handled (rather than the JAX-RS spec). :-)

jansupol:
Is it possible to add a test checking the behaviour for an empty stream?

media/json-binding/src/main/java/org/glassfish/jersey/jsonb/internal/JsonBindingProvider.java
```diff
@@ -70,7 +72,24 @@ public Object readFrom(Class<Object> type, Type genericType,
...
+            entityStream.mark(1);
+            if (entityStream.read() == -1) {
+                throw new NoContentException("JSON-B cannot process empty input stream");
+            }
+            entityStream.reset();
+        } else {
+            final PushbackInputStream buffer = new PushbackInputStream(entityStream);
+            final int firstByte = buffer.read();
+            if (firstByte == -1) {
+                throw new NoContentException("JSON-B cannot process empty input stream");
```
jansupol: Please use Localization.properties

media/json-binding/src/main/java/org/glassfish/jersey/jsonb/internal/JsonBindingProvider.java
```diff
@@ -70,7 +72,24 @@ public Object readFrom(Class<Object> type, Type genericType,
                            MediaType mediaType,
                            MultivaluedMap<String, String> httpHeaders,
                            InputStream entityStream) throws IOException, WebApplicationException {
+        if (entityStream.markSupported()) {
+            entityStream.mark(1);
+            if (entityStream.read() == -1) {
+                throw new NoContentException("JSON-B cannot process empty input stream");
```
jansupol: Please use localization.properties

mkarg:
> Is it possible to add a test checking the behaviour for an empty stream?

Added unit test covering this issue with 4c6b2acadac2190daa783295eaa9ef2fc694d57e.

media/json-binding/src/main/java/org/glassfish/jersey/jsonb/internal/JsonBindingProvider.java
mkarg: I will do that. Stay tuned.

media/json-binding/src/main/java/org/glassfish/jersey/jsonb/internal/JsonBindingProvider.java
mkarg: I will do that. Stay tuned.

media/json-binding/src/main/java/org/glassfish/jersey/jsonb/internal/JsonBindingProvider.java
mkarg: Done with bccd9008528477587b2eb340cbaf0a8762bcdadb.

media/json-binding/src/main/java/org/glassfish/jersey/jsonb/internal/JsonBindingProvider.java
mkarg: Done with bccd9008528477587b2eb340cbaf0a8762bcdadb.

senivam:
just another thought... why not to align everything to a Stream which supports marks and to separate to a private utility method? So, the main method would look like:
```
@Override
    public Object readFrom(Class<Object> type, Type genericType,
                           Annotation[] annotations,
                           MediaType mediaType,
                           MultivaluedMap<String, String> httpHeaders,
                           InputStream entityStream) throws IOException, WebApplicationException {

        validateInputStream(entityStream);

        Jsonb jsonb = getJsonb(type);

        try {
            return jsonb.fromJson(entityStream, genericType);
        } catch (JsonbException e) {
            throw new ProcessingException(LocalizationMessages.ERROR_JSONB_DESERIALIZATION(), e);
        }
    }
```
and then private utility method:
```
    private static void validateInputStream(InputStream entityStream) throws IOException {
        final InputStream streamWithMark = (entityStream.markSupported()) ? entityStream : new BufferedInputStream(entityStream);

        streamWithMark.mark(1);
        if (streamWithMark.read() == -1) {
            throw new NoContentException(LocalizationMessages.ERROR_JSONB_EMPTYSTREAM());
        }
        streamWithMark.reset();
        entityStream = streamWithMark;
    }
```
what do you think?

mkarg:
> just another thought... _previous comment was wrong :)_

Cool optimization! 😎 If you are sure it is as safe and as fast I would be happy to accept your PR ontop! 👍

senivam:
@mkarg, OK, I've submitted a PR to your branch :)

PR #4277, opened 2019-10-01
HK2 to skip fields injected by CDI in non bean-defining-annotated beans

jansupol:


jansupol:
This is for bean-discovery-mode="none" and bean-discovery-mode="annotated"

ext/cdi/jersey-cdi1x/src/main/java/org/glassfish/jersey/ext/cdi1x/internal/InjecteeSkippingAnalyzer.java
```diff
@@ -98,4 +105,12 @@ public InjecteeSkippingAnalyzer(ClassAnalyzer defaultAnalyzer,
 
         return compositeResult;
     }
+
+    private void addCdiInjectedFieldsToSkip(Set<Field> skippedFields, Set<Field> originalFields) {
+        for (Field field : originalFields) {
+            if (field.getAnnotation(Inject.class) != null && !beanManager.getBeans(field.getType()).isEmpty()) {
```
tomas-langer: If `@Inject` is defined on a field or constructor, CDI fails if it cannot inject it - so we should never get to this point with an "uninjected" field of such type.
Also the `beanManager.getBeans` ignores qualifiers, so that part of the if statement should be removed.

PR #4280, opened 2019-10-03
Move CDI integration tests to a common CDI-Integration module

jansupol:


jansupol:
Cleanup in the integration tests module

PR #4281, opened 2019-10-04
Less requirements for tests not to fail intermittently on Travis

jansupol:

### DIFF
--- a/tests/stress/pom.xml
+++ b/tests/stress/pom.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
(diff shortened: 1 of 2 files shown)


jansupol:
- `BroadcasterTest` fails on timeout - the response is sometimes not received within a second
- `MultiThreadingAggregatedReservoirTest` requires more than 1GB of RAM when TEST_DURATION_MILLIS = 10_000 as noted
- maven-surefire-plugin _forkMode_ is [deprecated](https://maven.apache.org/surefire/maven-surefire-plugin/examples/fork-options-and-parallel-execution.html) and the default for _forkMode=always_ is _forkCount=1_, _reuseForks=false_

jansupol:
Keeping just the OOME fix as the other fix did not help

PR #4283, opened 2019-10-09
Enable to use AsyncInvoker in Rx client

jansupol:


jansupol:
- Fixes #4225. Uses AsyncInvoker in Rx. 
- Separates redundant code into a common superclass.
- Allows for using InvocationCallback when using methods from AsyncInvoker if needed.

PR #4305, opened 2019-10-28
Fixes #4304: ResourceConfig not properly using specified ClassLoader

### LINKED ISSUE #4304: ResourceConfig not properly using specified ClassLoader
The issue is slightly related to #3227, but not exactly. In #3227 the issue is happening when loading jersey via a different ClassLoader, my issue is happening when routes/packages to scan use a different ClassLoader than the one jersey is loaded with. Eventually also related to #3515, which does not fix my problem.

Currently, you may register a package via ```ResourceConfig#packages```, but if the package is not found, it's just ignored without an error or notice. 
If you register a package using ```ResourceConfig#registerFinder``` (using the "default" PackageNamesScanner with a custom ClassLoader specified), the classes/routes are discovered but an error will raise once it'll try to load the class, because it's loaded using the default ClassLoader anyways and not the one, specified in ```PackageNamesScanner```. I know you may specify a ClassLoader in the ```ResourceConfig``` using ```#setClassLoader``` itself, but then that ClassLoader will be used for all packages, I need one ClassLoader per package.

I already fixed this issue and will create a PR in just a second. I hope you will merge that PR because I don't wanna maintain my own fork of jersey. As you can see in the PR, it's currently only for the ```PackageNamesScanner``` class. I also thought about adding a method to the ```ResourceFinder``` interface, but discarded that because I don't know whatelse uses the ResourceFinder class and whether a ```getClassLoader``` method would be useful there - I don't guess so, that's why I did it the way I did.

d0x7:
This PR fixes the problem (eventually a bug?) in the issue #4304.

Please review, tell me what to change and then eventually merge the PR. 
If you won't merge it, I'd like to know why, because I don't wanna maintain my own fork.

### DIFF
--- a/core-server/src/main/java/org/glassfish/jersey/server/ResourceConfig.java
+++ b/core-server/src/main/java/org/glassfish/jersey/server/ResourceConfig.java
@@ -656,6 +656,27 @@ public class ResourceConfig extends Application implements Configurable<Resource
     }
 
     /**
+     * Adds array of package names which will be used to scan for components.
+     * <p/>
+     * Package scanning ignores an inheritance and therefore {@link Path} annotation
+     * on parent classes and interfaces will be ignored.
+     * <p/>
+     * @param recursive defines whether any nested packages in the collection of specified
+     *                  package names should be recursively scanned (value of {@code true})
+     *                  as part of the package scanning or not (value of {@code false}).
+     * @param classLoader defines the classloader used for scanning the packages and loading the classes.
+     * @param packages  array of package names.
+     * @return updated resource configuration instance.
+     * @see #packages(String...)
+     */
+    public final ResourceConfig packages(final boolean recursive, final ClassLoader classLoader, final String... packages) {
+        if (packages == null || packages.length == 0) {
+            return this;
+        }
+        return registerFinder(new PackageNamesScanner(classLoader, packages, recursive));
+    }
+
+    /**
      * Adds array of file and directory names to scan for components.
      * <p/>
      * Any directories in the list will be scanned recursively, including their sub-directories.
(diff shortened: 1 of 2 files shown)


core-server/src/main/java/org/glassfish/jersey/server/ResourceConfig.java
```diff
@@ -877,11 +898,21 @@ public boolean isRegistered(final Class<?> componentClass) {
...
+        final AnnotationAcceptingListener parentAfl =
                 AnnotationAcceptingListener.newJaxrsResourceAndProviderListener(_state.getClassLoader());
+
         for (final ResourceFinder resourceFinder : rfs) {
+            AnnotationAcceptingListener afl = parentAfl;
+
+            // Workaround because otherwise classes from a different classloader can't be loaded
+            if (resourceFinder instanceof PackageNamesScanner) {
+                final ClassLoader classLoader = ((PackageNamesScanner) resourceFinder).getClassloader();
+                afl = AnnotationAcceptingListener.newJaxrsResourceAndProviderListener(classLoader);
```
jansupol: Do not do this when the default ClassLoader:
```
if (!getClassLoader().equals(classLoader) {
   afl = AnnotationAcceptingListener.newJaxrsResourceAndProviderListener(classLoader);
}
```

core-server/src/main/java/org/glassfish/jersey/server/ResourceConfig.java
```diff
@@ -896,10 +927,11 @@ public boolean isRegistered(final Class<?> componentClass) {
                         }
                     }
                 }
+
+                result.addAll(afl.getAnnotatedClasses());
```
jansupol: `if (parentAfl != afl)`

core-server/src/main/java/org/glassfish/jersey/server/ResourceConfig.java
```diff
@@ -896,10 +927,11 @@ public boolean isRegistered(final Class<?> componentClass) {
...
             }
         }
 
-        result.addAll(afl.getAnnotatedClasses());
```
jansupol: +`result.addAll(parentAfl.getAnnotatedClasses());`

jansupol:
See [Wiki](https://github.com/eclipse-ee4j/jersey/wiki/Missing-ECA,-Missing-Sign-off)

d0x7:
> See [Wiki](https://github.com/eclipse-ee4j/jersey/wiki/Missing-ECA,-Missing-Sign-off)

Apparently I forgot to add the "Signed off by: *" to the commit. According to the wiki, it needs to be included to every commit I made. What should I do now, because the two commits don't have it, even tho it's signed?

d0x7:
See #4306, sorry again.

PR #4307, opened 2019-10-29
Use Spring Context 4 in the Spring integration test

danielkec:


tests/integration/spring4/src/main/resources/applicationContext.xml
```diff
@@ -32,6 +32,15 @@
         </property>
```
senivam: please update copyright year

PR #4309, opened 2019-10-30
Spring 5 integration tests

danielkec:

### DIFF
--- a/tests/integration/spring5/src/main/java/org/glassfish/jersey/server/spring/test/AccountJerseyResource.java
+++ b/tests/integration/spring5/src/main/java/org/glassfish/jersey/server/spring/test/AccountJerseyResource.java
@@ -1,0 +1,164 @@
+/*
+ * Copyright (c) 2013, 2019 Oracle and/or its affiliates. All rights reserved.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0, which is available at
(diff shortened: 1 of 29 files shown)


tests/integration/spring5/src/main/java/org/glassfish/jersey/server/spring/test/AccountJerseyResource.java
```diff
@@ -0,0 +1,164 @@
...
+import javax.inject.Named;
+import javax.servlet.http.HttpServletRequest;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
+
+/**
+ * Jersey managed JAX-RS resource for testing jersey-spring.
+ *
+ * @author Marko Asplund (marko.asplund at yahoo.com)
```
senivam: please do not mention any author in new files...

jansupol: On the contrary, as long as the file is a copy of a file copied over all the way from the spring-3 module, I think it makes sense to keep the original author, as he might be possibly contacted if needed.

danielkec: Reverted back

PR #4310, opened 2019-11-04
Spring 5 integration tests with java config

danielkec:


tests/integration/spring5/src/main/java/org/glassfish/jersey/server/spring/test/resource/AccountSpringResource.java
```diff
@@ -0,0 +1,172 @@
...
+
+/**
+ * Spring managed JAX-RS resource for testing jersey-spring.
+ *
+ * @author Marko Asplund (marko.asplund at yahoo.com)
+ */
+@Path("/spring/account")
+@Singleton
+//@Component
+public class AccountSpringResource {
```
danielkec: @jansupol This is a problem I didn't find a way around yet, seems like with WebApplicationInitializer is Spring context started before HK2 bridge has anything to offer.

PR #4311, opened 2019-11-05
Automatically serialize/deserialize Java 8 datatype

### LINKED ISSUE #3637: Automatically serialize/deserialize Java 8 datatype with `jersey-media-json-jackson`
As of today, in order to serialize/deserialize the "new" datatypes introduced in Java 8 (e.g. `Optional<T>`), you need to provide a custom `ObjectMapper` (as explained [here](https://jersey.github.io/documentation/latest/media.html#d0e8188)) and register the corresponding Jackson extension for handling those types (i.e. [jackson-modules-java8](https://github.com/FasterXML/jackson-datatype-jdk8) for Jackson < 2.8.5 and [jackson-modules-java8](https://github.com/FasterXML/jackson-modules-java8) for Jackson >= 2.8.5).

I guess it would be nice to provide this functionality out of the box.

Praveen2112:
This patch adds support for automatically serializing and deserialzing for new "datatypes" introduced in Java 8 for jersey-media-json-jackson module.

Fixes #3637.

### DIFF
--- a/media/json-jackson/pom.xml
+++ b/media/json-jackson/pom.xml
@@ -117,6 +117,10 @@
             <groupId>com.fasterxml.jackson.module</groupId>
             <artifactId>jackson-module-jaxb-annotations</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.datatype</groupId>
+            <artifactId>jackson-datatype-jdk8</artifactId>
+        </dependency>
 
         <dependency>
             <groupId>junit</groupId>
--- a/media/json-jackson/src/main/java/org/glassfish/jersey/jackson/internal/jackson/jaxrs/json/JsonMapperConfigurator.java
+++ b/media/json-jackson/src/main/java/org/glassfish/jersey/jackson/internal/jackson/jaxrs/json/JsonMapperConfigurator.java
@@ -2,6 +2,7 @@ package org.glassfish.jersey.jackson.internal.jackson.jaxrs.json;
 
 import java.util.ArrayList;
 
+import com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
 import org.glassfish.jersey.jackson.internal.jackson.jaxrs.cfg.Annotations;
 import org.glassfish.jersey.jackson.internal.jackson.jaxrs.cfg.MapperConfiguratorBase;
 
@@ -44,6 +45,8 @@ public class JsonMapperConfigurator
     public synchronized ObjectMapper getDefaultMapper() {
         if (_defaultMapper == null) {
             _defaultMapper = new ObjectMapper();
+            // We can remove this once we move to Jackson 3.0
+            _defaultMapper.registerModule(new Jdk8Module());
             _setAnnotations(_defaultMapper, _defaultAnnotationsToUse);
         }
         return _defaultMapper;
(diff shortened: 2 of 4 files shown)


jansupol:
This PR does not look right:
- the newly used module `jackson-datatype-jdk8` is [deprecated](https://github.com/FasterXML/jackson-datatype-jdk8).
- the class `JsonMapperConfigurator.java` which is being modified is repackaged `JsonMapperConfigurator.java` from [Jackson ](https://github.com/FasterXML/jackson-jaxrs-providers/blob/jackson-jaxrs-providers-2.10.1/json/src/main/java/com/fasterxml/jackson/jaxrs/json/JsonMapperConfigurator.java)- and Jackson itself does not do this in 2.9 nor in 2.10 branch. Have you considered to create the PR to Jackson itself?
- this PR cries for a test - adding new functionality which allows for using `Optional`
- the PR breaks jersey-tests-osgi-functional tests

Praveen2112:
hi, thanks for your comment

> the newly used module jackson-datatype-jdk8 is deprecated.
I have checked it from [Jackson modules jdk8](https://github.com/FasterXML/jackson-modules-java8/tree/master/datatypes) but it looks like they have pointed the wrong dependency.

I'll add test for the same and debugging on why ` jersey-tests-osgi-functional` test fails

jansupol:
`jersey-tests-osgi-functional` likely failed because you need to modify the `JsonJacksonTest` to contain `mavenBundle().groupId("com.fasterxml.jackson.core").artifactId(jackson-datatype-jdk8").versionAsInProject()`.

However, I am not convinced we should modify the repackaged `JsonMapperConfigurator`.

Praveen2112:
Actually we can modify the `JsonMapperConfigurator` in the jackson library but the problem is Jackson-Jaxrs library can also run in Java version before 1.8 so adding them there might break it. And I think from 3.0 version Jackson runs only Java version above 8 so Jdk-8 module would be a part of it. (But it is not released ). So once we move to the 3.0 of Jackson we wouldn't require this dependency

jansupol:
The preferred way to do this should be to extend `JacksonJsonProvider` and pass the `ObjectMapper` with predefined `Jdk8Module` to be registered. But I am not sure this should be a class in Jersey, it sounds more like a user-specific class, especially when Jackson 3 would not need this.

What might be better is an extended JacksonJsonProvider that would get `com.fasterxml.jackson.databind.Module` from the `@Context Configuration` and pass it to JacksonJsonProvider, so that you could use the configuration mechanism to specify the needed modules.

Praveen2112:
Okay. Will close it and work on the latter approach. Thank you for your insights.

jansupol:
Actually, now I look at their repo, they have it as a [service](https://github.com/FasterXML/jackson-modules-java8/tree/master/datatypes/src/main/resources/META-INF/services). So it should be automatically loaded when on classpath by Jackson somehow.

jansupol:
I cannot find a code that uses [ServiceLoader](https://docs.oracle.com/javase/7/docs/api/java/util/ServiceLoader.html) and in the tests, the `Module` is registered manually. If that is not handled by Jackson itself, maybe it can be done in Jersey `JacksonFeature`.

Praveen2112:
But aren't those features are a part of Jackson 3.0 ?

jansupol:
Maybe. But 3.0 may have a few months before release and the services are available for Jackson 2.x.

Issue #4267, opened 2019-09-20
API for for async context propagation in Client API


pdudits:
Currently the client offers no API for capturing calling thread context when invocations are turned into async on RX ones. ClientRequestFilters are first executed on executing thread.

Proposal:

New component that can be registered into `ClientBuilder` with roughly following API:

```
interface AsyncContextProvider {
   void initAsyncContext(RequestContext asyncContext);
}
```

These providers would be looked up and invoked by `AsyncInvoker` or before turning an invocation into `RxInvoker`, to allow for capturing additional calling thread context into request context.

jansupol:
- There is [Invocation.Builder#property](https://javaee.github.io/javaee-spec/javadocs/javax/ws/rs/client/Invocation.Builder.html#property-java.lang.String-java.lang.Object-) to set properties to be accessible on the `ClientRequestFiter`

- RxInvoker has an [RxProvider](https://javaee.github.io/javaee-spec/javadocs/javax/ws/rs/client/RxInvokerProvider.html) spi that basically makes it hard to make sure the `AsyncContextProvider` would run in a non-executor thread. Mostly, the RxProvider runs the `SyncInvoker` in a thread given by the executor service, so it is late already. An option is to run the `AsyncContextProvider` as soon as the `async()`/`rx()` is executed. But the request may never be actually invoked, so it does not sound right to execute that at the point.

- For AsyncInvoker this can be done
- For default `RxInvoker` this can be done, too.  

@pdudits What do you think the right solution for `RxInvoker` would be?

jansupol:
I have created #4301, which supersedes this requirement. Does it look good to you?

pdudits:
Oh, forgot to react last time, the PR looks interesting, I will try it out.

jansupol:
@pdudits Thanks, I am open to suggestions for improvements.

PR #4313, opened 2019-11-13
new InvocationBuilderListener SPI

jansupol:

### DIFF
--- a/core-client/src/main/java/org/glassfish/jersey/client/InvocationBuilderListenerStage.java
+++ b/core-client/src/main/java/org/glassfish/jersey/client/InvocationBuilderListenerStage.java
@@ -1,0 +1,126 @@
+/*
+ * Copyright (c) 2011, 2019 Oracle and/or its affiliates. All rights reserved.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0, which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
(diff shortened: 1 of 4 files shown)


jansupol:
This new SPI interface allows for 
- creating decorators that modify the requests (`Invocation.Builder`) as needed.
- setting the ClientRequestContext properties (in the main thread rather than in the executor service thread when invoked by RxInvoker and set by Filter), possibly registered automatically using `AutoDiscoverable` interface
- providing functionality already requested by the MP [RestClientBuilderListener](https://github.com/eclipse/microprofile-rest-client/blob/master/api/src/main/java/org/eclipse/microprofile/rest/client/spi/RestClientBuilderListener.java)

jansupol:
@pdudits Would this help you with #4267?

jansupol:
@pdudits The `InvocationBuilderListener` is to be called when the `InvocationBuilder` is created, the same as the `RestClientBuilderListener`. Is it good enough for you to have it called at the time, i.e. long time before the `rx()` is called?

pdudits:
I'll have a detailed look tomorrow, at first sight it looks good. One usually uses RestClientBuilder to build client for many invocations (in real applications, contrary to tests), so I think the relation of lifecycles are but different

pdudits:
I tested this API with the same four usecases as in #4301, code is in [this branch](https://github.com/pdudits/patched-src-jersey/commits/pr-4313-experiments).

All of these usecases are satisfied by this, so I'm quite happy.

jansupol:
@tomas-langer I have added getters to the headers, configuration, and URI.

Issue #4315, opened 2019-11-14
@QueryParam annotated parameters doesn't seem to be using the ParamConverters as opposed to the @PathParam annotated parameters


dabla:
It seems that when using the JerseyClient, the @QueryParam annotated parameters doesn't seem to be using the ParamConverters as opposed to the @PathParam annotated parameters which do use those ParamConverters that are registered in the ClientConfig.  Is this a normal behaviour in JAX-RS?  Before we were using RestEasy and there we hadn't had that issue and @QueryParam where also converted using our custom registered ParamConverters.

On the server side in Grizzly, the @QueryParams are correctly converted using our custom ParamConverters.

dabla:
`public class LocationResourceImplIntegrationTest extends JerseyTest {
    private static final LocationId LOCATION_ID = locationId("2");
    private static final SubLocationId SUB_LOCATION_ID = subLocationId(1L);
    private static final Location LOCATION = new Location(LOCATION_ID, belgianNetId(), "symbolicName", EMPTY_TRANSLATED_VALUE, EMPTY_TRANSLATED_VALUE, false);
    private static final LocationDto LOCATION_DTO = new LocationDto(LOCATION_ID, MICRO, of(LOCATION_ID), EMPTY_TRANSLATED_VALUE, EMPTY_TRANSLATED_VALUE, "symbolicName", belgianNetId(), absent(), emptySet(), emptyList(), VERY_DARK_BLUE.hexValue(), aUUID(randomUUID()), false, false, BORDER, REGIONAL_BORDER, IMPORTANT, IMPORTANT);

    @Override
    protected void configureClient(ClientConfig config) {
        ApplicationContext contextConfig = applicationContext();
        config.register(contextConfig.getBean(LocationIdConverter.class))
              .register(contextConfig.getBean(SubLocationIdConverter.class))
              .register(contextConfig.getBean(ParamConverterProvider.class))
              .register(contextConfig.getBean(MarshallingFeature.class))
              .register(contextConfig.getBean(ClientExceptionInterceptor.class));
    }

    @Override
    public ResourceConfig configure() {
        ApplicationContext contextConfig = applicationContext();
        return new ResourceConfig()
                .property("contextConfig", contextConfig)
                .register(contextConfig.getBean(LocationIdConverter.class))
                .register(contextConfig.getBean(SubLocationIdConverter.class))
                .register(contextConfig.getBean(ParamConverterProvider.class))
[...]

PR #4316, opened 2019-11-15
When matching ParamConverter is found for given object in queryParam …

dabla:
…or matrixParam then object is converted to a String instead of simply invoking the toString method on the object (#4315)

Signed-off-by: David Blain <info@dabla.be>


jansupol:
I wonder if `JerseyWebTarget` is the right place to utilize `ParamConverter`. That would be used just on the client.

Would it be better to use `ParamConverters` in `JerseyUriBuilder`? Pass the configuration to the `JerseyUriBuilder` and utilize the `ParamConverters` there. Then the `RuntimeDelegateDecorator` could pass the config to the `JerseyUriBuilder` instance and the `ParamConverter` could be used later on the server side, too.

dabla:
> I wonder if `JerseyWebTarget` is the right place to utilize `ParamConverter`. That would be used just on the client.
> 
> Would it be better to use `ParamConverters` in `JerseyUriBuilder`? Pass the configuration to the `JerseyUriBuilder` and utilize the `ParamConverters` there. Then the `RuntimeDelegateDecorator` could pass the config to the `JerseyUriBuilder` instance and the `ParamConverter` could be used later on the server side, too.

True, that's what we also discovered today, as the ParamConverters for @PathParams are also ignore by the JerseyClient ,those could only be converted in the UriBuilder.  Will do an additional commit and move it to UriBuilder.

dabla:
Question: Is it allowed to use assertj instead of Junit asserts and to upgrade Mockito?

jansupol:
> Question: Is it allowed to use assertj instead of Junit asserts and to upgrade Mockito?

Generally, I am not a big fan of mocking frameworks. With JDK making the reflection usage harder, the newer versions of the mocking frameworks may have bigger requirements. We saw that with JMockit, which first required the agent to be run and then it stopped mocking the java.util classes. The agent defined in the pom file then is executed for every test, not just the mocked ones, which made the debugging harder as the agent came into action quite often. We had to separate the tests using Jmockit into a separate module. Not sure what the future of Mockito would be, but I'd better stay safe. For many use cases, JDK proxy is a fine replacement with just a couple of additional lines.

Regarding the new dependencies, add them only when necessary. First, each of them needs to be approved by the legal process, for each never version that is about to be used. Second, they potentially can bring a security vulnerability which we need to deal with ASAP, usually.

dabla:
I see that for the latest commit, following test is failing, dunno how to fix this one :-(

FeatureConstraintTest.test:164 ClientSever Features should have been configured
Expected: is <11110>
     but: was <22110>

Issue #4323, opened 2019-11-23
Add @PropertiesClass annonatated classes to config properties


jansupol:

jansupol:
Currently, only ClientProperties, ServerProperties, and CommonProperties properties are used for properties defined by system properties grabbed when SecurityManager is on.
It would make sense to try other @PropertiesClass annotated property classes.

jbescos:
If I understand it correctly:
The current class SystemPropertiesConfiguration.getExpectedSystemProperties() is only loading the fields from CommonProperties, CommonProperties and ClientProperties.

And we want to add any existing or future class having the annotation @PropertiesClass or 
@Property

I will develop it in this way:
Adding a new method in ReflectionHelper: List<Class<?>> findClassesInPackageWhere(Package root, Predicate<Class<?>> ... predicates)

So it will return the classes under a package given the predicates (having the annotation @PropertiesClass or @Property)

PR #4327, opened 2019-11-28
Allow to use additional properties with security manager/4323

jbescos:
This pull request is related to this issue:
https://github.com/eclipse-ee4j/jersey/issues/4323

The properties are not loaded by reflection, they are explicitly set in a list. The reason I didn't use reflection is we only want to load jersey properties, so we already know them.

In case we create a new property class and we forget to add it in the list, there is a new test that should fail.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,7 +25,6 @@ nb-configuration.xml
 .settings/*
 .project
 .classpath
-bin/
 
 # Maven plugins noise
 dependency-reduced-pom.xml
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,7 @@ nb-configuration.xml
 .settings/*
 .project
 .classpath
+bin/
 
 # Maven plugins noise
 dependency-reduced-pom.xml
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -25,6 +25,7 @@ nb-configuration.xml
 .settings/*
 .project
 .classpath
+bin/
```
jansupol: this bin should have not been here

tests/integration/property-check/src/test/java/org/glassfish/jersey/internal/config/SystemPropertiesConfigurationModelTest.java
```diff
@@ -0,0 +1,129 @@
...
+        Stream<Class<?>> steam = classpath.getTopLevelClassesRecursive(packageRoot).stream()
+                .filter(classInfo -> !BLACK_LIST_CLASSES.contains(classInfo.getName()))
+                .map(classInfo -> {
+                    try {
+                        return Class.forName(classInfo.getName());
+                    } catch (ClassNotFoundException e) {
+                        return Void.class;
+                    }
+                });
+        for (Predicate<Class<?>> predicate : predicates) {
```
senivam: since you are already using streams I suppose it would be appropriate to rewrite this FOR cycle with :

.filter(Arrays.stream(predicates).reduce(x->true, Predicate::and))

and then you can just return 

.collect(Collectors.toList())

Issue #4332, opened 2019-12-03
HttpUrlConnector relies on DefaultSSLSocketFactory value which is not a constant


Andremoniy:
https://github.com/eclipse-ee4j/jersey/blob/master/core-client/src/main/java/org/glassfish/jersey/client/internal/HttpUrlConnector.java#L302-L304

`HttpUrlConnector` sets the custom SSLFactory only when the one in the current `HttpsURLConnection` instance is the default one:

```
if (HttpsURLConnection.getDefaultSSLSocketFactory() == suc.getSSLSocketFactory()) {
                // indicates that the custom socket factory was not set
                suc.setSSLSocketFactory(sslSocketFactory.get());
            }
```

However, there is no guarantee that `HttpsURLConnection.getDefaultSSLSocketFactory()` remains a constant. Furthermore, there is an open bug about this issue:
https://bugs.openjdk.java.net/browse/JDK-8160347

This leads to the situation when the underlying `HttpsURLConnection` doesn't receive a custom SSLSocketFactory from the client.

The issue is certainly related to this one: https://github.com/eclipse-ee4j/jersey/issues/3293 but the fix was not proposed or has not been merged, so the issue is still remains.

Several screenshots with an evidence of this issue I've caught in debug.
On this screenshot we see that the current instance of `HttpUrlConnector` has a custom `SSLSocketFactory` instance with a custom `TrustManager` with one trusted certificate.
<img width="1374" alt="Screen Shot 2019-12-03 at 16 10 03" src="https://user-images.githubusercontent.com/2354206/70063109-8a4aad80-15e7-11ea-9b1c-ab4608742550.png">

The second screenshot demonstrates, that the `uc` object still has a default `SSLSocketFactory` with the default `TrustManager` containing 88 certificates.

<img width="1373" alt="Screen Shot 2019-12-03 at 16 12 30" src="https://user-images.githubusercontent.com/2354206/70063440-0c3ad680-15e8-11ea-91dc-2c6b4e872bda.png">

The final screenshot shows that even though both `HttpsURLConnection.getDefaultSSLSocketFactory()` and `suc.getSSLSocketFactory()` have the same `SSLContext`, the instances themselves are not equal.

<img width="1056" alt="Screen Shot 2019-12-03 at 16 16 04" src="https://user-images.githubusercontent.com/2354206/70063539-4015fc00-15e8-11ea-8eff-0bf2282c54e5.png">

Andremoniy:
[...]

PR #4335, opened 2019-12-03
This commit changes the way how the check for the default SSLSocketFa…

Andremoniy:
A proposal to fix https://github.com/eclipse-ee4j/jersey/issues/4332
Despite the fact that `HttpsURLConnection.getDefaultSSLSocketFactory()` result is not always unique, all default `SSLSocketFactory` share the same `SSLContextImpl`:

```
public final class SSLSocketFactoryImpl extends SSLSocketFactory {
    private final SSLContextImpl context;

    public SSLSocketFactoryImpl() throws Exception {
        this.context = DefaultSSLContext.getDefaultImpl();
    }
```

This is a proposed solution to use this additional check.

Signed-off-by: Andrey Lebedev <andremoniy@gmail.com>

### DIFF
--- a/core-client/src/main/java/org/glassfish/jersey/client/internal/HttpUrlConnector.java
+++ b/core-client/src/main/java/org/glassfish/jersey/client/internal/HttpUrlConnector.java
@@ -299,11 +299,34 @@ public class HttpUrlConnector implements Connector {
                 suc.setHostnameVerifier(verifier);
             }
 
-            if (HttpsURLConnection.getDefaultSSLSocketFactory() == suc.getSSLSocketFactory()) {
+            if (isDefaultSSLSocketFactory(suc)) {
                 // indicates that the custom socket factory was not set
                 suc.setSSLSocketFactory(sslSocketFactory.get());
             }
         }
+    }
+
+    static boolean isDefaultSSLSocketFactory(HttpsURLConnection suc) {
+        final boolean fastCheck = HttpsURLConnection.getDefaultSSLSocketFactory() == suc.getSSLSocketFactory();
+        if (fastCheck) {
+            return true;
+        }
+        final Object defaultContext = getContextObject(HttpsURLConnection.getDefaultSSLSocketFactory());
+        if (defaultContext == null) {
+            return false;
+        }
+        final Object sucContext = getContextObject(suc.getSSLSocketFactory());
+        return defaultContext == sucContext;
+    }
+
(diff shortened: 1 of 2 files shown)


core-client/src/main/java/org/glassfish/jersey/client/internal/HttpUrlConnector.java
```diff
@@ -299,13 +299,36 @@ protected void secureConnection(final JerseyClient client, final HttpURLConnecti
...
 
-            if (HttpsURLConnection.getDefaultSSLSocketFactory() == suc.getSSLSocketFactory()) {
+            if (isDefaultSSLSocketFactory(suc)) {
                 // indicates that the custom socket factory was not set
                 suc.setSSLSocketFactory(sslSocketFactory.get());
             }
         }
     }
 
+    static boolean isDefaultSSLSocketFactory(HttpsURLConnection suc) {
```
thibaudledent: About `suc`, I would suggest to use a more meaningful variable name

core-client/src/main/java/org/glassfish/jersey/client/internal/HttpUrlConnector.java
```diff
@@ -299,13 +299,36 @@ protected void secureConnection(final JerseyClient client, final HttpURLConnecti
...
+        final Object sucContext = getContextObject(suc.getSSLSocketFactory());
+        return defaultContext == sucContext;
+    }
+
+    static Object getContextObject(SSLSocketFactory sslSocketFactory) {
+        try {
+            final Field contextField = sslSocketFactory.getClass().getDeclaredField("context");
+            contextField.setAccessible(true);
+            return contextField.get(sslSocketFactory);
+        } catch (NoSuchFieldException | IllegalAccessException ignore) {
```
thibaudledent: Maybe it would be worth explaining why this exception is ignored with a comment or a log.

senivam (approved):

jbescos:
> What is preventing this fix from moving forward?

The PR still has one comment that requires a fix, related to: `contextField.setAccessible(true);`

dcitron:
> > What is preventing this fix from moving forward?
> 
> 
> 
> The PR still has one comment that requires a fix, related to: `contextField.setAccessible(true);`

Yeah, on further reflection (haha) it would be better if we can come up with a fix that doesn't require reflection. I'll think about it too.

Issue #4325, opened 2019-11-27
Build Jersey on JDK13


jansupol:

jansupol:
JDK 11 made certain Classes [deprecated](https://docs.oracle.com/en/java/javase/11/docs/api/jdk.javadoc/com/sun/javadoc/Tag.html) and they were removed in JDK 13. 
See the [migration guide](https://docs.oracle.com/javase/9/docs/api/jdk/javadoc/doclet/package-summary.html).
The new classes are not available in JDK 8, so a multirelease jar would likely be needed.

PR #4341, opened 2019-12-09
Build Jersey on JDK13

jbescos:
After JDK12 (included) there is a new way to generate javadoc.

The module jersey-wadl-doclet is working with that javadoc classes, so when you try to compile it from JDK12+ it fails.

I have created a different source folder that generates the same XML output than the previous. A multirelease jar file is created when it is compiled twice with JDK8_11 and JDK12+ or viceversa.

There is a second module (extended-wadl-webapp) that contains a plugin 'maven-javadoc-plugin'. This module will make use of this new classes and the test ExtendedWadlWebappTest will verify the output is correct.


ext/wadl-doclet/src/main/java8_11/org/glassfish/jersey/wadl/doclet/ResourceDoclet.java
```diff
@@ -96,6 +96,7 @@
      * @return true if no exception is thrown.
      */
     public static boolean start(final RootDoc root) {
+        boolean success = true;
```
jansupol: Please update copyright year

ext/wadl-doclet/pom.xml
```diff
@@ -32,6 +32,14 @@
         of resource classes, so that this can be used for extending generated wadl with useful
```
senivam: please update copyright year

jansupol:
Fixes #4325

Issue #4293, opened 2019-10-15
Remove XML and image related body providers from core-common module


tomas-langer:
Currently the class `org.glassfish.jersey.message.internal.MessagingBinders$MessageBodyProviders` contains the following providers (among others):
```
this.bindSingletonWorker(RenderedImageProvider.class);
((ClassBinding) this.bind(SourceProvider.StreamSourceReader.class).to(MessageBodyReader.class)).in(Singleton.class);
((ClassBinding) this.bind(SourceProvider.SaxSourceReader.class).to(MessageBodyReader.class)).in(Singleton.class);
((ClassBinding) this.bind(SourceProvider.DomSourceReader.class).to(MessageBodyReader.class)).in(Singleton.class);
((ClassBinding) this.bind(SourceProvider.SourceWriter.class).to(MessageBodyWriter.class)).in(Singleton.class);
```
The `RenderedImageProvider` depends on image rendering (and hence requires Linux with X installed.
The other providers are XML related (and drag the whole XML dependencies with them).

For normal JDK, this is not a major issue, as you have XML as part of the JDK and image processing would fail in runtime.
Already when using JPMS, we have a few required modules that may not be desired by an application.
When switching to ahead of time compilation using GraalVM (`native-image`), all of these dependencies MUST be part of the image - and as a result such image is quite big, and cannot be built in a docker container unless X is installed.

Please separate the registration of these providers to a different module, that can be excluded by the users.

jansupol:
relates to #4245

Issue #4245, opened 2019-09-07
Java 11 java.desktop module dependency


dconnelly:
Not sure why Jersey should have any dependency on AWT classes, but I am seeing this warning when running a Dropwizard/Jersey app under Java 11:
```
WARNING: HK2 service reification failed for [org.glassfish.jersey.message.internal.RenderedImageProvider] with an exception:
MultiException stack 1 of 2
java.lang.NoClassDefFoundError: java/awt/image/RenderedImage
```
We probably should move this provider to a separate maven module, and document that for Java 11+ the 'java.desktop' module would be required for usage.

arjantijms:
Maybe it's because of a property converter being used somewhere (this is often used for low level conversions, see e.g. https://arjan-tijms.omnifaces.org/2014/08/high-time-to-standardize-java-ee.html)

That one has an awkward historical dependency on AWT indeed. Many server apps use this one.

jansupol:
relates to #4293

PR #4342, opened 2019-12-10
Allow to disable certain default providers

jansupol:

### DIFF
--- a/core-common/src/main/java/org/glassfish/jersey/CommonProperties.java
+++ b/core-common/src/main/java/org/glassfish/jersey/CommonProperties.java
@@ -59,6 +59,16 @@ public final class CommonProperties {
     }
 
     /**
+     * Property which allows (if true) default System properties configuration provider.
+     *
+     * Effective if there are no any external properties providers
+     *
+     * Shall be set (if used) in system properties.
+     * @since 2.29
+     */
+    public static final String ALLOW_SYSTEM_PROPERTIES_PROVIDER = "jersey.config.allowSystemPropertiesProvider";
+
+    /**
      * If {@code true} then disable feature auto discovery globally on client/server.
      * <p>
      * By default auto discovery is automatically enabled. The value of this property may be overridden by the client/server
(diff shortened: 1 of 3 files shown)


jansupol:
Fixes #4293 and #4245

core-common/src/main/java/org/glassfish/jersey/message/internal/MessagingBinders.java
```diff
@@ -137,4 +150,110 @@ protected void configure() {
...
+            SOURCE ("javax.xml.transform.Source"),
+            STREAMSOURCE ("javax.xml.transform.stream.StreamSource");
+            Provider(String className) {
+                this.className = className;
+            }
+            private String className;
+        }
+
+        private static final String ALL = "ALL";
+        private HashSet<Provider> disabledProviders = new HashSet<>();
```
senivam: I would implement this logic vice-versa / instead of adding disabled providers to a set I would create the set of all providers and then remove disabled. So, you will have set of enabled providers only. For ALL this would be empty set. In such a case you would eliminate all IF CONTAINS checks and just loop over already prepared set thus you know all elements of that set are enabled.

jansupol: I did not like the logic that removes the disabled providers when the set of enabled would be created. If only one provider would be disabled (such as in the case of JDK11.desktop), the `if contains()` would still need to be there, checking whether the single provider is not disabled.

jansupol: I got your point. I have changed that.

core-common/src/main/java/org/glassfish/jersey/CommonProperties.java
```diff
@@ -58,6 +58,16 @@
                 "jersey.config.disableMoxyJson.server");
     }
 
+    /**
+     * Property which allows (if true) default System properties configuration provider.
+     *
+     * Effective if there are no any external properties providers
```
tomas-langer: "if there are no external properties providers" - the any is redundant and should not be there

jansupol: Nice spot

Issue #4256, opened 2019-09-11
HK2 AbstractBinders are configured twice


jeloba:
Since version 2.29 HK2 `AbstractBinders` are configured twice when registered directly in the application configuration. The following test demonstrates the issue:

```java
public class HK2BinderTest extends JerseyTest
{
    @Override
    protected Application configure()
    {
        final ResourceConfig config = new ResourceConfig();

        config.register(new AbstractBinder()
        {
            private int counter = 0;

            @Override
            protected void configure()
            {
                assertThat(counter++, is(0));
            }
        });

        return config;
    }

    @Test
    public void test()
    {
    }
}
```

Expected behavior would be only one call to `AbstractBinder.configure()`.

Issue #4344, opened 2019-12-11
Jersey 2.29 AbstractBinder.configure() called twice


mr-aixigo:
After upgrading von 2.28 to 2.29 I noticed that the `configure()` methods of all binders are called twice. This happens with `org.glassfish.hk2.utilities.binding.AbstractBinder` and can be solved by switching to `org.glassfish.jersey.internal.inject.AbstractBinder`.

Since this occured only in 2.29 I would like to know:
* Is this a bug which will be fixed?
* Should `org.glassfish.hk2.utilities.binding.AbstractBinder` be considered as deprecated? I could not find anything in the documentation that suggests I should stop using this class.

```
import org.glassfish.hk2.utilities.binding.AbstractBinder;       // bad
//import org.glassfish.jersey.internal.inject.AbstractBinder;    // good
import org.glassfish.jersey.server.ResourceConfig;

public class DemoConfig extends ResourceConfig
{
   public DemoConfig()
   {
      registerInstances( new DemoBinder() );
   }

   static class DemoBinder extends AbstractBinder
   {
      @Override
      protected void configure()
      {
         System.out.println( "configure()" );
      }
   }
}
```

jansupol:
Thanks, this will be fixed.
Duplicates #4256

jansupol:
As a workaround, you can use:
```
   public DemoConfig()
   {
      register(DemoBinder.class);
   }
```

PR #4349, opened 2019-12-16
Prevent HK2 AbstractBinder from being configured twice.

jansupol:
Revert the change that introduced BinderConfigurationFactory SPI.
Added possibility to register HK2 AbstractBinder class along with the instance option.


jansupol:
Fixes #4256, #4344. 

Revert #4122.

jansupol:
Fixes #4111.

PR #4351, opened 2019-12-18
Add initial implementation of JUnit Jupiter Extension (#3662)

### LINKED ISSUE #3662: JerseyTest is not compatible with JUnit 5
When using JerseyTest with JUnit 5 a null pointer exception gets thrown:

```
java.lang.NullPointerException
	at org.glassfish.jersey.test.JerseyTest.target(JerseyTest.java:564)
	at org.glassfish.jersey.test.JerseyTest.target(JerseyTest.java:578)
```

The issue seems to be that the setup() and tearDown() functions in the JerseyTest class are not called in JUnit 5 as they would be in JUnit 4. JUnit 5 uses the convention of @Before and @After annotations.

Here is my workaround, in the test class that extends JerseyTest add the following two methods:

```java
// do not name this setup()
@BeforeAll
public void before() throws Exception {
    super.setUp();
}

// do not name this tearDown()
@AfterAll
public void after() throws Exception {
    super.tearDown();
}

### PR DESCRIPTION
**Pull request by lordjaxom**
Fix #3662 

We are using this initial implementation in our internal projects successfully with JUnit Jupiter 5.2+. It provides a JUnit Jupiter Extension (and annotation) that sets up and configures the Jersey Test Framework and is able to inject Client, WebTarget and URI instances into fields and/or test method parameters.

ECA is signed.

Signed-off-by: Sascha Volkenandt <sascha@akv-soft.de>

### DIFF
--- a/test-framework/jupiter-extension/pom.xml
+++ b/test-framework/jupiter-extension/pom.xml
@@ -1,0 +1,61 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright (c) 2015, 2018 Oracle and/or its affiliates. All rights reserved.
+
+    This program and the accompanying materials are made available under the
+    terms of the Eclipse Public License v. 2.0, which is available at
+    http://www.eclipse.org/legal/epl-2.0.
+
+    This Source Code may also be made available under the following Secondary
+    Licenses when the conditions for such availability set forth in the
+    Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
+    version 2 with the GNU Classpath Exception, which is available at
+    https://www.gnu.org/software/classpath/license.html.
+
+    SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+
+-->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Review by Boosted — APPROVED** [review]

**Comment by DNAlchemist** [comment]
Is there any progress with this PR?

**Comment by sparsick** [comment]
This PR would help a lot of writing Jersey tests with Junit5. Maybe @jansupol or @senivam can have a look?

**Comment by jbescos** [comment]
I created another PR. I was not aware of this. Here is my PR:
https://github.com/eclipse-ee4j/jersey/pull/4824

I found the only issue is that there are 2 methods annotated with @After and @Before in JerseyTest. Those methods are not invoked in JUnit5 because it has other annotations for this (@AfterEach and @BeforeEach)

In my opinion we only need to add those new annotations together with the Junit4 ones. This does not throw runtime errors when JUnit5 is not in the classpath.

**Comment by lordjaxom** [comment]
@jbescos I think the two PRs are mostly unrelated. Enabling JerseyTest to work with JUnit 5 does not provide a JUnit-Extension, which is IMO preferable to having a test class extend JerseyTest. But the extension only provides an additional (more JUnit-y) approach to launching the Jersey Testframework.

File: test-framework/jupiter-extension/pom.xml [hunk 1]
```diff
@@ -0,0 +1,61 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter</artifactId>
+            <version>5.5.2</version>
+        </dependency>
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <version>3.14.0</version>
```
jbescos, line 57: It is preferable to create a property in pom.xml and reference that property here with ${assert.version}

PR #4359, opened 2020-01-07
Fix check style

jbescos:
[ERROR] src/main/java8_11/org/glassfish/jersey/wadl/doclet/ResourceDoclet.java[391] (regexp) RegexpSinglelineJava: File contains trailing whitespace.


senivam:
I would advise to update copyright year to 2020 in the modified file... even though only a white space is being removed....

PR #4364, opened 2020-01-12
Updated checkstyle plugin to latest 3.1.0

jansupol:
+ puppycrawl 8.28 + corresponding necessary changes in checkstyle XML files

### DIFF
--- a/etc/config/checkstyle-verify.xml
+++ b/etc/config/checkstyle-verify.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
-    Copyright (c) 2015, 2018 Oracle and/or its affiliates. All rights reserved.
+    Copyright (c) 2015, 2020 Oracle and/or its affiliates. All rights reserved.
 
     This program and the accompanying materials are made available under the
     terms of the Eclipse Public License v. 2.0, which is available at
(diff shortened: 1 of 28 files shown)


etc/config/checkstyle-verify.xml
```diff
@@ -60,13 +63,16 @@
             <property name="allowSingleLineStatement" value="true"/>
         </module>
         <module name="LeftCurly">
-            <property name="maxLineLength" value="130"/>
+            <property name="tokens" value="ANNOTATION_DEF, CLASS_DEF, CTOR_DEF, ENUM_CONSTANT_DEF, ENUM_DEF,
+                                           INTERFACE_DEF, LITERAL_CASE, LITERAL_CATCH, LITERAL_DEFAULT,
+                                           LITERAL_DO, LITERAL_ELSE, LITERAL_FINALLY, LITERAL_FOR, LITERAL_IF,
+                                           LITERAL_SWITCH, LITERAL_SYNCHRONIZED, LITERAL_TRY, LITERAL_WHILE,
+                                           METHOD_DEF, OBJBLOCK, STATIC_INIT" />
```
Verdent: This is default, if you were up to removing max line length, then just remove that property.

jansupol: LAMBDA is removed, added a comment

Issue #3306, opened 2016-01-04
Please reopen the task ...


jerseyrobot:
Please reopen the task. I have added a reproducing code.
There seems to be a HTTPS problem ... may be not only in the LoggingFilter!

ToYonos:
Same issue here with https, version 2.27.
It's even cut below 8kb

cen1:
Same experience in 2.28. Should be easy to reproduce, just create a client and execute a few requests which return a large JSON payload (like 30K), JSON is cut off and `...more...` is missing. Changing `LOGGING_FEATURE_MAX_ENTITY_SIZE` does not seem to do anything. I noticed it with HTTP so it's not really exclusive to HTTPS.

maximthomas:
Seems there's an issue with this function, in line 
```java
 final int entitySize = stream.read(entity); 
```
`entitySize` variable value is lower than actual entity size

https://github.com/eclipse-ee4j/jersey/blob/da3f1f9f36cffa1ae51ac0f66254ea630739b1ef/core-common/src/main/java/org/glassfish/jersey/logging/LoggingInterceptor.java#L190-L204

PR #4367, opened 2020-01-16
fix large body entity truncation

maximthomas:
While using BufferedInputStream, need multiple reads (issue #3306)


jansupol:
Please see the [wiki](https://github.com/eclipse-ee4j/jersey/wiki/Missing-ECA,-Missing-Sign-off-by)

PR #4368, opened 2020-01-16
Allow to register the resource by instance or proxy

dhenneke:
This PR resubmits the abandoned PR #3820 with a signed-off commit and corrected license headers as proposed in https://github.com/eclipse-ee4j/jersey/pull/3820#issuecomment-555462746.

We would be very happy if this could be merged in this state.

Closes #3820

### DIFF
--- a/core-server/src/main/java/org/glassfish/jersey/server/model/IntrospectionModeller.java
+++ b/core-server/src/main/java/org/glassfish/jersey/server/model/IntrospectionModeller.java
@@ -61,6 +61,8 @@ final class IntrospectionModeller {
 
     // introspected annotated JAX-RS resource class
     private final Class<?> handlerClass;
+    // introspected annotated JAX-RS resource instance
+    private final Object handlerInstance;
     // validation flag
     private final boolean disableValidation;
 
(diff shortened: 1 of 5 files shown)


dhenneke:
It looks like the ECA check is still not satisfied...

ryandens:
@dhenneke I'm not associated with the eclipse foundation but am definitely very excited about this change. The reason the ECA check is failing, to my understanding, is either because you haven't signed the Eclipse Contributor Agreement or the fact that you signed it isn't associated with your GitHub account. Here is a snippet from this repository's CONTRIBUTING.md

[Contributing.md ECA](https://github.com/eclipse-ee4j/jersey/blob/master/CONTRIBUTING.md#eclipse-contributor-agreement)

> Commits that are provided by non-committers must have a Signed-off-by field in the footer indicating that the author is aware of the terms by which the contribution has been provided to the project. The non-committer must additionally have an Eclipse Foundation account and must have a signed Eclipse Contributor Agreement (ECA) on file.

dhenneke:
It looks like the ECA still needs to be signed by the original author of the commit. When I understand it correctly, it should work if I become the author and make the original contributor an `Also-by`. But this feels a bit hacky and should be approved by someone from the maintainers before trying.

PR #4373, opened 2020-01-27
Fixed stacktraces caused by incorrect JNDI lookup

dmatej:
* Fixed processing default local interface
  - cause:
    - the stateless bean implemented some interface from library
    - the interface was annotated only with `@Path`
    - neither interface nor class was annotated by `@Local` or `@Remote` (then `@Local` is implicit in ejb3.2)

* Implemented support of `Stateless.name` attribute
  - it is used in JNDI names on Payara

* Limited default local interfaces
  - partial impl of the ejb 3.2 spec 4.9.7, some interfaces are not allowed

Signed-off-by: David Matejcek <dmatej@seznam.cz>

### DIFF
--- a/containers/glassfish/jersey-gf-ejb/src/main/java/org/glassfish/jersey/gf/ejb/internal/EjbComponentProvider.java
+++ b/containers/glassfish/jersey-gf-ejb/src/main/java/org/glassfish/jersey/gf/ejb/internal/EjbComponentProvider.java
@@ -16,6 +16,8 @@
  */
 package org.glassfish.jersey.gf.ejb.internal;
 
+import java.io.Externalizable;
+import java.io.Serializable;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.InvocationTargetException;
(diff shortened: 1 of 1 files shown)


dmatej:
Is the failing test stable?

jansupol:
This test is intermittently failing, sorry for that, for all JDKs. Should be fixed by #4387.

containers/glassfish/jersey-gf-ejb/src/main/java/org/glassfish/jersey/gf/ejb/internal/EjbComponentProvider.java
```diff
@@ -92,23 +95,43 @@
...
-                return (T) lookup(ctx, clazz, clazz.getSimpleName(), ejbProvider);
+                return (T) lookup(ctx, clazz, beanName, ejbProvider);
             } catch (NamingException ex) {
                 Logger.getLogger(ApplicationHandler.class.getName()).log(Level.SEVERE, null, ex);
                 return null;
             }
         }
 
+        private static <T> String getBeanName(final Class<T> clazz) {
+            final Stateless stateless = clazz.getAnnotation(Stateless.class);
```
jansupol: Why don't you deal with `Stateful`, too?

dmatej: It is not requested by specification. Isn't REST/Jersey stateless?

jansupol: You are right, JAX-RS Spec discusses Stateless and Singleton, only.

Issue #4108, opened 2019-04-24
Create META-INF/services for JAX-RS RuntimeDelegate and ClientBuilder


jansupol:
[JAX-RS is about to remove default names of services](https://github.com/eclipse-ee4j/jaxrs-api/pull/755), which would break Jersey. Jersey should add META-INF/services to point to services in Jersey. For RuntimeDelegate, this is already done in JAXRS_2_2 branch.

cowwoc:
Version 3.0.0

When I attempt to use `UriBuilder` I get:

```
java.lang.RuntimeException: java.lang.ClassNotFoundException: Provider for jakarta.ws.rs.ext.RuntimeDelegate cannot be found
	at jakarta.ws.rs/jakarta.ws.rs.ext.RuntimeDelegate.findDelegate(RuntimeDelegate.java:113)
	at jakarta.ws.rs/jakarta.ws.rs.ext.RuntimeDelegate.getInstance(RuntimeDelegate.java:84)
	at jakarta.ws.rs/jakarta.ws.rs.core.UriBuilder.newInstance(UriBuilder.java:69)
```

I attempted to add:

```
<dependency>
	<groupId>org.glassfish.jersey.core</groupId>
	<artifactId>jersey-common</artifactId>
        <version>3.0.0</version>
</dependency>
```

This issue is tagged with version 2.30 but it doesn't look like it was ever fixed in version 3.0.0 (which was just released last week). Also note that I am using Java Modules so that might further complicate the situation. How should I proceed?

jansupol:
We never updated 2.x with this, but you can check 3.x jersey-common.jar; it contains META-INF/services/jakarta.ws.rs.ext.RuntimeDelegate. It contains a 0x0A at the end, though...

If that causes an issue, the workaround would be to add a dependency to core-server, which has the proper runtime delegate and does not have 0x0A at the end of the META-INF/services file.

cowwoc:
@jansupol Thanks for the feedback.

As I mentioned above jersey-common 3.0.0 does not work so this needs to get fixed.

PR #4382, opened 2020-02-01
[#4108] Create META-INF/services for JAX-RS ClientBuilder

mcserra:
Signed-off-by: Miguel Serra <Miguel.Serra@criticaltechworks.com>


jansupol:
Why did you do this? For the current Master, it is not needed. This would be needed for JAX-RS 3.0.0 and this is handled in JAXRS_2_2 branch and it is cherry-picked to 3.x branch.

Issue #4046, opened 2019-02-01
netty container disregards path component of baseUri


Chris-Lercher:
No matter, which path component is specified in the baseUri when creating the server via org.glassfish.jersey.netty.httpserver.NettyHttpContainerProvider#createHttp2Server (probably the same for `createServer`?), all the resources are located relative to "/".

e.g. use a `@Path("example")` resource together with `createHttp2Server(URI.create("http://localhost:13333/abc/"), rc, ...)`

The client sees the resource wrongly located at "http://localhost:13333/example".

Using the debugger, I found the problem to be in org.glassfish.jersey.netty.httpserver.JerseyHttp2ServerHandler#createContainerRequest . The issue is, that it creates the `requestUri` from the baseUri + the path from the HTTP header, so it believes, the HTTP request was for "http://localhost:13333/abc/example".

Note: Please compare this with `GrizzlyHttpServerFactory.createHttpServer(URI.create("http://localhost:13333/abc/", rc, ...)`, which behaves correctly. (There, the client sees the resource at "http://localhost:13333/abc/example")

tqh:
I think it actually doesn't disregard it, but the logic in JerseyServerHandler.java and JerseyHttp2ServerHandler is not correct.
See https://github.com/eclipse-ee4j/jersey/blob/8c60fd2734c333cbca4dc9e0ab7d81306494532a/containers/netty-http/src/main/java/org/glassfish/jersey/netty/httpserver/JerseyServerHandler.java#L122

To me it seems like there are two bugs in that line:
 * baseURI contains a path, but the incoming request path (s) is a full path so now you have the basepath twice.
 * The request url is also encoding chars with a function to encode query part only. Not sure if this is safe or not, but if it is safe it probably should have a clarification. URIBuilder would seem like a better option here.

PR #4385, opened 2020-02-03
[#4046] netty container disregards path component of baseUri

mcserra:
Signed-off-by: Miguel Serra <Miguel.Serra@criticaltechworks.com>


jansupol:
Can you please rebase atop the master and resolve conflicts? Thanks.

mcserra:
Done 👍

containers/netty-http/src/main/java/org/glassfish/jersey/netty/httpserver/JerseyServerHandler.java
```diff
@@ -173,8 +181,28 @@ public void removeProperty(String name) {
                         properties.remove(name);
                     }
                 }, resourceConfig);
+    }
+
+    private String getPath(HttpRequest req) {
+        String s = req.uri();
+        if (baseUriHasBasePath()) {
+            s = s.startsWith("/") ? s.substring(1) : s;
+            if (!baseUri.toString().endsWith("/")) {
```
jansupol: It is unclear what this condition does, but it is not needed for either of your tests

containers/netty-http/src/main/java/org/glassfish/jersey/netty/httpserver/JerseyServerHandler.java
```diff
@@ -173,8 +181,28 @@ public void removeProperty(String name) {
...
+                s = s.replaceFirst("/", "");
+            }
+            s = "/" + s;
+        }
+        return s;
+    }
+
+    private boolean baseUriHasBasePath() {
+        String path = baseUri.getPath();
+        return path != null && !path.isEmpty() && !"/".equals(path);
```
jansupol: `!path.isEmpty` is not needed

senivam: this can be simplified to `return !"/".equals(path);`

jansupol:
So far it looks like the change in `JerseyServerHandler`
```
        String s = getPath(req);
        URI requestUri = URI.create(getRequestUri() + ContainerUtils.encodeUnsafeCharacters(s));
```
can be simplified to 
```
        String s = !req.uri().startsWith("/") ? "/" + req.uri() : req.uri();
        URI requestUri = URI.create(getRequestUri() + ContainerUtils.encodeUnsafeCharacters(s));
```
Can you more elaborate on `getPath` method?

mcserra:
Added the missing test for the condition `if (!baseUri.toString().endsWith("/"))`.

containers/netty-http/src/main/java/org/glassfish/jersey/netty/httpserver/JerseyServerHandler.java
senivam: for case when base is: http://localhost:9994/base
and resource is: /base/resource/
the request URI will be: http://localhost:9994/baseresource/

I suppose that is not what's expected, however tests are passing OK. 
I suppose that's not which is expected

mcserra: Yes, that's the case. It's a bit weird,  but the tests now reflect that.

senivam: so, you are saying, that having "http://localhost:9994" + "/baseresource/" is expected behavior?

jansupol: This is mixing the baseUri and path segments into a single segment, which is not what has been intended. It would bring confusion to the users.

mcserra: Ok, that makes sense. I was thinking it would be better not to change the code-server, so I made the changes there. 
After some digging I rediscovered the problem root. 
When the baseUri doesn't end with a delimiter, a call to the ContainerRequest#encodedRelativePath will retrieve a relative path starting with a delimiter, which will then result in a double starting delimiter when called from MatchResultInitializerRouter#apply. That will result in a non matching path when resolving in  RoutingStage.
I tried two solutions:
The first is adding the delimiter, on the baseUri, when creating the ContainerRequest (in JerseyServerHandler#createContainerRequest). The other is adding it on ContainerRequest#encodedRelativePath.
I think those are the simplest options, and both pass the tests.
If you guys can give me your opinion, i can make another commit with the changes. 
Thanks

Issue #4286, opened 2019-10-10
Netty Connector does not reuse and leaks connections


olotenko:
https://github.com/eclipse-ee4j/jersey/blob/8dcbaf4/connectors/netty-connector/src/main/java/org/glassfish/jersey/netty/connector/NettyConnector.java#L192 - for every new request a new connection is established. This is suboptimal, since generally one would want to reuse a JAX-RS Client, and benefit from `keep-alive`.

Also, there appears to be no code to close unused connections.

olotenko:
[netty_conn_reuse.zip](https://github.com/eclipse-ee4j/jersey/files/3713414/netty_conn_reuse.zip)
POC reusing connections. It is really a proof of concept demonstrating socket reuse, does not take care of unused connections, does not limit the connection pool size.

PR #4387, opened 2020-02-05
netty connector/container modifications

senivam:
POC adaptation for netty-connector (from #4286) + fix for netty-container (which depends on NettyInputStream). 

### DIFF
--- a/connectors/netty-connector/src/main/java/org/glassfish/jersey/netty/connector/internal/NettyInputStream.java
+++ b/connectors/netty-connector/src/main/java/org/glassfish/jersey/netty/connector/internal/NettyInputStream.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016, 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2016, 2020 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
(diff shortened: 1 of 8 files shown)


senivam:
CQ #:[21591](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=21591)

jansupol:
Please rebase on top of #4393

connectors/netty-connector/src/main/java/org/glassfish/jersey/netty/connector/internal/NettyInputStream.java
```diff
@@ -16,107 +16,141 @@
...
+          current.release();
+       }
+
+       current = null;
+       buffer = null;
+       cleanup(true);
     }
 
-    @Override
-    public int available() throws IOException {
```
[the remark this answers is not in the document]
senivam: modified

connectors/netty-connector/src/main/java/org/glassfish/jersey/netty/connector/internal/NettyInputStream.java
```diff
@@ -16,107 +16,150 @@
...
     }
 
     @Override
     public int available() throws IOException {
-        ByteBuf peek = isList.peek();
-        if (peek != null && peek.isReadable()) {
-            return peek.readableBytes();
-        }
-        return 0;
+        return buffer == null ? 0 : buffer.remaining();
```
jansupol: should be synchronized

senivam: ok

olotenko: Totally disagree. `NettyInputStream` has been, and remains, not thread safe. `read` and `close` methods have been, and are not `synchronized`, therefore it has always been assumed to be used by a single thread, or access is synchronized by out of band means.

jansupol: I see your point. We need to synchronize `isList` and `reading`. So only `awaitNext()`, `publish()`, and `cleanup()` are needed to be synchronized. We can remove it from `available`, `close`, `releaseByteBuf` and `complete`. Or do I miss something?

olotenko: No, just do not add to `available`.

`publish`, `awaitNext` and `close` interact with each other. A simple way to achieve total order of state transitions between things that are mostly done by two threads, and very rarely, is to use `synchronized`.

So we have one thread (Netty event loop) that receives byte buffers from the socket, and hands these buffers off to a different thread (Jersey thread that wants to `readEntity`). This is done through `publish`, and is potentially observed by `awaitNext`. `publish` also needs to see if maybe `close` has been called, so it does not assume anyone is going to release the buffers in the future. That's the purpose of `synchronized` around those three.

`synchronized` on `cleanup` is not harmful, but not necessary, because it is invoked from `synchronized` sections. `synchronized` on `available` I think is low cost, but because it does not achieve any improved correctness, I would recommend not adding it.

jansupol: Sounds like we agree here.
- Either we have `publish`, `awaitNext`, `complete` and `close` synchronized
- Or we have `publish`, `awaitNext`, and we move the synchronized from `complete` and `close` down to `cleanup`, where `end=true` and mainly `if (reading)` is used. 
Both are fine, though having `close()` synchronized could be a bit misleading as the NettyInputStream is not threadsafe.

senivam: OK, as per discussion, I've implemented the B) way - compete and close (including available and releaseByteBuf) are not synchronized, cleanup became synchronized.

olotenko: LGTM

PR #4389, opened 2020-02-07
Transition Jakarta REST API from javax to jakarta package

jansupol:


jansupol:
Closing in favor of #4392 which does not have issues with sign-off-by.

PR #4390, opened 2020-02-07
Fix #3433 - Multiple cookies with same name are not supported

### LINKED ISSUE #3433: Multiple cookies with same name are not supported
It's possible to have multiple cookies with the same name in the cookie header (e.g. cookies in subdomains/subpaths). So, the following is a valid header:
`Cookie:"token=val1; token=val2"`.

However, `org.glassfish.jersey.server.ContainerRequest.getCookies` returns a `Map<String, Cookie>`, indicating that it returns only one cookie per name. It calls `HttpRequestHeader.readCookies` which in turn calls `CookieParser.parseCookies`. The implementation of `parseCookies` makes it clear that for multiple cookies with the same name, the last one is returned in the map.
#### Environment
Server
#### Affected Versions
[2.22]

tvallin:

NewCookie :
     - pick the latest expiration date
     - if same date, take the longest path string

Cookie : 
     - pick the longest string

### DIFF
--- a/core-common/src/main/java/org/glassfish/jersey/message/internal/HeaderUtils.java
+++ b/core-common/src/main/java/org/glassfish/jersey/message/internal/HeaderUtils.java
@@ -301,23 +301,6 @@ public final class HeaderUtils {
     }
 
     /**
-     * Compare two objects according to comparator
-     *
-     * @param first first object to be compared
-     * @param second second object to be compared
-     * @param comparator criteria
-     * @return the preferred object according to comparator
-     */
-    public static <T> T compareNullable(T first, T second, Comparator<T> comparator) {
-        if (first == null) {
-            return second;
-        } else if (second == null) {
-            return first;
-        }
-        return comparator.compare(first, second) > 0 ? first : second;
-    }
-
-    /**
      * Compare two NewCookies having the same name. See documentation RFC.
      *
      * @param first    NewCookie to be compared.
@@ -336,9 +319,9 @@ public final class HeaderUtils {
         }
 
         if (first.getMaxAge() != second.getMaxAge()){
-            return compareNullable(first, second, Comparator.comparing(NewCookie::getMaxAge));
+            return Comparator.comparing(NewCookie::getMaxAge).compare(first, second) > 0 ? first : second;
         } else if (first.getExpiry() != null && second.getExpiry() != null && !first.getExpiry().equals(second.getExpiry())) {
-            return compareNullable(first, second, Comparator.comparing(NewCookie::getExpiry));
+            return Comparator.comparing(NewCookie::getExpiry).compare(first, second) > 0 ? first : second;
         } else {
             return first.getPath().length() > second.getPath().length() ? first : second;
         }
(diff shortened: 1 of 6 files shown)


tests/e2e-core-common/src/test/java/org/glassfish/jersey/tests/e2e/common/message/internal/HeaderUtilsTest.java
```diff
@@ -180,4 +184,97 @@ public void testAsHeaderString() throws Exception {
...
     }
+
+    @Test
+    public void testgetPreferedNewCookie(){
+
+        Date earliestDate;
+        Date latestDate;
+
+        try {
+            earliestDate = HttpDateFormat.readDate("Fri, 1 Jan 2020 00:00:00 GMT");
```
jansupol: Please use Calendar or LocalDateTime to set the dates you want, to prevent parsing exception with foreign locales

core-common/src/main/java/org/glassfish/jersey/message/internal/HeaderUtils.java
```diff
@@ -298,6 +299,24 @@ public static void checkHeaderChanges(final Map<String, String> headersSnapshot,
...
+     *
+     * @param one    NewCookie to be compared.
+     * @param second NewCookie to be compared.
+     * @return the prefered NewCookie according to rules :
+     *              - the latest expiration date.
+     *              - if same expiration date, the longest path.
+     */
+    public static NewCookie getPreferedNewCookie(NewCookie one, NewCookie second) {
+
+        if (!one.getExpiry().equals(second.getExpiry())){
```
senivam: expiry date can be NULL. Isn't it better to use maxAge value?

core-common/src/main/java/org/glassfish/jersey/message/internal/HeaderUtils.java
```diff
@@ -298,6 +299,24 @@ public static void checkHeaderChanges(final Map<String, String> headersSnapshot,
...
+    /**
+     * Compare two NewCookies having the same name. See documentation RFC.
+     *
+     * @param one    NewCookie to be compared.
+     * @param second NewCookie to be compared.
+     * @return the prefered NewCookie according to rules :
+     *              - the latest expiration date.
+     *              - if same expiration date, the longest path.
+     */
+    public static NewCookie getPreferedNewCookie(NewCookie one, NewCookie second) {
```
senivam: Preferred ? may be rename the whole method to getPreferredCookie?

core-common/src/main/java/org/glassfish/jersey/message/internal/HeaderUtils.java
jansupol: There is a difference between expiry and maxAge. 
IE does not support maxAge, so most of the servers set expiry rather than maxAge. It is possible to set one or the other.
NewCookie is what comes from the server, and even though maxAge is recommended, expires seems to be more often. But it seems to be one or the other, so we should compare the later from both.

But yes, we should check for NULL, too.

senivam: Well ,regarding check of expiry VS max-age it's good to check them both :) 

Regarding setting one or another attribute in request - actually server shall put both. For now I've found that the only browser which does not support max-age attribute is IE. 

Per RFC 6265 max-age attribute has precedence over expiry attribute if both are present in request. In my opinion this shall be respected while choosing preferred cookie.

Issue #4369, opened 2020-01-16
NettyConnectorProvider (jersey-netty-connector) doesn't send query parameters in the Get Request


luanjr:
I am trying to send an HTTP request with query parameters using Jersey Client and setting NettyConnectorProvider (org.glassfish.jersey.connectors:jersey-netty-connector:2.25.1) as a provider to the config (org.glassfish.jersey.client.ClientConfig) by the method connectorProvider().
I tried to send the query parameters using the method queryparam(), but the values are not sent on the request.

Looking in the source code of NettyConectorProvider, it uses the method requestUri.getRawPath()) during the creation of DefaultFullHttpRequest, but this method returns the path without the query parameters.

nettyRequest = new DefaultFullHttpRequest(HttpVersion.HTTP_1_1,
                                          HttpMethod.valueOf(jerseyRequest.getMethod()),
                                          requestUri.getRawPath());

How can I send query parameters in a GET request using NettyConnectorProvider or is there a knowledge issue related to query parameters?

senivam:
Did you try:
```
requestUri.toASCIIString();
```
or 
```
requestUri.getRawPath() + "?" + requestUri.getRawQuery()
```
instead of 
```
requestUri.getRawPath()
```
as DefaultFullHttpRequest constructor parameter?

jansupol:
Fixed by #4393

PR #4393, opened 2020-02-13
Query parameters were not included in netty URI

jbescos:
This is the issue:
https://github.com/eclipse-ee4j/jersey/issues/4369

I didn't find a better solution for this because the Netty HttpRequest doesn't have a method to add query parameters.


connectors/netty-connector/src/main/java/org/glassfish/jersey/netty/connector/NettyConnector.java
```diff
@@ -282,6 +283,14 @@ public void run() {
         return settableFuture;
     }
 
+    private String buildPathWithQueryParameters(URI requestUri) {
+        if (requestUri.getRawQuery() != null) {
+            return new StringBuilder(requestUri.getRawPath()).append("?").append(requestUri.getRawQuery()).toString();
```
senivam: Why not to use `return String.format("%s?%s", requestUri.getRawPath(), requestUri.getRawQuery());`?

And do not forget about copyright year please...

Issue #4388, opened 2020-02-07
Jerey 2.30 breaks HK2 AbstractBinder injection in Features


cykl:
Since I upgraded to Jersey 2.30, it is no longer possible to inject an instance defined by an HK2 AbstractBinder into a Feature. 

Here is a minimal test case:

```java
import javax.inject.Inject;
import javax.inject.Singleton;
import javax.ws.rs.core.Application;
import javax.ws.rs.core.Feature;
import javax.ws.rs.core.FeatureContext;

import org.glassfish.hk2.utilities.binding.AbstractBinder;
import org.glassfish.jersey.server.ResourceConfig;
import org.glassfish.jersey.test.JerseyTest;

public class MTC {

    public static void main( String[ ] args ) throws Exception {
        var test = new JerseyTest( ) {

            @Override
            protected Application configure( ) {
                final ResourceConfig config = new ResourceConfig( );

                var binder = new AbstractBinder( ) {
                    @Override
                    protected void configure( ) {
                        System.out.println( "Configuring binder" );
                        bindAsContract( Dependency.class ).in( Singleton.class );
                    }
                };

                config.register( binder );
                config.register( MyFeature.class );
                return config;
            }
        };

        test.setUp( );
    }

    static final class MyFeature implements Feature {

        private final Dependency service;

        @Inject
        public MyFeature( Dependency service ) {
            this.service = service;
        }

        @Override
        public boolean configure( FeatureContext context ) {
            return true;
        }
    }

    static final class Dependency {

        @Inject Dependency( ) {
            System.out.println( "Constructing Service" );
        }
    }
}
```

It used to work on Jersey 2.29. Switching to a Jersey's AbstractBinder also works. This is likely related to https://github.com/eclipse-ee4j/jersey/pull/4349.

cykl:
Root cause is likely https://github.com/eclipse-ee4j/jersey/commit/29913b90f63a29bfdd35bf465c553e07b0a6c99c#diff-546b2ffbf565a0242fdfb1dd4329380bR626

CommonConfig:

```java
    public void configureMetaProviders(InjectionManager injectionManager, ManagedObjectsFinalizer finalizer) {
        // First, configure existing binders
        Set<Binder> configuredBinders = configureBinders(injectionManager, Collections.emptySet());

        // Check whether meta providers have been initialized for a config this config has been loaded from.
        if (!disableMetaProviderConfiguration) {
            // Configure all features
[...]

PR #4394, opened 2020-02-18
Allow HK2 AbstractBinder class to bind before the Feature is called

jansupol:


jansupol:
Fixes #4388

PR #4408, opened 2020-03-11
Minor modification on documentation chapter

tvallin:

### DIFF
--- a/docs/src/main/docbook/custom-di.xml
+++ b/docs/src/main/docbook/custom-di.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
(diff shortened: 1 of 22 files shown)


docs/src/main/docbook/custom-di.xml
```diff
@@ -65,8 +65,8 @@
         as a good example that will help me demonstrate implementation of the use cases described above.
         The following examples should work on top of Jersey Servlet integration module. The approach that will be
         demonstrated could be further generalized.
-        Bellow i will show how to make actual Servlet &jee6.servlet.HttpSession; injectable into JAX-RS components
```
jansupol: Please use "we", not "I"

senivam: and below, not bellow

PR #4411, opened 2020-03-18
Add new interfaces (PreInvocationInterceptor, PostInvocationInterceptor, and InvocationBuilderListener) documentation

tvallin:
Author/Source: Jan Supol from his blog http://blog.supol.cz/?p=171

### DIFF
--- a/docs/src/main/docbook/client.xml
+++ b/docs/src/main/docbook/client.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 <!--
 
-    Copyright (c) 2010, 2018 Oracle and/or its affiliates. All rights reserved.
+    Copyright (c) 2010, 2020 Oracle and/or its affiliates. All rights reserved.
 
     This program and the accompanying materials are made available under the
     terms of the Eclipse Public License v. 2.0, which is available at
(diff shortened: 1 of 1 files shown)


docs/src/main/docbook/client.xml
```diff
@@ -999,4 +999,139 @@ Client client = ClientBuilder.newBuilder().sslContext(sslContext).build();</prog
 
 
     </section>
+
+    <section>
+        <title>New Features</title>
```
jansupol: Please do not say it's a new feature. The documentation is there for good, it will be an old feature in a few releases.

docs/src/main/docbook/client.xml
```diff
@@ -999,4 +999,139 @@ Client client = ClientBuilder.newBuilder().sslContext(sslContext).build();</prog
...
+        <para>
+            Jersey 2.30 comes with multiple new features, and this post describes three new interfaces on the client.
```
jansupol: "comes with multiple new features" is not good enough for the documentation. The documentation needs to mention the exact features. Maybe I would create 
- Section, InvocationInterceptors: 
   `<para> Suppose a case that the start of the request is to be logged and even measured....</para>`
- Section InvocationBuilderListener. 

These two sections would be on the same level as New Features. 

Drop `Jersey 2.30 comes... They are...` and start with `Suppose`.

PR #4413, opened 2020-03-23
Sync 3.x with master

jansupol:


jansupol:
There seems to be a bug in the ECA check. The validation says:
> Anonymous (serramiguel1@****.com) is not covered by the necessary legal agreements in order to proceed.

But there always is Signed-off-by: Miguel Serra <Miguel.Serra[at]xxx.com>

PR #4418, opened 2020-03-24
Remove Jackson1

jansupol:


jansupol:
Jackson 1 has CWEs and depends on javax based JAX-RS API.

Issue #4410, opened 2020-03-16
ClientResponse.readEntity fails if the server returns a blank Content-Type


ewanmellor:
o.g.jersey.client.ClientResponse.readEntity throws an exception if the server has responded with a Content-Type header with an empty string.  The exception is 'o.g.jersey.message.internal.HeaderValueException: Unable to parse "Content-Type" header value: ""' (full exception below).

There is code in InboundMessageContext.readEntity that is trying to handle an invalid mediaType by defaulting it to application/octet-stream.  (This would be the correct behavior in my opinion.)  However, that code does not work because:

1. InboundMessageContext.getMediaType is documented as returning a value or null.
2. InboundMessageContext.singleHeader is documented as returning a value, null, or throwing ProcessingException.
3. getMediaType calls singleHeader but doesn't catch exceptions.
4. readEntity doesn't catch exceptions from getMediaType either, so the fallback to application/octet-stream is not reached.  The exception bubbles up.

In my opinion, readEntity is violating its contract by throwing exceptions when they're not documented, so the fix belongs there.  However, this is a public method so that might have consequences that I'm not aware of.

```
org.glassfish.jersey.message.internal.HeaderValueException: Unable to parse "Content-Type" header value: ""
	at org.glassfish.jersey.message.internal.InboundMessageContext.exception(InboundMessageContext.java:341)
	at org.glassfish.jersey.message.internal.InboundMessageContext.singleHeader(InboundMessageContext.java:336)
	at org.glassfish.jersey.message.internal.InboundMessageContext.getMediaType(InboundMessageContext.java:449)
	at org.glassfish.jersey.message.internal.InboundMessageContext.readEntity(InboundMessageContext.java:872)
	at org.glassfish.jersey.message.internal.InboundMessageContext.readEntity(InboundMessageContext.java:811)
	at org.glassfish.jersey.client.ClientResponse.readEntity(ClientResponse.java:298)
	at org.glassfish.jersey.client.InboundJaxrsResponse$1.call(InboundJaxrsResponse.java:91)
	at org.glassfish.jersey.internal.Errors.process(Errors.java:292)
	at org.glassfish.jersey.internal.Errors.process(Errors.java:274)
	at org.glassfish.jersey.internal.Errors.process(Errors.java:205)
	at org.glassfish.jersey.process.internal.RequestScope.runInScope(RequestScope.java:365)
[...]

PR #4419, opened 2020-03-25
Allow to use HeaderDelegateProvider to parse the response MediaType

jansupol:


jansupol:
Relates to #4410

PR #4424, opened 2020-03-28
Additionally building on OpenJDK14

mkarg:
Since JDK 14 was released, it makes sense to test buildability on OpenJDK14.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ jdk:
   - oraclejdk8
   - openjdk11
   - openjdk13
+  - openjdk14
 
 cache:
   directories:


jansupol:
I would be surprised if the build would pass on JDK 14. We need to wait for eclipse groovy plugin that supports groovy 3 that supports JDK14. The information should appear on their [news page](https://github.com/groovy/groovy-eclipse/wiki). Given that they used to release every 3 months, I expect a release soon.

jansupol:
> Caused by: org.apache.maven.plugin.PluginExecutionException: Execution default-compile of goal org.apache.maven.plugins:maven-compiler-plugin:3.8.0:compile failed: A required class was missing while executing org.apache.maven.plugins:maven-compiler-plugin:3.8.0:compile: Could not initialize class org.codehaus.groovy.vmplugin.v7.Java7

mkarg:
That is correct, but why did you *close* (i. e. finally reject) the PR instead of simply put it aside while it is in a draft state?

jansupol:
Because the PR did not work

mkarg:
> Because the PR did not work

Well, that's why it was in *draft* state. ;-)

PR #4425, opened 2020-03-28
Updated GF, Jetty, Mimepull, Moxy, Yasson dependencies

jansupol:


jansupol:
We can piggyback Junit 5.6.0 CQ ~21545~

PR #4426, opened 2020-04-01
Support jdk15

jbescos:
Groovy updated their libs to support jdk14, and it also works with jdk15.


jansupol:
Filed [CQ 21871](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=21871)

PR #4430, opened 2020-04-02
Fix Extended Wadl Example OSGi Test for JDK11

jansupol:


examples/extended-wadl-webapp/pom.xml
```diff
@@ -235,24 +235,17 @@
...
-                            </excludes>
-                        </configuration>
-                    </plugin>
-                </plugins>
-            </build>
+            <dependencies>
+                <dependency>
+                    <groupId>com.sun.activation</groupId>
+                    <artifactId>jakarta.activation</artifactId>
+                    <version>${jakarta.activation.version}</version>
```
senivam, line 246: this shall be used from dependencyManagement, or is there a reason why not to use dependencyManagement?

PR #4431, opened 2020-04-02
Adopt ASM 8.0

jansupol:


jansupol:
Filed [CQ 21873](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=21873)

PR #4432, opened 2020-04-03
Upgraded bean validation to 6.1.2.Final

jansupol:


jansupol:
The CQs were approved

PR #4438, opened 2020-04-08
Fix jersey examples

tvallin:


examples/helloworld-cdi2-se/pom.xml
```diff
@@ -81,6 +81,20 @@
...
+        <profile>
+            <id>jdk11</id>
+            <activation>
+                <jdk>[11,)</jdk>
+            </activation>
+            <dependencies>
+                <dependency>
+                    <groupId>org.jboss.weld.se</groupId>
+                    <artifactId>weld-se-core</artifactId>
+                    <version>3.1.1.Final</version>
```
senivam, line 94: this shall be taken from dependenciesManagement, or is there some reason why it's not in dependenciesManagement?

tvallin: If the weld version in dependenciesManagement is updated to 3.1.1.Final, the build fails in other modules. 
Update the version into the example pom make it fails as well, which is why I have decided to do it this way where it does not produce any failing tests in Jersey and make helloworld-cdi2-se run with jdk11.

senivam: CQ #[21928](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=21928) created to allow related weld-se-core to be used in the project

jansupol: CQ is approved

PR #4441, opened 2020-04-09
Fix some build warnings and a multithread warning, when building with…

thomasmey:
… -T 1C

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -392,6 +392,7 @@
             <plugin>
                 <groupId>org.glassfish.copyright</groupId>
                 <artifactId>glassfish-copyright-maven-plugin</artifactId>
+                <version>2.0</version>
                 <inherited>true</inherited>
             </plugin>
             <plugin>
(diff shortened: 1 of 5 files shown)


jansupol:
Thank you for contribution. It looks good. Except, can you please update the copyright year in the modified files?

thomasmey:
hi, yes can do, but one question, what does it mean that there are two years, e.g.:
"Copyright (c) 2015, 2018..."

In what should I change it?
a.) Copyright (c) 2020
b.) Copyright (c) 2015, 2020
c.) Copyright (c) 2015, 2018, 2020

I think this copyright notice is only relevant for US law only?!

Also this seems to indicate that the notice is even not necessary any more at all:
https://en.wikipedia.org/wiki/Copyright_symbol
but OTOH IANAL!

jansupol:
We use two years, the year of creation and the year of the last change. So option B, please.

The [Eclipse Foundation](https://www.eclipse.org/legal/cpl2epl/cpl_copyrightandlicensenotice.php) says:
> {DATE} is either a year or a range of years with the first and last years of the range separated by a comma. So for example: "2004" or "2000, 2004". The first year is when the contents of the file were first created and the last year is the last year is when the contents were last modified.

Issue #4270, opened 2019-09-23
JerseyInvocation should override toString()


thomasmey:
Hi,

JerseyInvocation should override toString() with a sensible implementation. so if you log a given JerseyInvocation object it should at least print the request URI and the method of the contained requestContext.

jansupol:
Hi, sure, we would welcome your PR.

PR #4442, opened 2020-04-09
JerseyInvocation should override toString()

thomasmey:
Override toString() with a sensible implementation.
So if you log a given JerseyInvocation object at least the request method and
URI of the contained requestContext is printed.

See issue #4270


jansupol:
Please see the [wiki](https://github.com/eclipse-ee4j/jersey/wiki/Missing-ECA,-Missing-Sign-off-by).

jansupol:
Please adjust a copyright year in the file. 
Have you thought about using a `StringBuilder` in the `toString` method?

thomasmey:
Hi,

why use a StringBuilder here?
The JVM seems to optimise this itself anyway to a StringBuilder, the current code is more dense with the same performance:
see also https://stackoverflow.com/a/1532499

Issue #4003, opened 2018-12-05
Jersey Client - Response status code ignored if server closes connection


sumitkdey:
Hello,

If during a PUT call, the web server responds with the error status code and closes the connection (before the client finishes streaming all the data), jersey client raises an IOException and ignores the status code returned by the server.
```
Caused by: javax.ws.rs.ProcessingException: java.io.IOException: Error writing request body to server
	at org.glassfish.jersey.client.internal.HttpUrlConnector.apply(HttpUrlConnector.java:287)
	at org.glassfish.jersey.client.ClientRuntime.invoke(ClientRuntime.java:252)
	at org.glassfish.jersey.client.JerseyInvocation$1.call(JerseyInvocation.java:684)
	at org.glassfish.jersey.client.JerseyInvocation$1.call(JerseyInvocation.java:681)
	at org.glassfish.jersey.internal.Errors.process(Errors.java:315)
	at org.glassfish.jersey.internal.Errors.process(Errors.java:297)
	at org.glassfish.jersey.internal.Errors.process(Errors.java:228)
	at org.glassfish.jersey.process.internal.RequestScope.runInScope(RequestScope.java:444)
	at org.glassfish.jersey.client.JerseyInvocation.invoke(JerseyInvocation.java:681)
	at org.glassfish.jersey.client.JerseyInvocation$Builder.method(JerseyInvocation.java:437)
	at org.glassfish.jersey.client.JerseyInvocation$Builder.put(JerseyInvocation.java:326)
	at com.oracle.bmc.http.internal.ForwardingInvocationBuilder.put(ForwardingInvocationBuilder.java:134)
	at com.oracle.bmc.http.internal.RestClient.put(RestClient.java:351)
	... 7 more
Caused by: java.io.IOException: Error writing request body to server
	at sun.net.www.protocol.http.HttpURLConnection$StreamingOutputStream.checkError(HttpURLConnection.java:3518)
	at sun.net.www.protocol.http.HttpURLConnection$StreamingOutputStream.write(HttpURLConnection.java:3501)
	at org.glassfish.jersey.message.internal.CommittingOutputStream.write(CommittingOutputStream.java:229)
	at org.glassfish.jersey.message.internal.WriterInterceptorExecutor$UnCloseableOutputStream.write(WriterInterceptorExecutor.java:299)
	at org.glassfish.jersey.message.internal.ReaderWriter.writeTo(ReaderWriter.java:116)
[...]

PR #4460, opened 2020-05-07
Offer the Client (partial) response in ProcessingException.

jansupol:
In the HttpUrlConnector, when there is a response,
throw ResponseProcessingException instead of a simple
ProcessingException.

### DIFF
--- a/tests/integration/jersey-4003/pom.xml
+++ b/tests/integration/jersey-4003/pom.xml
@@ -1,4 +1,21 @@
 <?xml version="1.0" encoding="UTF-8"?>
+<!--
+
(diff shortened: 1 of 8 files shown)


tests/integration/jersey-4003/pom.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
```
senivam, line 2: No copyright header here

jansupol:
Fixes #4003

Issue #4446, opened 2020-04-21
Content type for Viewable can be wrong


michal-trnka:
I am using Viewable in ExceptionMapper. I cannot specify the correct content-type as it always gets overridden by `*/*`.

The ExceptionMapper
```
 public Response toResponse(NotFoundException e) {
        if (headers.getAcceptableMediaTypes().contains(MediaType.TEXT_HTML_TYPE)) {
            return Response
                    .status(Response.Status.NOT_FOUND)
                    .type(MediaType.TEXT_HTML)
                    .entity(new Viewable("/errors/error_404.ftl", new HashMap<String, Object>()))
                    .build();
        }
        return Response.status(Response.Status.NOT_FOUND).entity("Not found").build();
    }
```

I have dug into the source code and I found that content type defined on the resource is used and if not found, `*/*` is default one. It completely ignores the type set on the Response object.
Link to the place, where `*/*` is set: https://github.com/eclipse-ee4j/jersey/blob/7421515fbe3924691f057c859da32342d3f8d18c/ext/mvc/src/main/java/org/glassfish/jersey/server/mvc/internal/TemplateHelper.java#L108

michalhabala:
I have the exact same issue and really would appreciate a fix or some workaround. Thanks

PR #4461, opened 2020-05-11
Allow the user for overriding the default Viewable MediaType

jansupol:


jansupol:
Fixes #4446

PR #4464, opened 2020-05-13
Fix Jersey META-INF/services provider configuration file

jGauravGupta:


smillidge:
The service name filenames should also change if not already.

Issue #4462, opened 2020-05-12
InvocationInterceptors only used once when registered on a Client.


wirehell:
After registering an PreInvocationInterceptor or PostInvocationInterceptor on a Client, it is only used for the first request. 
The iterator containing the interceptors seems to be saved between requests.
This is on version 2.30.1.

jansupol:
fixed by #4467

PR #4467, opened 2020-05-14
Be able to use invocation interceptor for multiple requests

jansupol:
within a singe ClientRuntime


jansupol:
fixes #4462

PR #4469, opened 2020-05-14
ClientDestroyTest fix

jansupol:


jansupol:
Prevent intermittent test failure on Travis caused by the Garbage Collector invoking a finalize method on ClientRuntime before the Client is closed.

Issue #3446, opened 2016-10-07
SAXParserFactory doesn't disable XML entity expansion


jerseyrobot:
The DocumentBuilderFactory and XmlInputFactory providers aren't vulnerable to Billion Laughs attacks (exponential entity expansion) by disabling entity expansion altogether.

But the SAXParserFactory provider is only disabling external entities, and it's possible to use the following input to generate a very high memory consumption (specially if the application using it allows parallel requests):

```
<?xml version="1.0" encoding="UTF-8"?> 
<!DOCTYPE lolz [
  <!ENTITY lol "lollollollollollollol[...]">
  <!ENTITY lol2 "&lol;&lol;&lol;&lol;&lol;&lol;&lol;&lol;&lol;&lol;">
  <!ENTITY lol3 "&lol2;&lol2;&lol2;&lol2;&lol2;&lol2;&lol2;&lol2;&lol2;&lol2;">
  <!ENTITY lol4 "&lol3;&lol3;&lol3;&lol3;&lol3;&lol3;&lol3;&lol3;&lol3;&lol3;">
  <!ENTITY lol5 "&lol4;&lol4;&lol4;&lol4;&lol4;&lol4;&lol4;&lol4;&lol4;&lol4;">
  <!ENTITY lol6 "&lol5;&lol5;&lol5;&lol5;&lol5;&lol5;&lol5;&lol5;">
]>
<Quote> 
<fName>FIRST NAME &lol6;</fName> 
</Quote>
```

There is I actually a limit, but very high (like 100000 I think, and couldn't configure it with the entityExpansionLimit JDK property). If it doesn't work try removing the last level of recursion. Even if the depth is limited, there is no maximum for the expanded size so adding a few kilobytes to the entity text would make the attack successful.

Check disallow-doctype-decl feature in this example:
[https://gist.github.com/asudhakar02/45e2e6fd8bcdfb4bc3b2](https://gist.github.com/asudhakar02/45e2e6fd8bcdfb4bc3b2)
#### Affected Versions
[2.24]

rwalkerands:
There is also the system property `jdk.xml.totalEntitySizeLimit`. It would be nice to be able to set this for incoming message bodies.

gjoseph:
Hi there,

For some reason, this issue has started being reported as a vulnerability by SourceClear - https://www.sourceclear.com/vulnerability-database/security/xml-entity-expansion-xee-/java/sid-3606 - with no CVE, however.

[...]

PR #4472, opened 2020-05-16
Set additional security features on SecureSaxParserFactory.

jansupol:
Provide a way to override.


jansupol:
Solves #3446.

PR #4473, opened 2020-05-19
Execute tests in servlet-2.5-mvc-3

jbescos:
Locally I'm having an error in one test that checks one ISO character encoding, but it seems it is running fine in Travis.


jbescos:
Same issue in Travis, as far as I can see. I will try to fix it.

jbescos:
I deployed the war file in a local Jetty 9 and I saw the charset was not specified in the header.

The test was reading the page. As the charset was not specified it was reading it as UTF-8 by default.

I had to add a filter to explicitly set the header with the required charset.

PR #4475, opened 2020-05-19
Set Bundle-ActivationPolicy on core-common module

jansupol:


jansupol:
Fixes #3588

PR #4497, opened 2020-06-01
Jakartification: arrange properties in the main pom

senivam:


senivam:
arrangement of properties and attempt to comment out not used properties which are left from previous releases.
Version properties are grouped into Jakartified wo CQs/Jakartified with CQs/others groups

PR #4498, opened 2020-06-02
Fix resources in containers

jansupol:


containers/grizzly2-http/pom.xml
```diff
@@ -61,19 +61,6 @@
                 <inherited>true</inherited>
             </plugin>
         </plugins>
-
-        <resources>
-            <resource>
-                <directory>${basedir}/src/main/java</directory>
```
Tomas-Kraus, line 67: Why do you have META-INF in src/main/java? I would move it to src/main/resources.

containers/jetty-http/pom.xml
```diff
@@ -73,29 +73,11 @@
...
             <resource>
                 <directory>${basedir}/src/main/resources</directory>
                 <filtering>true</filtering>
             </resource>
         </resources>
     </build>
 
-    <profiles>
-        <profile>
-            <id>testsSkipJdk6</id>
```
Tomas-Kraus, line 91: Does it make sense to support such historical artifact?

senivam: probably that is why this profile is being removed...

containers/grizzly2-http/pom.xml
senivam: this whole block is redundant and is being removed from Jersey...

containers/jetty-http/pom.xml
Tomas-Kraus: Yes. My fault, I somehow did not realize that this is "red". :)

containers/grizzly2-http/pom.xml
Tomas-Kraus: Yes, but still, things like META-INF shall be in `src/main/resources`.

PR #4506, opened 2020-06-16
Clean unused dependencies

senivam:


senivam:
CQ #[19861](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=19861) is already resolved

Issue #4507, opened 2020-06-16
Intermittent HK2 ServiceLocatorImpl has been shut down


jansupol:
When the reference on ClientRuntime is not kept in the code (such as in the Response), the ClientRuntime can get garbage collected. HK2 ServiceLocator is closed then.

Example:
```
entity = ClientBuilder.newClient().target().request().get(<some.entity.that.uses.messagebodyworkers>)
entity.read()
```
entity can be `EventInput`, `InboundEvent`, `ChunkedInput`,...

The following exception can be thrown:
> java.lang.IllegalStateException: ServiceLocatorImpl(__HK2_Generated_13,13,1665167146) has been shut down
	at org.jvnet.hk2.internal.ServiceLocatorImpl.checkState(ServiceLocatorImpl.java:2383)
	at org.jvnet.hk2.internal.ServiceLocatorImpl.internalGetInjecteeDescriptor(ServiceLocatorImpl.java:524)
	at org.jvnet.hk2.internal.ServiceLocatorImpl.getInjecteeDescriptor(ServiceLocatorImpl.java:571)
	at org.jvnet.hk2.internal.IterableProviderImpl.justInTime(IterableProviderImpl.java:75)
	at org.jvnet.hk2.internal.IterableProviderImpl.get(IterableProviderImpl.java:84)

PR #4508, opened 2020-06-16
Fix intermittent premature ClientRuntime finalization

jansupol:


jansupol:
Fixes #4507

PR #4511, opened 2020-06-17
Fixed HK2 AbstractActiveDescriptor Test

jansupol:


jansupol:
HK2 Binder registering was added back in 2.29

PR #4512, opened 2020-06-18
project-info plugin configuration

senivam:


jansupol:
[Effectively turn-on](https://maven.apache.org/plugins/maven-site-plugin/site-mojo.html) generating project-info.

PR #4516, opened 2020-06-20
add Dependabot

sullis:
https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/


jansupol:
I am not convinced this is what we want. We have multiple branches, and the current master branch definitely cannot use the latest dependencies; the latest dependencies are Jakarta EE 9 incompatible dependencies.

I also am not convinced that for the 3.x branch we would want it; each new dependency requires a CQ, and we may not want to use the latest, there, too.

On the other hand, reports about new versions related to security patches may come handy.

arjantijms:
@jansupol 
>  each new dependency requires a CQ

Just thinking out loud, but maybe we could look at a CQaBot? A bot that semi-automatically files a list of CQs for all dependencies?

PR #4520, opened 2020-06-23
Test Gzip + JSP

jbescos:
Adding a new example that tests JSP is GZipped with no issues.

Recently an issue was found in one servlet container, and this demonstrates the issue is not related to jersey.


jbescos:
I have to move it to servlet 4.x because 2.5 is going to be deprecated.

jbescos:
@senivam @jansupol what is your opinion on the commit:
https://github.com/eclipse-ee4j/jersey/pull/4520/commits/7676148632e8b6665604fa6e9e74794833dc233a

I found that in Weblogic, after the method HttpServletResponse#flushBuffer is invoked, it will refuse to add new headers in the response. It is possible that in other Servlet implementations it behaves similar. It is not the case of Jetty because I tested it is fine.

I think it is possible to fix it in Weblogic instead of Jersey but probably it makes more sense to do this in Jersey because other implementations could also make problems. Let me know if you think in other way.

senivam:
@jbescos - just a question - why you do not use flush() method of the PrintWriter which actually wraps responseStream?

jbescos:
> @jbescos - just a question - why you do not use flush() method of the PrintWriter which actually wraps responseStream?

I already tried it, but the problem is Weblogic creates internally a ServletOutputStream that wraps the original (weblogic.servlet.internal.ServletOutputStreamImpl). This wrapper will not accept more headers after flush.

Then when ServletOutputStream#flush is executed, it is too late and it reach the status of not accepting more headers.

Issue #4437, opened 2020-04-08
InboundJaxrsResponse.hasEntity() returns false when the entity is both Buffered and read before.


IrminAuwerda:
Version: Jersey-client 2.29.1

Scenario:

1. Receive an `InboundJaxrsResponse` after a HTTP call (a Response).
2. Call `Response.hasEntity()` returns **TRUE** (calls `EntityInputStream.isEmpty()` which checks if there is an input stream, and it has more bytes)
3. Call `Response.bufferEntity()` returns **TRUE** (this will replace the `InputStream` in `InboundMessageContext.EntityContent` with a `ByteArrayOutputStream`)
4. Read the entity (this will read the full `ByteArrayOutputStream`, setting its offset to the latest byte)
5. Call `Response.hasEntity()` returns **FALSE** (the implementation `EntityInputStream.isEmpty()` sees that `input` is not null or closed, but the offset is at the last byte. This offset at the end of the stream causes the check `in.read()` to return **-1**, making the check assume that there are 0 bytes in the stream (while in reality there are 0 _more_ bytes in the stream)).

I've tested the behaviour with Wiremock.

A potential fixes could be: On `InboundJaxrsResponse.hasEntity()` first checking if (buffered) and if so resetting the stream by calling `entityContent.reset()` (similar to what is done for reads).

lkorth:
It looks like #4523 has fixed this and this can be closed now.

IrminAuwerda:
I just looked at the commit @lkorth. That fixes it. Closed.
And thanks a lot for fixing :-)!

PR #4523, opened 2020-06-24
Response.hasEntity to return true if buffered after readEntity

jansupol:


jansupol:
Fixes #4437

jbescos (approved):

PR #4528, opened 2020-07-13
Support TLSv1.3

jbescos:
I fixed recently the same issue in Tyrus.
https://github.com/eclipse-ee4j/tyrus/pull/707

Find in that PR the information.


connectors/jdk-connector/src/main/java/org/glassfish/jersey/jdk/connector/internal/SslFilter.java
```diff
@@ -416,12 +419,8 @@ private boolean doHandshakeStep(ByteBuffer networkData) {
                         }
 
                         case FINISHED: {
-                            /* According to SSLEngine javadoc FINISHED status can be returned only in SSLEngineResult,
-                            but just to make sure we don't end up in an infinite loop when presented with an SSLEngine
-                            implementation that does not respect this:*/
-                            stepFinished = true;
-                            handshakeFinished = true;
-                            break;
+                            throw new IllegalStateException("Trying to handshake, but SSL engine not in HANDSHAKING state."
```
senivam, line 422: I would advice to move direct text into localization.properties and use LocalizableMessageFactory instead.

jbescos: @senivam although that is unreachable (in theory), I decided to keep it as it was. It will not throw exception.

I added the localization property for the case of NOT_HANDSHAKING, because it was throwing the same exception.

I also modified the test to force to test all supported SSL protocols. For the case of JDK11 the TLSv13 is included.

jbescos: I just verified with Wireshark that we need to execute the tests in different JVMs. So I had to set `<reuseForks>false</reuseForks>` and move each protocol to a different test class. It seems when you use `@Parametrized` tests are executed in the same JVM.

connectors/jdk-connector/src/main/resources/org/glassfish/jersey/jdk/connector/internal/localization.properties
```diff
@@ -71,3 +71,4 @@ thread.pool.core.size.too.small="Core thread pool size cannot be smaller than 0.
 http.connection.establishing.illegal.state="Cannot try to establish connection if the connection is in other than CREATED state\
   . Current state: {0}.
 http.connection.not.idle="Http request cannot be sent over a connection that is in other state than IDLE. Current state: {0}" 
+http.connection.invalid.handshake.status="Trying to handshake, but SSL engine not in HANDSHAKING state. SSL filter state: {0}"
```
senivam, line 74: please consider updating of copyright year in this file.

jansupol (approved):

PR #4531, opened 2020-07-21
Fix #4522 - override write method in LoggingStream

### LINKED ISSUE #4522: org.glassfish.jersey.logging.LoggingInterceptor.LoggingStream does not override write(byte[] b, int off, int len)
less efficient

tvallin:

### DIFF
--- a/core-common/src/main/java/org/glassfish/jersey/logging/LoggingInterceptor.java
+++ b/core-common/src/main/java/org/glassfish/jersey/logging/LoggingInterceptor.java
@@ -290,18 +290,11 @@ abstract class LoggingInterceptor implements WriterInterceptor {
         public void write(byte[] ba, int off, int len) throws IOException {
             if ((off | len | ba.length - (len + off) | off + len) < 0) {
                 throw new IndexOutOfBoundsException();
-            } else {
-                if ((baos.size() + len) <= maxEntitySize) {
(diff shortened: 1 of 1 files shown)


core-common/src/main/java/org/glassfish/jersey/logging/LoggingInterceptor.java
```diff
@@ -285,6 +285,24 @@ public void write(final int i) throws IOException {
...
+                        baos.write(ba[off + i]);
+                        out.write(ba[off + i]);
+                    }
+                } else {
+                    for (int i = 0; i < len; ++i) {
+                        out.write(ba[off + i]);
+                    }
+                }
+            }
+        }
```
senivam, line 305: is there any objection to rewrite it like this:
```
        @Override
        public void write(byte[] ba, int off, int len) throws IOException {
            if ((off | len | ba.length - (len + off) | off + len) < 0) {
                throw new IndexOutOfBoundsException();
            }
            if ((baos.size() + len) <= maxEntitySize) {
                baos.write(ba, off, len);
            } 
            out.write(ba, off, len);
        }
```
?

tvallin: No objection, thanks for the tips !

PR #4541, opened 2020-08-04
Make Kryo use setRegistrationRequired(true) by default

jansupol:

### DIFF
--- a/incubator/kryo/src/test/java/org/glassfish/jersey/kryo/KryoContextResolver.java
+++ b/incubator/kryo/src/test/java/org/glassfish/jersey/kryo/KryoContextResolver.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright (c) 2020 Oracle and/or its affiliates. All rights reserved.
+ *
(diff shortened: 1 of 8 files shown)


jansupol:
See https://github.com/EsotericSoftware/kryo/issues/398

incubator/kryo/src/test/java/org/glassfish/jersey/kryo/KryoContextResolver.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015, 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2020 Oracle and/or its affiliates. All rights reserved.
```
senivam, line 2: shouldn't it be 2015, 2020?

jansupol: No, this is a new file. Git thinks the file was renamed and changed, but the previous file was erased and the new file created. The files are completely different, with different purpose

Issue #3738, opened 2018-04-20
Place holder issue for Pull Request 3738


jerseyrobot:
This is a place holder issue for Pull request from https://github.com/jersey/jersey/pull/3738

MikeTheyCallMe:
I see this PR is closed but not merged? 

My project is using still using version 2.25.1 of the jersey-client and we just recently encountered the concurrency bug as reported on 12/10/2015 by glassfishrobot here: https://github.com/jersey/jersey/issues/3293

And also reported in version 2.27 by two users in the summers of 2018 and 2019(!): https://github.com/eclipse-ee4j/jersey/issues/3293

Was wondering if this was ever resolved?

ahaasler:
I'm also interested in knowing if this has been resolved or just closed.

PR #4566, opened 2020-09-15
Fix custom SSLSocketFactory not being set because of an unsafe lazy-initialization in JDK

### LINKED ISSUE #3293: HttpUrlConnector.secureConnection not always setting custom SSL socket factory
To be more precise, due to changes introduced by #3171 ([https://github.com/jersey/jersey/commit/57f5dafed38d27fcb70e7f4ee5051554118d305a#diff-2af8c9e0f0f9963b8b1ce356cf17ecb7](https://github.com/jersey/jersey/commit/57f5dafed38d27fcb70e7f4ee5051554118d305a#diff-2af8c9e0f0f9963b8b1ce356cf17ecb7)), a custom SSL socket factory is not set on an HttpsURLConnection 'suc' in HttpUrlConnector unless HttpsURLConnection.getDefaultSSLSocketFactory returns the exact same object as suc.getSSLSocketFactory(). This is not a robust way of determining if the custom socket factory has already been set because HttpsURLConnection.getDefaultSSLSocketFactory() can be called simultaneously by multiple threads, returning a different object for each invocation, thus resulting in up to n-1 threads getting a mismatch in HttpUrlConnector.secureConnection, even though suc.getSSLSocketFactory() is actually returning a default SSL socket factory (just not necessarily the exact instance we expected, which is ultimately irrelevant).

This is fairly easy to reproduce, given an HTTPS server with a self-signed certificate. The following code is not a self-contained unit test, but will demonstrate the issue nevertheless, by spitting out at least one SSLHandshakeException (caused by the custom SSLContext not being used). If numThreads is set to 1, then it will never fail due to a certificate validation error. The code will also run without error on Jersey 2.21 or earlier.

```
import javax.net.ssl.*;
import javax.ws.rs.client.Client;
import javax.ws.rs.client.ClientBuilder;
import java.security.SecureRandom;
import java.security.cert.CertificateException;
import java.security.cert.X509Certificate;

public class HttpsTest {

    private static SSLContext sslContext = null;

    private static synchronized SSLContext getSslContext() {

        if (sslContext == null) {
            try {
sslContext = SSLContext.getInstance("TLS");
sslContext.init(null, new TrustManager[]{new X509TrustManager() {
    public void checkClientTrusted(X509Certificate[] arg0, String arg1) throws CertificateException {
    }

    public void checkServerTrusted(X509Certificate[] arg0, String arg1) throws CertificateException {
    }

    public X509Certificate[] getAcceptedIssuers() {
        return new X509Certificate[0];
    }

}}, new SecureRandom());

            } catch (Exception e) {
throw new RuntimeException(e);
            }
        }

        return sslContext;
    }

    private Client client;

    public HttpsTest() {

        // Create a Client instance that won't complain about self-signed SSL certs
        client = ClientBuilder.newBuilder().sslContext(getSslContext()).hostnameVerifier(new LenientHostnameVerifier())
.build();
    }

    private void runTest() {

        try {
            client.target("https://lists.gno.org/self-signed.html").request().get();
        } catch (Exception e) {
            e.printStackTrace();
        }
    }

    public static void main(String... args) throws InterruptedException {

        final int numThreads = 4;

        for (int i = 1; i <= numThreads; i++) {

            new Thread(() -> new HttpsTest().runTest()).start();
        }

        Thread.sleep(10000L);
    }

    /**
     * A HostnameVerifier that always returns true
     */
    private class LenientHostnameVerifier implements HostnameVerifier {

        @Override
        public boolean verify(String s, SSLSession sslSession) {
            return true;
        }
    }
}
```

I would suggest that the change to HttpUrlConnector.secureConnection could be reverted, or at least if the purpose of it was to avoid redundant calls to setSSLSocketFactory (the cost of which is arguably insignificant), that comparing suc.getSSLSocketFactory to sslSocketFactory.get() might be more appropriate than comparing it to HttpsUrlConnection.getDefaultSSLSocketFactory().
#### Environment
Ubuntu 14.04, JDK 1.8.0_45
#### Affected Versions
[2.22]

ahaasler:
Hi,

this PR fixes #3293 with an approach similar to the one contributed by @kevinconaway on jersey/jersey#3738 (check up his well written explanation). We (@amplia-iiot) think his contribution was forgotten when the project was transitioned (#3738) over to the Eclipse Foundation and have implemented this more maintainable approach (the change is easier to notice when looking at `secureConnection(JerseyClient, HttpURLConnection)` and avoids multiple calls to `HttpsURLConnection.getDefaultSSLSocketFactory()` making it cleaner).

I've not included the original test by @kevinconaway because I'm not aware of his status regarding the Eclipse Contributor Agreement, but it can be checked at [@6a8e80ec625f2ce9a7585dc45e645fab61aa504c](https://github.com/amplia-iiot/jersey/commit/6a8e80ec625f2ce9a7585dc45e645fab61aa504c) ([bugfix/jersey\#3293-test](https://github.com/amplia-iiot/jersey/tree/bugfix/jersey%233293-test)).

We think this is the way to go for Jersey until the JDK properly makes this lazy-initialization thread-safe like the [implementation](https://android.googlesource.com/platform/libcore/+/af8ca5f/luni/src/main/java/javax/net/ssl/HttpsURLConnection.java) in Android's JDK.

Nevertheless, we are open to feedback regarding this change and discussing alternate approaches.

### DIFF
--- a/core-client/src/main/java/org/glassfish/jersey/client/internal/HttpUrlConnector.java
+++ b/core-client/src/main/java/org/glassfish/jersey/client/internal/HttpUrlConnector.java
@@ -71,6 +71,9 @@ public class HttpUrlConnector implements Connector {
 
     private static final Logger LOGGER = Logger.getLogger(HttpUrlConnector.class.getName());
     private static final String ALLOW_RESTRICTED_HEADERS_SYSTEM_PROPERTY = "sun.net.http.allowRestrictedHeaders";
+    // Avoid multi-thread uses of HttpsURLConnection.getDefaultSSLSocketFactory() because it does not implement a
+    // proper lazy-initialization. See https://github.com/jersey/jersey/issues/3293
+    private static final SSLSocketFactory DEFAULT_SSL_SOCKET_FACTORY = HttpsURLConnection.getDefaultSSLSocketFactory();
     // The list of restricted headers is extracted from sun.net.www.protocol.http.HttpURLConnection
     private static final String[] restrictedHeaders = {
             "Access-Control-Request-Headers",
@@ -299,7 +302,7 @@ public class HttpUrlConnector implements Connector {
                 suc.setHostnameVerifier(verifier);
             }
 
-            if (HttpsURLConnection.getDefaultSSLSocketFactory() == suc.getSSLSocketFactory()) {
+            if (DEFAULT_SSL_SOCKET_FACTORY == suc.getSSLSocketFactory()) {
                 // indicates that the custom socket factory was not set
                 suc.setSSLSocketFactory(sslSocketFactory.get());
             }
--- a/tests/e2e-client/src/test/java/org/glassfish/jersey/tests/e2e/client/connector/ssl/SslHttpUrlConnectorTest.java
+++ b/tests/e2e-client/src/test/java/org/glassfish/jersey/tests/e2e/client/connector/ssl/SslHttpUrlConnectorTest.java
@@ -17,10 +17,18 @@
 package org.glassfish.jersey.tests.e2e.client.connector.ssl;
 
 import java.io.IOException;
+import java.io.PrintWriter;
+import java.io.StringWriter;
 import java.net.HttpURLConnection;
 import java.net.InetAddress;
 import java.net.Socket;
 import java.net.URL;
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.CyclicBarrier;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 
 import javax.ws.rs.client.Client;
 import javax.ws.rs.client.ClientBuilder;
@@ -77,6 +85,61 @@ public class SslHttpUrlConnectorTest extends AbstractConnectorServerTest {
         final Response response = client.target(Server.BASE_URI).path("/").request().get();
         assertEquals(200, response.getStatus());
         assertTrue(socketFactory.isVisited());
+    }
+
+    /**
+     * Test for https://github.com/jersey/jersey/issues/3293
+     *
+     * @author Kevin Conaway
+     */
+    @Test
+    public void testConcurrentRequestsWithCustomSSLContext() throws Exception {
+        final SSLContext sslContext = getSslContext();
+
+        final Client client = ClientBuilder.newBuilder()
+            .sslContext(sslContext)
+            .register(HttpAuthenticationFeature.basic("user", "password"))
+            .register(LoggingFeature.class)
+            .build();
+
+        int numThreads = 5;
+        CyclicBarrier barrier = new CyclicBarrier(numThreads);
+        ExecutorService service = Executors.newFixedThreadPool(numThreads);
+        List<Exception> exceptions = new CopyOnWriteArrayList<>();
+
+        for (int i = 0; i < numThreads; i++) {
+            service.submit(() -> {
+                try {
+                    barrier.await(1, TimeUnit.MINUTES);
+                    for (int call = 0; call < 10; call++) {
+                        final Response response = client.target(Server.BASE_URI).path("/").request().get();
+                        assertEquals(200, response.getStatus());
+                    }
+                } catch (Exception ex) {
+                    exceptions.add(ex);
+                }
+            });
+        }
+
+        service.shutdown();
+
+        assertTrue(
+            service.awaitTermination(1, TimeUnit.MINUTES)
+        );
+
+        assertTrue(
+            toString(exceptions),
+            exceptions.isEmpty()
+        );
+    }
+
+    private String toString(List<Exception> exceptions) {
+        StringWriter writer = new StringWriter();
+        PrintWriter printWriter = new PrintWriter(writer);
+
+        exceptions.forEach(e -> e.printStackTrace(printWriter));
+
+        return writer.toString();
     }
 
     public static class CustomSSLSocketFactory extends SSLSocketFactory {


jansupol:
The [test ](https://github.com/amplia-iiot/jersey/commit/6a8e80ec625f2ce9a7585dc45e645fab61aa504c) nicely tests what is fixed by this PR and I think it would be good to include it. Given that the original PR of @kevinconaway contained the code, and it could have been contributed to the Eclipse Foundation along with other Jersey code, I think it is safe to include it.

kevinconaway:
Please let me know if you need me to sign a CLA or something for the original code.

ahaasler:
I've added you with the same name and email as your original commit:

```
Add test for concurrent custom SSLSocketFactory

Test made by Kevin Conaway @kevinconaway

Also-by: Kevin Conaway <kevin.conaway@walmart.com>
Signed-off-by: Adrian Haasler García <adrian.haasler@amplia.es>
```

I don't know if the `Also-by` section causes the _eclipsefdn/eca_ check to fail if it has not been signed. Let me know if you sign the [agreement](https://www.eclipse.org/legal/ECA.php) with a different email so I can change the commit message.

jansupol:
@kevinconaway If you can sign the ECA, it would be great. Please follow the steps [here](https://wiki.eclipse.org/Development_Resources/Contributing_via_Git#Eclipse_Contributor_Agreement).

ahaasler:
How should I proceed with this contribution?

jansupol:
@kevinconaway We would be happy if you help us here. If you cannot, please let us know. Thank you.

kevinconaway:
I have signed the ECA

ahaasler:
Thanks @kevinconaway, did you sign it with the same email that appears on the commit message https://github.com/amplia-iiot/jersey/commit/6a8e80ec625f2ce9a7585dc45e645fab61aa504c or should I update it?

ahaasler:
Some tests have failed:

For [JDK11 -Ptravis_e2e/https://travis-ci.org/github/eclipse-ee4j/jersey/jobs/746712439]
```
[ERROR]   ClientExecutorTest.testDefaultExecutorRx:88 
Expected: a string containing "jersey-client-async-executor"
     but: was ""
```

For [JDK15 -Ptravis_e2e/https://travis-ci.org/github/eclipse-ee4j/jersey/jobs/746712442]
```
[ERROR]   BroadcasterTest.test:188 Waiting for resultsA1 to be complete failed.
```

I doubt they have anything to do with the change in the first commit, how may I help?

kevinconaway:
>Thanks @kevinconaway, did you sign it with the same email that appears on the commit message amplia-iiot@6a8e80e or should I update it?

You may update it to the email address that was used to sign the ECA

jbescos:
Are we sure this fixes the issue?. I don't agree with what is written here: https://github.com/jersey/jersey/issues/3293

> HttpsURLConnection.getDefaultSSLSocketFactory() can be called simultaneously by multiple threads, returning a different object for each invocation...

It is true that HttpsURLConnection.getDefaultSSLSocketFactory() is not thread safe, but it doesn't really matter because it relies in SSLSocketFactory.getDefault(). This one is thread safe and it will always returns the same instance.

So it is possible that 1+ threads invokes HttpsURLConnection.getDefaultSSLSocketFactory() when defaultSSLSocketFactory is null, and then it is set more than 1 time. But it doesn't matter because the instance is the same.

ahaasler:
This is the piece of code in the SDK that contains the lazy-initialization unsafe code:
```java
public static SSLSocketFactory getDefaultSSLSocketFactory() {
    if (defaultSSLSocketFactory == null) {
        defaultSSLSocketFactory =
            (SSLSocketFactory)SSLSocketFactory.getDefault();
    }
    return defaultSSLSocketFactory;
}
```

It doesn't matter that `SSLSocketFactory.getDefault()` is thread-safe because `getDefaultSSLSocketFactory` may return to the second thread an object that is not null but may not have been constructed yet. More info on the matter:
- http://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html
- https://www.oracle.com/technical-resources/articles/javase/bloch-effective-08-qa.html

This causes the jersey comparison to fail and the custom SSLSocketFactory is not set.

We have seen this in production and this patch solves it.

jbescos:
> ```java
> SSLSocketFactory
> ```

These are the first lines of the SSLSocketFactory.getDefault():

```
/*     */   public static synchronized SocketFactory getDefault()
/*     */   {
/*  89 */     if (theFactory != null) {
/*  90 */       return theFactory;
/*     */     }
```
The method is synchronized and once the factory is instanced it doesn't instance it more. It is not possible to get 2 different instances from that method.

To put it in one example, lets imagine 2 threads are invoking `getDefaultSSLSocketFactory()` and `defaultSSLSocketFactory` is null, so both threads are invoking `SSLSocketFactory.getDefault()`.

First thread will lock the method `SSLSocketFactory.getDefault()` meanwhile the other thread is waiting. 'theFactory' is instanced , returned to the first thread and the lock is released. Then the second thread goes in and receives 'theFactory' already instanced.

jbescos (approved):

jansupol (approved):

dvogelbacher:
I don't think that this fully fixes https://github.com/eclipse-ee4j/jersey/issues/3293 and it might even make it worse in some circumstances. Different instances of `HttpsUrlConnection` can still have a different SSLSocketFactories set. A simple unit test shows this:
 ```
    @Test
    public void testDefaultSSLSocketFactory() throws Exception {
        int numThreads = 5;
        ExecutorService service = Executors.newFixedThreadPool(numThreads);
        for (int i = 0; i < numThreads; i++) {
            service.execute(() -> {
                try {
                    URL url = new URL("https://github.com/");
                    HttpsURLConnection connection = (HttpsURLConnection) url.openConnection();
                    System.out.println(connection.getSSLSocketFactory());
                } catch (IOException e) {
                    e.printStackTrace();
                }
            });
        }
        System.out.println(HttpsURLConnection.getDefaultSSLSocketFactory());
    }
```
    
This prints:
    
```
sun.security.ssl.SSLSocketFactoryImpl@241fbec
sun.security.ssl.SSLSocketFactoryImpl@4f6186b1
sun.security.ssl.SSLSocketFactoryImpl@21998385
sun.security.ssl.SSLSocketFactoryImpl@6c75fa5b
sun.security.ssl.SSLSocketFactoryImpl@592ee1f8
sun.security.ssl.SSLSocketFactoryImpl@76c94d54
```

That means that for those connections we won't set the custom ssl factory in `HttpUrlConnector`. 
The reason that I say that this fix might have made things worse is that we now only read the default ssl factory in `HttpUrlConnector` once. If that differs from the default ssl factory set in `HttpsUrlConnector`, then this problem is irrecoverable now. All connections will fail. I'm actually a bit surprised that the unit test that was added as part of this PR works. Am I missing something?


Overall, the way that jersey decides whether to set the custom SSLSocketFactory or not seems very brittle given that the underlying java classes aren't thread safe.

I'm working around this issue now by specifying a `ConnectorProvider` on my client, which sets the custom ssl factory:
```
        return new HttpUrlConnectorProvider().connectionFactory(url -> {
            HttpURLConnection connection = (HttpURLConnection) url.openConnection();
            if (connection instanceof HttpsURLConnection) {
                HttpsURLConnection httpsConnection = (HttpsURLConnection) connection;
                httpsConnection.setSSLSocketFactory(sslContext.getSocketFactory());
            }
            return connection;
        });
```

PR #4588, opened 2020-10-01
Do not retrieve the headers when request was rejected by server when using Expect header

skunjuma:
HTTPURLConnection throws ProtocolException with "Server rejected operation" message when a non-100 response is received and does not read the full response body or store the headers. HttpUrlConnector._apply method tries to set the headers to ClientResponse using HTTPURLConnection.getHeaderFields. getHeaderFields method first tries to getInputStream which in turn call getOutputStream. This results in sending the request again since both streams are null due the the ProtocolException. Get the header fields only when exception is not ProtocolException and the message is not "Server rejected operation".


jansupol:
This should be fixed by https://github.com/eclipse-ee4j/jersey/pull/4613.

senivam:
closing in favor of #4613

PR #4629, opened 2020-10-31
merge of actual master to jakartified jersey

senivam:


tests/integration/pom.xml
```diff
@@ -126,6 +126,7 @@
             </activation>
             <modules>
                 <module>async-jersey-filter</module>
+                <module>externalproperties</module>
```
jansupol, line 129: Oddly enough, [this line](https://github.com/eclipse-ee4j/jersey/blob/684bbf85f996121ff0a1ba2b16a383f18bbf6e9f/tests/integration/pom.xml#L126) did not make it to this merge request

jansupol (changes requested):

tests/integration/pom.xml
senivam: it's one commit after merge, because otherwise this module does not work in JDK 1.8 (due to jakartified Jetty), so it shall be moved to profile with other tests which uses jetty (with JDK 11+)

jansupol (approved):

PR #4641, opened 2020-11-19
Support for new property to ignore responses in exceptions thrown by the Client API

spericas:
Support for new property to ignore responses in exceptions thrown by the Client API. If the property jersey.config.client.ignoreExceptionResponse is set to true, any response in an exception thrown by the Client API will be mapped to an empty response that only includes the status code of the original one. This is to prevent accidental leaks of confidential data.

Signed-off-by: Santiago Pericasgeertsen <santiago.pericasgeertsen@oracle.com>

### DIFF
--- a/core-client/src/main/java/org/glassfish/jersey/client/ClientProperties.java
+++ b/core-client/src/main/java/org/glassfish/jersey/client/ClientProperties.java
@@ -168,6 +168,27 @@ public final class ClientProperties {
     public static final String USE_ENCODING = "jersey.config.client.useEncoding";
 
     /**
+     * Ignore a response in an exception thrown by the client API by not forwarding
+     * it to this service's client. A value of {@code true} indicates that responses
+     * will be ignored, and only the response status will return to the client. This
+     * property will normally be specified as a system property; note that system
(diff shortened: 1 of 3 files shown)


jansupol (approved):

senivam (approved):

dansiviter:
@spericas I'm interested to learn a little more about the use-case. Ignoring the response at the client-side feels like closing the stable door _after_ the horse has bolted. Can you explain a little more?

jansupol:
@dansiviter This is really only about preventing accidental leaks of confidential data, should the response contain them, such as from the third-party servers, similar to the test case (in the PR), when the `second` resource would be the confidential endpoint. The response is meant for the `first` resource, but not for the client requesting the `first` resource. 
While the users should use a try-catch block for handling the exceptions, they rarely do. The option is meant to be the framework support for customers who do not want to propagate the confidential info, but for some reason, they cannot modify their code.

dansiviter:
@jansupol Thanks. I'd hope the sensitive error data would not even be transmitted over the wire but I can see how poor error handling could lead to this also.

PR #4647, opened 2020-11-30
Add support RFC 5987 for attribute filename* in HTTP header Content-Disposition

aserkes:
Add support RFC 5987 for attribute filename* in HTTP header Content-Disposition

### DIFF
--- a/media/multipart/src/main/java/org/glassfish/jersey/media/multipart/ContentDisposition.java
+++ b/media/multipart/src/main/java/org/glassfish/jersey/media/multipart/ContentDisposition.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2010, 2020 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
--- a/media/multipart/src/main/java/org/glassfish/jersey/media/multipart/ContentDisposition.java
+++ b/media/multipart/src/main/java/org/glassfish/jersey/media/multipart/ContentDisposition.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2020 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2010, 2021 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
--- a/media/multipart/src/main/java/org/glassfish/jersey/media/multipart/ContentDisposition.java
+++ b/media/multipart/src/main/java/org/glassfish/jersey/media/multipart/ContentDisposition.java
@@ -20,6 +20,7 @@ import java.text.ParseException;
 import java.util.Collections;
 import java.util.Date;
 import java.util.Map;
+import java.util.regex.Pattern;
 
 import org.glassfish.jersey.message.internal.HttpDateFormat;
 import org.glassfish.jersey.message.internal.HttpHeaderReader;
@@ -40,6 +41,8 @@ public class ContentDisposition {
     private Date modificationDate;
     private Date readDate;
     private long size;
+    private static final Pattern FILENAME_EXT_VALUE_PATTERN =
+            Pattern.compile("(UTF-8|ISO-8859-1)'([a-z]{2,8}(-[a-z0-9]+)?)?'(%[a-f0-9]{2}|[a-z0-9!#$&+.^_`|~-])+");
 
     protected ContentDisposition(final String type, final String fileName, final Date creationDate,
                                  final Date modificationDate, final Date readDate, final long size) {
@@ -182,6 +185,11 @@ public class ContentDisposition {
 
     private void createParameters() throws ParseException {
         fileName = parameters.get("filename");
+
+        String fileNameExt = parameters.get("filename*");
+        if (fileNameExt != null && FILENAME_EXT_VALUE_PATTERN.matcher(fileNameExt).matches()) {
+            fileName = fileNameExt;
+        }
 
         creationDate = createDate("creation-date");
 
--- a/media/multipart/src/main/java/org/glassfish/jersey/media/multipart/ContentDisposition.java
+++ b/media/multipart/src/main/java/org/glassfish/jersey/media/multipart/ContentDisposition.java
@@ -225,15 +225,17 @@ public class ContentDisposition {
                 return fileNameExt;
             }
 
+            final String charset = matcher.group(CHARSET_GROUP_NAME);
             if (matcher.group(CHARSET_GROUP_NAME).equalsIgnoreCase("UTF-8")) {
-                return new StringBuilder(matcher.group(CHARSET_GROUP_NAME))
+                final String language = matcher.group(LANG_GROUP_NAME);
+                return new StringBuilder(charset)
                         .append("'")
-                        .append(matcher.group(LANG_GROUP_NAME) == null ? "" : matcher.group(LANG_GROUP_NAME))
+                        .append(language == null ? "" : language)
                         .append("'")
                         .append(encodeToUriFormat(fileNameValueChars))
                         .toString();
             } else {
-                throw new ParseException(matcher.group(CHARSET_GROUP_NAME) + " charset is not supported", 0);
+                throw new ParseException(charset + " charset is not supported", 0);
             }
         }
 
--- a/media/multipart/src/main/java/org/glassfish/jersey/media/multipart/ContentDisposition.java
+++ b/media/multipart/src/main/java/org/glassfish/jersey/media/multipart/ContentDisposition.java
@@ -42,7 +42,8 @@ public class ContentDisposition {
     private Date readDate;
     private long size;
     private static final Pattern FILENAME_EXT_VALUE_PATTERN =
-            Pattern.compile("(UTF-8|ISO-8859-1)'([a-z]{2,8}(-[a-z0-9]+)?)?'(%[a-f0-9]{2}|[a-z0-9!#$&+.^_`|~-])+");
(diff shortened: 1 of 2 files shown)


media/multipart/src/main/java/org/glassfish/jersey/media/multipart/ContentDisposition.java
```diff
@@ -40,6 +41,8 @@
     private Date modificationDate;
     private Date readDate;
     private long size;
+    private static final Pattern FILENAME_EXT_VALUE_PATTERN =
+            Pattern.compile("(UTF-8|ISO-8859-1)'([a-z]{2,8}(-[a-z0-9]+)?)?'(%[a-f0-9]{2}|[a-z0-9!#$&+.^_`|~-])+");
```
jbescos, line 45: I am not expert on RFC 5987 but for example this (ISO-8859-1 in lower case) will not match:
`iso-8859-1'language-us'abc%a1abc%a2%b1!#$&+.^_`|~-`

It would be good to pick many examples of valid strings for testing.

senivam: as per RFC, it's only possible to send capitalized encoding name, so this should be fine. However the purpose of the PR should be to allow multi-language (non ASCII) file names. And, for example, this pattern ``fileNameExt = "UTF-8''українська_назва.pdf";`` does not work. Same applies for the pattern: ``fileNameExt = "UTF-8''nombre_español.pdf";``. And what is the most disapointing here, it does not allow capital letters (even ASCII) in the name. 
@aserkes, could you please fix this?

aserkes: RFC 5987 does not allow any symbols except in ASCII encoding. Symbols in any other encoding have to be encoded as described in https://tools.ietf.org/html/rfc3986#section-2.1

senivam: OK, so the pattern for multi-language file name shall look like ``fileNameExt = "UTF-8''nombre_espa%c3%b1ol.pdf"`` 
And what about language-tag? The RFC 5987 references [Section 2.3 of [RFC2978]](https://tools.ietf.org/html/rfc5646#section-2.1) but I'm not sure it can include the word ``language`` itself. It might be that the whole ``[ language ]`` construct in the ``ext-value`` description shall be substituted by language code from RFC 2978. Or do I understand this wrong?

tests/e2e/src/test/java/org/glassfish/jersey/tests/api/ContentDispositionTest.java
```diff
@@ -97,6 +97,136 @@ public void testToString() {
...
+            fileNameExt = "ISO-8859-1'language-us'李.txt";
+            header = prefixHeader + fileNameExt + "\"";
+            contentDisposition = new ContentDisposition(HttpHeaderReader.newInstance(header), true);
+            assertEquals(fileName, contentDisposition.getFileName());
+
+            //correct fileNameExt
+            fileNameExt = "ISO-8859-1'language-us'FILEname.tXt";
+            header = prefixHeader + fileNameExt + "\"";
+            contentDisposition = new ContentDisposition(HttpHeaderReader.newInstance(header), true);
+            assertEquals(fileNameExt, contentDisposition.getFileName());
```
senivam, line 223: regarding those tests - it would be good to extract duplicated code into some method and call it only with those parameters which differs.

media/multipart/src/main/java/org/glassfish/jersey/media/multipart/ContentDisposition.java
aserkes: Language-Tag can exist or can be omitted. Requirements to this tag are described in https://tools.ietf.org/html/rfc5646#section-2.1 . This tag can have various content that is not always standardized and can expand, but main pattern for it looks like `([a-z]{2,8}(-[a-z0-9]+)?)?` (if I understand RFC 5646 correctly).

senivam: the pattern seems to be relevant, 

and another thing comes to mind - it might be good to extend existing constructor (along with related builders) with the fileNameExt parameter. Just not to take fileNameExt from header and to have the whole ContentDisposition class consistent. 

@jansupol what do you think?

tests/e2e/src/test/java/org/glassfish/jersey/tests/api/ContentDispositionTest.java
aserkes: fixed

media/multipart/src/main/java/org/glassfish/jersey/media/multipart/ContentDisposition.java
jansupol: The RFC 5987 describes:
1. The character set the recipient is about to support. Mandatory is UTF-8, ISO-88592-1 is optional as per RFC 8187, and other charsets are possible. However, Jersey does not support the charsets per se, the PR only checks the EBNF. From this point of view, I am not sure it makes sense to prohibit other charsets.
2. Jersey encodes data such as URL, query params,... for the users. This PR does not encode anything, however. Since the encoding is left on the user, I am not sure whether the code should restrict `filename*` values for characters that do not match the regular code. The encoding is left on the users. Should they use a non-encoded symbol it is they who would deal with it at the end. The corresponding bug specifically asks for being able to add non-encoded symbols.

For these two reasons, should we just have:
```
   private void createParameters() throws ParseException {
        fileName = parameters.get("filename");

        String fileNameExt = parameters.get("filename*");
        if (fileNameExt != null) {
            fileName = fileNameExt;
        }
        ....
```

jansupol: The ideal behaviour would be:
1) Check whether the file name is encoded. If not, check the char set. Only UTF-8 needs to be supported per RFC 8187, and only UTF-8 is supported by Jersey internal econding utility `UriComponent#encode`. If the not encoded and not UTF-8, throw exception. If UTF-8, and non encoded, encode.
2) If encoded, we do not care about the char set. UTF-8, ISO-8859-2, or any other is fine. Just check for valid characters, if not valid, throw exception. If valid check language.
3) The language should be checked by the regular expression you have in PR. If not valid, throw exception.

aserkes: Made some changes :
1) Added encoding for a filename* parameter if it is valid, not encoded, and its charset is UTF-8.  
2) Jersey will throw an exception if a filename* parameter is not valid. 
At the same time I left only 2 possible charset for a filename* parameter (ISO-8859-1 and UTF-8) because as mentioned in RFC 5987 :

> Producers MUST use either the "UTF-8" ([RFC3629]) or the "ISO-8859-1"
   ([ISO-8859-1]) character set.

media/multipart/src/main/java/org/glassfish/jersey/media/multipart/ContentDisposition.java
```diff
@@ -192,6 +205,42 @@ private void createParameters() throws ParseException {
...
+
+        final String fileNameExt = parameters.get("filename*");
+        if (fileNameExt == null) {
+            return fileName;
+        }
+
+        final Matcher matcher = FILENAME_EXT_VALUE_PATTERN.matcher(fileNameExt);
+        if (matcher.matches()) {
+            if (isEncodedInUriFormat(fileNameExt)) {
+                return fileNameExt;
```
jansupol, line 219: This never happens in your tests, since `isEncodedInUriFormat(fileNameExt)` always fail on `'`. Perhaps you meant `if (isEncodedInUriFormat(matcher.group(FILENAME_GROUP_NAME)))`?

But then again, why you do `if (matcher.matches())` ? If `isEncodedInUriFormat(matcher.group(FILENAME_GROUP_NAME))` is true, we likely should not care about the charset. You are right, the obsolete RFC 5987 requires UTF-8 or ISO-8859-1, but then new RFC 8187 changes the charset to UTF-8, but then next month another RFC supporting UTF-16 can arise - do you see a reason for not supporting it at the moment when it is already encoded by the user?

jansupol (changes requested):

media/multipart/src/main/java/org/glassfish/jersey/media/multipart/ContentDisposition.java
aserkes: You are right about `if (isEncodedInUriFormat(matcher.group(FILENAME_GROUP_NAME)))` and I fixed it. 

`if (matcher.matches())` uses to check if filename* parameter has correct structure and also it uses to check language value in this parameter.

I agree with you that it is not necessary to require only UTF-8 or ISO-8859-1 charset if filename is encoded, so I made necessary changes.

media/multipart/src/main/java/org/glassfish/jersey/media/multipart/ContentDisposition.java
```diff
@@ -192,6 +207,47 @@ private void createParameters() throws ParseException {
...
+
+            final String fileNameValueChars = matcher.group(FILENAME_GROUP_NAME);
+            if (isFilenameValueCharsEncoded(fileNameValueChars)) {
+                return fileNameExt;
+            }
+
+            if (matcher.group(CHARSET_GROUP_NAME).equalsIgnoreCase("UTF-8")) {
+                return new StringBuilder(matcher.group(CHARSET_GROUP_NAME))
+                        .append("'")
+                        .append(matcher.group(LANG_GROUP_NAME) == null ? "" : matcher.group(LANG_GROUP_NAME))
```
jansupol, line 231: Regexps are possibly not the fastest piece of code. Can you store the result of `matcher.group(LANG_GROUP_NAME)` and reuse?

aserkes: Done.

jansupol (approved):

senivam:
I would advise to bump copyright year to 2021 since those files where modified in the current year already.

senivam (approved):

PR #4650, opened 2020-12-02
Enable copyright-check and update copyrights

jbescos:
Currently the build doesn't fail when copyright fails because we need to execute this one: glassfish-copyright:check

There are many dates to update.
- Jersey-bot updates pom files version, but the date was not updated. When jersey-bot runs it should invoke 'glassfish-copyright:repair'
- Many files were updated in this commit 'Move CDI integration tests to a common CDI-Integration module (#4280)' but the date was not updated.

### DIFF
--- a/connectors/jetty-connector/src/test/java/org/glassfish/jersey/jetty/connector/ProxyTest.java
+++ b/connectors/jetty-connector/src/test/java/org/glassfish/jersey/jetty/connector/ProxyTest.java
@@ -1,4 +1,5 @@
 /*
+ * Copyright (c) 2020 Oracle and/or its affiliates. All rights reserved.
  * Copyright (c) 2019 Banco do Brasil S/A. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
@@ -12,9 +13,8 @@
  * https://www.gnu.org/software/classpath/license.html.
  *
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
- * Contributors:
- *    Marcelo Rubim
  */
+
 package org.glassfish.jersey.jetty.connector;
 
 import org.eclipse.jetty.server.Request;
--- a/examples/entity-filtering-selectable/src/main/java/org/glassfish/jersey/examples/entityfiltering/selectable/App.java
+++ b/examples/entity-filtering-selectable/src/main/java/org/glassfish/jersey/examples/entityfiltering/selectable/App.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2014, 2019Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2014, 2020 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Distribution License v. 1.0, which is available at
--- a/pom.xml
+++ b/pom.xml
@@ -642,7 +642,6 @@
                 <plugin>
                     <groupId>org.glassfish.copyright</groupId>
                     <artifactId>glassfish-copyright-maven-plugin</artifactId>
-                    <version>2.0</version>
                     <configuration>
                         <excludeFile>etc/config/copyright-exclude</excludeFile>
                         <!--svn|mercurial|git - defaults to svn-->
@@ -652,7 +651,7 @@
                         <!-- skip files not under SCM-->
                         <scmOnly>true</scmOnly>
                         <!-- turn off warnings -->
-                        <warn>true</warn>
+                        <warn>false</warn>
                         <!-- for use with repair -->
                         <update>false</update>
                         <!-- check that year is correct -->
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,9 +19,12 @@ env:
   matrix:
     - TEST_SET="-Ptravis_e2e_skip"
     - TEST_SET="-Ptravis_e2e"
-    - TEST_SET="glassfish-copyright:copyright"
+    - TEST_SET="glassfish-copyright:check"
 
 install: true
 
 script:
   - bash etc/travis/travis.sh $TEST_SET
+
+git:
+  depth: false
--- a/archetypes/jersey-example-java8-webapp/pom.xml
+++ b/archetypes/jersey-example-java8-webapp/pom.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
-    Copyright (c) 2015, 2019 Oracle and/or its affiliates. All rights reserved.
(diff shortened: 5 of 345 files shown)


senivam:
actually I do not agree with automatic build failure (that is why it was copyright:copyright). Typical use case is when new Jersey release is published and all pom.xml files are modified and they are being modified with jersey-bot which belongs to Eclipse and has nothing to do with Oracle. Any other contribution from community also does not necessary shall update Oracle's copyright. Payara for instance has its own copyright.
 
Basically the whole community shall not bother with Oracle's copyright, except people from Oracle :)

jansupol:
This is NOT Oracle copyright. This is Oracle and affiliates' copyright. If I understand it correctly, any contributor is understood as an affiliate.

Regarding the release, I believe it is possible to update the copyright year in poms by a sed swarm, is it not?

connectors/jetty-connector/src/test/java/org/glassfish/jersey/jetty/connector/ProxyTest.java
```diff
@@ -1,4 +1,5 @@
 /*
+ * Copyright (c) 2020 Oracle and/or its affiliates. All rights reserved.
```
jansupol, line 2: There seems to be something wrong. This file was not modified in 2020 and yet the copyright plugin requires a copyright change.

jbescos: It is because the plugin removed the section, and this is considered as a new change.
 * Contributors:
 *    Marcelo Rubim

examples/entity-filtering-selectable/src/main/java/org/glassfish/jersey/examples/entityfiltering/selectable/App.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2014, 2019Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2020 Oracle and/or its affiliates. All rights reserved.
```
jbescos, line 2: This seems wrong, I will fix it.

senivam:
> This is NOT Oracle copyright. This is Oracle and affiliates' copyright. If I understand it correctly, any contributor is understood as an affiliate.

As I remember, this project was contributed to Eclipse, since then any contributor is understood as Eclipse's affiliate. That is why ECA sign is required for any IP contribution.

If it would be Helidon project, then yes, it belongs to Oracle, but this project belongs to Eclipse.

m0mus:
Just want to confirm the copyright should be in this format:
```
Copyright (c) 2002, 2020 Oracle and/or its affiliates. All rights reserved.
```
where `2002` is the year when this file was originally created and `2020` the year of last change.
If the file is new it's just
```
Copyright (c) 2020 Oracle and/or its affiliates. All rights reserved.
```

senivam:
OK, after all discussions I have to agree with build failure in Travis build due to copyright year changes.

senivam (approved):

jbescos:
For some reason in Travis it fails in the first module, but there is no information about why it fails. Locally it is working.

I need to push some commits to print more information, but the problem is that Travis seems to be running nightly. It is going to take some time to find why is it failing in Travis.

media/json-jackson/src/main/resources/META-INF/NOTICE.markdown
```diff
@@ -1,4 +1,20 @@
-# Notice for Jersey Json Jackson module 
+#
+# Copyright (c) 2020 Oracle and/or its affiliates. All rights reserved.
```
senivam, line 2: this is kinda .md file (only with .markdown extension), shall be excluded from Copyright check.

pom.xml
```diff
@@ -642,17 +642,16 @@
...
                     <groupId>org.glassfish.copyright</groupId>
                     <artifactId>glassfish-copyright-maven-plugin</artifactId>
-                    <version>2.0</version>
                     <configuration>
                         <excludeFile>etc/config/copyright-exclude</excludeFile>
                         <!--svn|mercurial|git - defaults to svn-->
                         <scm>git</scm>
                         <!-- turn on/off debugging -->
-                        <debug>false</debug>
+                        <debug>true</debug>
```
senivam, line 650: why in debug mode?

senivam (changes requested):

pom.xml
jbescos: I have added it temporarily to know why is it failing in Travis. Currently it fails but there is no information of the error.

media/json-jackson/src/main/resources/META-INF/NOTICE.markdown
jbescos: Done, it is also added in the .markdown in the jersey/etc/config/copyright-exclude

pom.xml
senivam: for now build did not even started since the last change :) Travis seems to be really slow recently.
Regarding additional info - it might not show it because of reduced output in Travis. We had to reduce it due to Travis' limitations, otherwise all build threads would fail. May be you can temporary try to run it with full log output (just for copyright-check thread) and this will show what actually happened there?

senivam: ah, no, copyright already runs with full output, so may be -e key will help when build finally starts :)

jbescos: I tried it yesterday with -e already but the stacktrace doesn't help, you can see it here:
https://travis-ci.org/github/eclipse-ee4j/jersey/jobs/747370995

Hopefully debug=true will help.

Locally there are no problems. I don't really understand why in Travis it works different.

senivam: the issue with no output on failures seems to be caused by this hard-coded logic inside the plugin:
https://github.com/eclipse-ee4j/glassfish-copyright-plugin/blob/94c7ba4c6d3151b57f92cf030f77661f1ff15d17/src/main/java/org/glassfish/copyright/CheckCopyrightMojo.java#L33
It's being always quiet in other words. And when you remove this hard-coded line (default is false), you can see why build fails. However fix requires new PR and then new plugin version shall be released...

senivam (commented):
current content looks good to me, however due to discovered issue with glassfish-copyright-plugin it's required to fix (or provide some workaround) for error output, and this will introduce some changes in the current PR (like plugin version update or something), so, it shall be merged only after reliable solution for error output inside glassfish-copyright-plugin is provided (and this PR is adjusted accordingly).

senivam:
@jbescos could you please update plugin's version to 2.4?

jbescos:
@senivam finally it is working in Travis

senivam (approved):
it looks like all passes OK, settings for copyright plugin and related build configuration are adjusted correctly, changes can be merged.

jbescos:
There are some issues with the rebase of my commits, because my commits belongs to 2020 and new files have been modified in 2021. The plugin is not dealing well with that.

PR #4668, opened 2021-01-05
Jakartified heroku archetype

senivam:


jansupol:
We should likely update the [pom file](https://github.com/eclipse-ee4j/jersey/blob/3.x/archetypes/jersey-heroku-webapp/src/main/resources/archetype-resources/pom.xml), too.

jansupol (approved):

PR #4669, opened 2021-01-05
Add Jaxb properties to External Properties and tests

tvallin:

### DIFF
--- a/tests/integration/externalproperties/pom.xml
+++ b/tests/integration/externalproperties/pom.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
-    Copyright (c) 2020 Oracle and/or its affiliates. All rights reserved.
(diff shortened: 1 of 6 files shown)


tests/integration/externalproperties/pom.xml
```diff
@@ -42,6 +42,10 @@
             <artifactId>jersey-test-framework-provider-jetty</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.sun.xml.bind</groupId>
+            <artifactId>jaxb-impl</artifactId>
```
jbescos, line 47: I think you need this dependency only for test, right?. Then it should contain the test scope.

jbescos: Actually I am wrong. This is already in a test folder, so ignore this comment.

jbescos (approved):

PR #4674, opened 2021-01-06
Merge of the current master into jakartified branch

senivam:


media/json-jackson/pom.xml
```diff
@@ -77,7 +77,12 @@
                 <configuration>
                     <instructions>
                         <Export-Package>org.glassfish.jersey.jackson.*</Export-Package>
-                        <Import-Package>${jakarta.annotation.osgi.version},*</Import-Package>
```
jansupol, line 80: So this should not be in the PR.

jansupol (changes requested):

media/json-jackson/pom.xml
jansupol: Ah, sorry it is commented out

jansupol (approved):

PR #4675, opened 2021-01-07
Make logger static into Hk2RequestScope.Instance class

### LINKED ISSUE #4665: Hk2RequestScope.Instance logger is not static
It appears that `Hk2RequestScope.Instance#logger` is accidentally created once per instance instead of being declared static, which means it calls into the logging subsystem to get the logger instance on each request. The line of code in question can be found here: https://github.com/eclipse-ee4j/jersey/blob/eafb9bdcb82dfa3fd76dd957d307b99d4a22c87f/inject/hk2/src/main/java/org/glassfish/jersey/inject/hk2/Hk2RequestScope.java#L45

tvallin:
Fix #4665 


jansupol (approved):

senivam (approved):

pettyjamesm:
Thanks for the fix!

Issue #4646, opened 2020-11-30
Jetty connector client response buffer is hard limited to 2MB


cen1:
Jetty http client uses 2MB fixed/max buffer size when receiving responses and there is no way to override this in jersey currently. The problem is that this is not even configurable in jetty HttpClient, you need to do it by instantiating a response listener at request time, so supporting this requires change in connector code.

One proposal is this patch by introducing a new jetty client property to override the max buffer size and creating a new listener per request (I only did the synchronous part to test):
https://github.com/cen1/jersey/commit/680d63b86aea6d193504fe36692b5de391ffc62e

Second possible approach is what I observed in camel https://github.com/apache/camel/pull/1388/commits/41cd808b813d0919dea28c31851c1eb056310e82 where an InputStreamListener is used instead to read arbitrary big responses. However, this opens a whole new set of problems at huge payloads that do not fit in ram (there is a comment on the merge about this which I am not sure was addressed). So perhaps this is not the better idea.

Perhaps an even better (best?) idea would be if jetty exposed this as part of HttpClient configuration, this way there is no code change needed in jersey and this can be solved when building jax-rs client instead of request time inside the connector. I am leaning towards this one but I would seek some feedback first before going upstream to jetty project.

jansupol:
The third option seems best for Jersey, since any changes in the buffering mechanics would be handled by Jetty itself. 
I would not mind option 1 or 2, but only as an alternative code branch, for the case when the new property is set. Should it not be set, the existing code would better be executed. That would prevent any possibles issues with the listeners for the existing Jersey users.

cen1:
Thank you for the response, I will check with the jetty project and come back once anything is determined.

cen1:
Where I'm at

1. Jetty not really too happy to allow overriding this hardcoded default since it is per-request and only used in default synchronous method. User already has full freedom to provide their own listeners.

2. One interesting idea was brought up to just allow user to provide their own supplier for `Response.Listener` but since all methods are consumers there is nowhere to write buffer to, so it is not really suppliable. A wrapper would be needed.
In addition, you would need two suppliers, one for sync and for async path since you can't cover both in single listener implementation. I tried this but abandoned the idea.

3. The async path currently has no problem with size since onData is called repeatedly and final buffer built that way. My guess is that memory is the only limit.


[...]

PR #4677, opened 2021-01-08
Jetty synchronous max buffer size property

cen1:
Adds a new jetty specific client property to set the max response buffer size. If property is not set, the old code path is used and exception is thrown for large responses (bigger than 2MiB).

This addresses the immediate paint point and is least invasive. Exception is still thrown if response size is too big even for the custom size. Since this behaviour differs from async path it might be a good idea to just
replace it completely in some major release.


cen1:
Fixes https://github.com/eclipse-ee4j/jersey/issues/4646

jansupol:
I like the PR. Would you mind updating the copyright year to 2021 in the modified files?

jansupol (approved):

senivam (approved):

PR #4686, opened 2021-01-16
Add two missing configuration for Apache - DefaultAuthSchemeRegistry …

nddipiazza:
Add two missing configuration for Apache HTTP client properties:

```
builder.setDefaultAuthSchemeRegistry
builder.setRedirectStrategy
```

These are blockers for several use cases. 

In particular, without the ability to update the default auth scheme registry, you cannot do Spnego AuthScheme.

### DIFF
--- a/connectors/apache-connector/src/main/java/org/glassfish/jersey/apache/connector/ApacheClientProperties.java
+++ b/connectors/apache-connector/src/main/java/org/glassfish/jersey/apache/connector/ApacheClientProperties.java
@@ -156,6 +156,19 @@ public final class ApacheClientProperties {
      */
     public static final String KEEPALIVE_STRATEGY = "jersey.config.apache.client.keepAliveStrategy";
 
+    /**
+     * RedirectStrategy which can customize how redirects are handled.
+     * <p/>
+     * The value MUST be a subclass of {@link RedirectStrategy}.
+     */
+    public static final String REDIRECT_STRATEGY = "jersey.config.apache.client.redirectStrategy";
+
+    /**
+     * DefaultAuthSchemeRegistry which can customize the auth scheme used.
+     * <p/>
+     * The value MUST be a subclass of {@link org.apache.http.config.Registry}.
+     */
+    public static final String DEFAULT_AUTH_SCHEME_REGISTRY = "jersey.config.apache.client.defaultAuthSchemeRegistry";
 
     /**
      * Strategy that closes the Apache Connection. Accepts an instance of {@link ApacheConnectionClosingStrategy}.
--- a/connectors/apache-connector/src/main/java/org/glassfish/jersey/apache/connector/ApacheConnector.java
+++ b/connectors/apache-connector/src/main/java/org/glassfish/jersey/apache/connector/ApacheConnector.java
@@ -62,6 +62,7 @@ import org.apache.http.ConnectionReuseStrategy;
 import org.apache.http.Header;
 import org.apache.http.HttpEntity;
 import org.apache.http.HttpHost;
+import org.apache.http.auth.AuthSchemeProvider;
 import org.apache.http.auth.AuthScope;
 import org.apache.http.auth.UsernamePasswordCredentials;
 import org.apache.http.client.AuthCache;
@@ -69,6 +70,7 @@ import org.apache.http.client.CookieStore;
 import org.apache.http.client.CredentialsProvider;
 import org.apache.http.client.HttpClient;
 import org.apache.http.client.HttpRequestRetryHandler;
+import org.apache.http.client.RedirectStrategy;
 import org.apache.http.client.config.CookieSpecs;
 import org.apache.http.client.config.RequestConfig;
 import org.apache.http.client.methods.CloseableHttpResponse;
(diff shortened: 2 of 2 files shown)


senivam:
There is a valid ECA on file for nicholas.dipiazza@lucidworks.com. But check fails for some reason.

connectors/apache-connector/src/main/java/org/glassfish/jersey/apache/connector/ApacheConnector.java
```diff
@@ -249,6 +251,34 @@
             }
         }
 
+        Object redirectStrategy = config.getProperties().get(ApacheClientProperties.REDIRECT_STRATEGY);
+        if (redirectStrategy != null) {
+            if (!(redirectStrategy instanceof RedirectStrategy)) {
```
senivam, line 256: instanceof is null safe operation, so I would advise to eliminate NULL check here (even though in original connector there are such a redundant checks). 

same is valid for ``defaultAuthSchemeRegistry`` NULL check

connectors/apache-connector/src/main/java/org/glassfish/jersey/apache/connector/ApacheConnector.java
```diff
@@ -317,6 +347,14 @@
             requestConfig = requestConfigBuilder.build();
         }
 
+        if (redirectStrategy != null) {
+            clientBuilder.setRedirectStrategy((RedirectStrategy) redirectStrategy);
+        }
+
+        if (defaultAuthSchemeRegistry != null) {
+            clientBuilder.setDefaultAuthSchemeRegistry((Registry<AuthSchemeProvider>) defaultAuthSchemeRegistry);
```
senivam, line 355: shoudn't it be more general ``Lookup<I>`` instance?

connectors/apache-connector/src/main/java/org/glassfish/jersey/apache/connector/ApacheConnector.java
```diff
@@ -249,6 +251,34 @@
...
+                        redirectStrategy.getClass().getName(),
+                        RedirectStrategy.class.getName())
+                );
+                redirectStrategy = null;
+            }
+        }
+
+        Object defaultAuthSchemeRegistry = config.getProperties().get(ApacheClientProperties.DEFAULT_AUTH_SCHEME_REGISTRY);
+        if (defaultAuthSchemeRegistry != null) {
+            if (!(defaultAuthSchemeRegistry instanceof Registry)) {
```
senivam, line 270: ``Registry`` is a class which inherits ``Lookup<I>`` interface, may be it's better to check for an instance of ``Lookup<I>``?

connectors/apache-connector/src/main/java/org/glassfish/jersey/apache/connector/ApacheClientProperties.java
```diff
@@ -156,6 +156,19 @@
...
+     * The value MUST be a subclass of {@link RedirectStrategy}.
+     */
+    public static final String REDIRECT_STRATEGY = "jersey.config.apache.client.redirectStrategy";
+
+    /**
+     * DefaultAuthSchemeRegistry which can customize the auth scheme used.
+     * <p/>
+     * The value MUST be a subclass of {@link org.apache.http.config.Registry}.
+     */
+    public static final String DEFAULT_AUTH_SCHEME_REGISTRY = "jersey.config.apache.client.defaultAuthSchemeRegistry";
```
senivam, line 171: I would advice to rename the ``DEFAULT_AUTH_SCHEME_REGISTRY`` to ``AUTH_SCHEME_REGISTRY`` which later becomes default for Apache connector (by set default..() method in builder). However for us default means that already there is some predefined registry which can be set automatically if there is nothing else. However in this case there is no such a Registry predefined.

connectors/apache-connector/src/main/java/org/glassfish/jersey/apache/connector/ApacheConnector.java
```diff
@@ -249,6 +251,34 @@
...
+                    LocalizationMessages.IGNORING_VALUE_OF_PROPERTY(
+                        ApacheClientProperties.REDIRECT_STRATEGY,
+                        redirectStrategy.getClass().getName(),
+                        RedirectStrategy.class.getName())
+                );
+                redirectStrategy = null;
+            }
+        }
+
+        Object defaultAuthSchemeRegistry = config.getProperties().get(ApacheClientProperties.DEFAULT_AUTH_SCHEME_REGISTRY);
```
senivam, line 268: can this variable be named without ``default`` word? For reasons described in previous comment.

senivam (changes requested):
Could you provide some tests for those changes?

And when editing those files, could you please modify Copyright year on top of each file to be 2021 (instead 2020 for example)? The 2nd CP year shell be modified.

nddipiazza:
@senivam I am trying to add a unit test, but the master branch will not compile. Is there some quick contributing eclipse/intellij setup guide so i can get a workspace set up?

senivam:
@nddipiazza if you previously compiled the branch (using JDK 8 for example) and now it does not pass again you can try to remove all target/* folders and compile it again using same JDK. If this does not help it's possible to clean local .m2/repository/org/glassfish/jersey folder and build jersey again. 

BTW, we have another approach of supplying parameters which are not explicitly indicated in properties to Apache connector which is manual implementing of ApacheHttpClientBuilderConfigurator SPI interface and passing all required properties via custom ``HttpClientBuilder configure(HttpClientBuilder httpClientBuilder);`` method. Please consider which approach suits better you needs - direct properties config of SPI interface extension.

nddipiazza:
Yep the .m2 repositories thing fixed it. Weird.

Issue #4152, opened 2019-06-10
Revisit the Monitoring feature


jansupol:
- Revisit the behaviour in cases where OutOfMemoryError is thrown.
- Think of optional strategies to limit the memory data consumed by the feature

jbescos:
There are many residual AggregatedValueObject instances that are not trim. These objects are holded by AggregatedSlidingWindowTimeReservoir.

Adding some log lines it is possible to see the issue. In the lines we can see 8 instances of AggregatedSlidingWindowTimeReservoir.

Something is not working properly because there are PUTs of AggregatedValueObject are insterted in many AggregatedSlidingWindowTimeReservoir, but it only trims in one. This causes residual AggregatedValueObject instances that can be seen when you run a heapdump.
```

New instance of org.glassfish.jersey.server.internal.monitoring.AggregatedSlidingWindowTimeReservoir@600e965c
New instance of org.glassfish.jersey.server.internal.monitoring.AggregatedSlidingWindowTimeReservoir@35a4060f
New instance of org.glassfish.jersey.server.internal.monitoring.AggregatedSlidingWindowTimeReservoir@5d49099
New instance of org.glassfish.jersey.server.internal.monitoring.AggregatedSlidingWindowTimeReservoir@6f554bf8
New instance of org.glassfish.jersey.server.internal.monitoring.AggregatedSlidingWindowTimeReservoir@5cf23ef0
New instance of org.glassfish.jersey.server.internal.monitoring.AggregatedSlidingWindowTimeReservoir@4973df1f
New instance of org.glassfish.jersey.server.internal.monitoring.AggregatedSlidingWindowTimeReservoir@6802def0
New instance of org.glassfish.jersey.server.internal.monitoring.AggregatedSlidingWindowTimeReservoir@2d4d5283
PUT: 6799281429504000000. Size: 1 in org.glassfish.jersey.server.internal.monitoring.AggregatedSlidingWindowTimeReservoir@35a4060f
PUT: 6799281429504000001. Size: 2 in org.glassfish.jersey.server.internal.monitoring.AggregatedSlidingWindowTimeReservoir@35a4060f
PUT: 6799281429504000000. Size: 1 in org.glassfish.jersey.server.internal.monitoring.AggregatedSlidingWindowTimeReservoir@5d49099
PUT: 6799281429504000000. Size: 1 in org.glassfish.jersey.server.internal.monitoring.AggregatedSlidingWindowTimeReservoir@6f554bf8
[...]

PR #4697, opened 2021-01-22
Stop filling monitoring queues when processor fails

jbescos:
Related to this https://github.com/eclipse-ee4j/jersey/issues/4152 but there are more things to investigate.

This PR is about not produce new events in the queues when there is no consumer.


tests/e2e-server/src/test/java/org/glassfish/jersey/tests/e2e/server/monitoring/MonitoringEventListenerTest.java
```diff
@@ -0,0 +1,178 @@
...
+        return resourceConfig;
+    }
+
+    @Test
+    public void exceptionInScheduler() throws Exception {
+        final Long ERRORS_BEFORE_FAIL = 10L;
+        // Send some requests to process some statistics.
+        request(ERRORS_BEFORE_FAIL);
+        // Give some time to the scheduler to collect data.
+        Thread.sleep(TIMEOUT);
```
jansupol, line 130: Can CountdownLatch be used instead of sleep? Sleep tends to create intermittent failures.

jansupol (changes requested):

tests/e2e-server/src/test/java/org/glassfish/jersey/tests/e2e/server/monitoring/MonitoringEventListenerTest.java
jbescos: In the first Thread.sleep we can do it because it is about waiting till something happens.

But in the second we cannot do it, because nothing should happen. So we need to wait explicitly and check.

jbescos: I was trying to add a CountDownLatch for the first one but it is not easy. There are 2 ways:
Easy one. It is about checking the queue till the size is 0:
```
    private void waitTillQueueIsEmpty() throws InterruptedException {
        final CountDownLatch latch = new CountDownLatch(1);
        ScheduledExecutorService scheduler = Executors.newSingleThreadScheduledExecutor();
        scheduler.scheduleAtFixedRate(() -> {
            Response response = target("/example/queueSize").request().get();
            int queueSize = response.readEntity(Integer.class);
            if (queueSize == 0) {
                latch.countDown();
            }
        }, 0, 10, TimeUnit.MILLISECONDS);
        boolean wait = latch.await(TIMEOUT, TimeUnit.MILLISECONDS);
        scheduler.shutdown();
        assertTrue("The queue was not empty in the given time", wait);
    } 
```
But this doesn't work always, because it could happen the scheduler consumed all the events and the size is 0, but new events are coming.

The difficult option:
Pass the CountDownLatch reference to the Runnable in the Scheduler of MonitoringStatisticsProcessor, so it will count it down every time an event is consumed. This is tricky because the way it is done it is not possible. I would need to make some changes in MonitoringStatisticsProcessor to achieve this.

I think it is not necessary to over complicate it, because the scheduler is running every 1 millisecond and it is very unlikely that during 1 second it was not able to do the work. But let me know what you think.

jansupol (approved):

senivam (approved):

Issue #4654, opened 2020-12-09
MicroProfile Rest Client 2.0 support


jGauravGupta:
Features in MicroProfile Rest Client 2.0

- Defined that CDI-managed providers should be used instead of creating a new instance, if applicable.
- Support different configurations for collections used in query parameters.
- Added proxy server configuration support.
- Added configuration for automatically following redirect responses.
- Added support for JSON-B configuration via `ContextResolver<Jsonb>`.
- Added support for Server-Sent Events.
- Changed dependency scope for most dependencies to `provided`.
- Update to use Jakarta EE 8 dependencies.

jGauravGupta:
Created ticket to avoid duplication of efforts.

PR #4699, opened 2021-01-25
MicroProfile Rest Client 2.0 support

jGauravGupta:
Fixes Github issue: https://github.com/eclipse-ee4j/jersey/issues/4654

**MP Rest Client 2.0 - QueryParamStyle support:**
`JerseyClient`, `JerseyClientBuilder`, and `JerseyUriBuilder` is extended by the MP Rest Client module to modify the query param based on the `QueryParamStyle`.

**MP Rest Client 2.0 - CDI-managed provider:**
Commit cherry-picked from https://github.com/eclipse-ee4j/jersey/pull/4695


jansupol:
For the PR we need to get the following CQs approved:
```
        <dependency>
            <groupId>io.projectreactor</groupId>
            <artifactId>reactor-core</artifactId>
            <version>3.4.0</version>
        </dependency>
        <dependency>
            <groupId>com.github.tomakehurst</groupId>
            <artifactId>wiremock</artifactId>
            <version>2.24.1</version>
            <scope>test</scope>
        </dependency>
        <dependency>
            <groupId>org.reactivestreams</groupId>
            <artifactId>reactive-streams</artifactId>
            <version>1.0.3</version>
        </dependency>
```
Do we need all of them?

Verdent (changes requested):
In terms of QueryParamStyle, I might suggest to create some equivalent enum directly in Jersey and implement this query param formatting in Jersey itself but based on Jersey QueryParamStyle equivalent. I think @jansupol would be more then fine with taking this approach.

With that said, I also believe that it would be much more suitable to define some normal property and use that to propagate this new Jersey enum value.

jGauravGupta:
Hi @jansupol 

The `org.reactivestreams:reactive-streams` dependency is required to provide [Server-Sent Events](https://github.com/eclipse/microprofile-rest-client/blob/2.0/spec/src/main/asciidoc/server_sent_events.asciidoc) support. Removed the `wiremock` upgrade and unused dependency.

Verdent:
Hi @jGauravGupta ,
what I actually meant by not using field for storing proxy string at all is to assign that proxy value directly to the properties and not doing it in build method. :-) Feel free to give me some feedback if you don't agree.

jGauravGupta:
Hi @Verdent,

If proxy address will be assigned directly to the properties then either host & port validation need to be duplicated in `RestClientProducer`  or skipped as you suggested above to split the value.

> I might suggest to split host and port in RestClientProducer and set it to the method above. By using this approach you will validate the host and port even from config.

jansupol:
Filed CQ [23011](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23011) for reactive streams 1.0.3

Verdent:
@jGauravGupta The splitting is OK to do, what I meant is that you do not have to save proxy parts to the fields and do that property setting directly in that proxy method after validation and not in build method.

jansupol:
So the PR brings two main parts as I see it:

- The QueryParamStyle
- The SSE

Both parts are handled separately from the rest of Jersey so that Jersey without MP Rest Client may not benefit from the functionality and some code duplication takes place.

For the 
- QueryParamStyle: is it possible to make the new URIBuilder part of Jersey, so that JerseyWebTarget can have a new method `setQueryParamStyle(JerseyQueryParamStyle style)` for instance?
- SSE: the SSE is implemented from scratch and does not use the SSE implementation from Jersey. Is there any problem to provide a bridge to the Jersey SSE impl?

jGauravGupta:
Hi @jansupol and @Verdent ,

I have updated the PR with the requested changes.

jansupol:
@jGauravGupta can you please rebase atop master so that the PR does contain just the relevant changes? Thank you

jansupol:
The CQ has been approved.

senivam (approved):

Verdent (approved):
There are some minor problems and absence of tests, but I will fix that in separate PR.

jansupol (approved):
The PR passes the RestClient TCKs and it nicely incorporates JerseyUriBuilder with the new QueryParamStyle. We'll need to provide additional tests for the behaviour of different styles. But that can be done with additional PR. The open question is the SSE which does not provide a client reconnect mechanism required by the JAX-RS Spec. The RestClient TCK test does not expect the reconnect behaviour so we can go with provided SSE for the time being.

Verdent:
Adding link to the rest client TCK issue which @jansupol mentioned: [SSE reconnection](https://github.com/eclipse/microprofile-rest-client/issues/302)

PR #4701, opened 2021-01-26
Copyright check

jbescos:
Copy of https://github.com/eclipse-ee4j/jersey/pull/4650 

There was one problem after the rebase, because my commits from 2020 goes on top of other commits in 2021. The copyright plugin couldn't deal with that.

It should work if I merge instead of rebasing because the order of commits would match the dates order, but I think it is cleaner to create a new PR.

### DIFF
--- a/core-common/src/main/java8/org/glassfish/jersey/internal/jsr166/SubmissionPublisher.java
+++ b/core-common/src/main/java8/org/glassfish/jersey/internal/jsr166/SubmissionPublisher.java
@@ -1,20 +1,4 @@
 /*
- * Copyright (c) 2018, 2020 Oracle and/or its affiliates. All rights reserved.
- *
- * This program and the accompanying materials are made available under the
- * terms of the Eclipse Public License v. 2.0, which is available at
- * http://www.eclipse.org/legal/epl-2.0.
- *
- * This Source Code may also be made available under the following Secondary
- * Licenses when the conditions for such availability set forth in the
- * Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
- * version 2 with the GNU Classpath Exception, which is available at
- * https://www.gnu.org/software/classpath/license.html.
- *
- * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
- */
-
-/*
  * Written by Doug Lea with assistance from members of JCP JSR-166
  * Expert Group and released to the public domain, as explained at
  * http://creativecommons.org/publicdomain/zero/1.0/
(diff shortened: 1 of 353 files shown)


senivam (approved):

core-common/src/main/jsr166/org/glassfish/jersey/internal/jsr166/Flow.java
```diff
@@ -1,3 +1,19 @@
+/*
```
jansupol, line 1: This file has CC0 license. We must not put EPL in this file.

core-common/src/main/java8/org/glassfish/jersey/internal/jsr166/SubmissionPublisher.java
```diff
@@ -1,3 +1,19 @@
+/*
```
jansupol, line 1: This file has CC0 license. We must not put EPL in this file.

jansupol (changes requested):

core-common/src/main/jsr166/org/glassfish/jersey/internal/jsr166/Flow.java
jbescos: core-common/src/main/jsr166/org/glassfish/jersey/internal/jsr166/ should by skipped if I get this right

jbescos: This was excluded already /core-common/src/main/java/org/glassfish/jersey/internal/jsr166
So I have removed the copyright on that file and I added:
/core-common/src/main/jsr166/org/glassfish/jersey/internal/jsr166

core-common/src/main/java8/org/glassfish/jersey/internal/jsr166/SubmissionPublisher.java
jbescos: I have removed the copy right and added this path for exclusion:
/core-common/src/main/java8/org/glassfish/jersey/internal/jsr166

ext/bean-validation/src/main/java/org/glassfish/jersey/server/validation/internal/hibernate/HibernateInjectingConstraintValidatorFactory.java
```diff
@@ -1,3 +1,19 @@
+/*
```
jansupol, line 1: This file has Apache License. We must not add EPL here

ext/bean-validation/src/main/java/org/glassfish/jersey/server/validation/internal/hibernate/DestructibleBeanInstance.java
```diff
@@ -1,3 +1,19 @@
+/*
```
jansupol, line 1: This file has Apache License. We must not add EPL here

jansupol (changes requested):

jansupol (approved):

Issue #4683, opened 2021-01-15
Setting ExecutorService causes connection leak


olotenko:
https://github.com/eclipse-ee4j/jersey/commit/3d04682047c29a636c00494492af7acec3f6f611#r45995045

```
@Path("/jax-rs-books")
@ApplicationScoped
public class JaxRsIndirectBookResource {
   @Inject
   protected Client client;

   @Inject
   @ConfigProperty(name = "book-service/mp-rest/url")
   protected String bookServiceUrl;

   @Inject
   @ConfigProperty(name = "sleep-service/mp-rest/url")
   protected String sleepServiceUrl;

   protected static final GenericType<Book[]> responseType = new GenericType(new Book[0].getClass());

   @GET
   @Path("/sync")
   @Produces(MediaType.APPLICATION_JSON)
   public Response getBooks() {
      Book[] books = client.target(bookServiceUrl + "/books")
                           .request(MediaType.APPLICATION_JSON)
                           .get(responseType);
      return Response.ok(books).build();
   }

   @GET
   @Path("/rx")
   @Produces(MediaType.APPLICATION_JSON)
   public void getBooksRx(@Suspended AsyncResponse response) {
      CompletionStage<Book[]> futureBooks =
         client.target(bookServiceUrl + "/books")
               .request(MediaType.APPLICATION_JSON)
               .rx()
               .get(responseType);
      futureBooks.thenApply(response::resume)
                 .exceptionally(response::resume);
   }
}
```
1. `GET` against `/jax-rs-books/sync`, one at a time, no concurrency needed - pooled connections get reused with all Connectors
2. `GET` against `/jax-rs-books/rx`, one at a time - pooled connections are not reused with Jetty and Netty connectors.

This can be seen to happen due to this (added an Exception that traces the creation point for `NettyConnector`, it is not really thrown, just logging):
```
java.lang.Exception: Eh?...New connector
	at org.glassfish.jersey.netty.connector.NettyConnector.<init>(NettyConnector.java:108)
	at org.glassfish.jersey.netty.connector.NettyConnectorProvider.getConnector(NettyConnectorProvider.java:56)
	at org.glassfish.jersey.client.ClientConfig$State.initRuntime(ClientConfig.java:459)
	at org.glassfish.jersey.internal.util.collection.Values$LazyValueImpl.get(Values.java:317)
	at org.glassfish.jersey.client.ClientConfig.getRuntime(ClientConfig.java:814)
[...]

PR #4705, opened 2021-01-28
Do not create a connector multiple times for each rx() call

jansupol:
Priority is to use @ClientAsyncExecutor annotated ExecutorServiceProvider


jansupol:
The change from `request().getClientConfig().executorService(provided);` to `((ClientConfig) request().getConfiguration()).executorService(provided);` itself causes the Connector is not recreated, as it keeps the inner State.

The other `JerseyInvocation` changes prioritize user provided `@ClientAsyncExecutor` annotated `ExecutorServiceProvider` over another user provided `ExecutorServiceProvider`.

jbescos (approved):

senivam (approved):

jansupol:
Fixes #4683

PR #4707, opened 2021-01-29
Configurable COLLISION_BUFFER_POWER

jbescos:
Relates to https://github.com/eclipse-ee4j/jersey/issues/4152

Running test with 100000 requests and keeping default value (8):
image

Running test with 100000 requests and COLLISION_BUFFER_POWER = 3:
image

### DIFF
--- a/core-server/src/main/java/org/glassfish/jersey/server/internal/monitoring/core/ReservoirConstants.java
+++ b/core-server/src/main/java/org/glassfish/jersey/server/internal/monitoring/core/ReservoirConstants.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015, 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright 2015, 2021 Oracle and/or its affiliates. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,6 +16,12 @@
 
 package org.glassfish.jersey.server.internal.monitoring.core;
 
+import java.security.PrivilegedAction;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+import org.glassfish.jersey.server.ServerProperties;
+
 /**
  * The constants that determine the behaviour of sliding windows and their trimmers.
  *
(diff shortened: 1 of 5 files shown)


core-server/src/test/java/org/glassfish/jersey/server/internal/monitoring/TimeWindowStatisticsImplTest.java
```diff
@@ -18,8 +18,11 @@
 
 import java.util.concurrent.TimeUnit;
 
+import org.glassfish.jersey.server.internal.monitoring.core.ReservoirConstants;
```
senivam, line 21: copyright please :)

senivam (changes requested):

core-server/src/main/java/org/glassfish/jersey/server/internal/monitoring/core/ReservoirConstants.java
```diff
@@ -16,28 +16,59 @@
...
  *
  * @author Stepan Vavra
  */
 public final class ReservoirConstants {
 
+    /**
+     * JVM argument to define the value of {@link COLLISION_BUFFER_POWER}.
+     * Lower values reduce the memory footprint.
+     */
+    public static final String COLLISION_BUFFER_POWER_JVM_ARG = "monitoring.collision.buffer.power";
```
jansupol, line 34: Please move this to ServerProperties and use the usual "jersey.config.server" prefix.

jansupol (changes requested):

core-server/src/main/java/org/glassfish/jersey/server/internal/monitoring/core/ReservoirConstants.java
jbescos: Done

core-server/src/main/java/org/glassfish/jersey/server/internal/monitoring/core/ReservoirConstants.java
```diff
@@ -16,28 +16,55 @@
...
+    static {
+        PrivilegedAction<Integer> action = new PrivilegedAction<Integer>() {
+            @Override
+            public Integer run() {
+                return Integer.getInteger(ServerProperties.COLLISION_BUFFER_POWER_JVM_ARG, DEFAULT_COLLISION_BUFFER_POWER);
+            }
+        };
+        int collisionBuffePower = DEFAULT_COLLISION_BUFFER_POWER;
+        try {
+            collisionBuffePower = action.run();
```
senivam, line 60: actually I'm not sure this will work in favor of ``AccessController.doPrivileged(action);`` because purpose of the AccessController is to provide to privileged action those privileges which previously that action did not have. So, AccessControler shall supply those 
```
 // Needed by TimeWindowStatisticsImplTest
  permission java.util.PropertyPermission "jersey.config.server.monitoring.collision.buffer.power", "read,write";
  ```
  you are providing for test purposes to the privileged action of getting property. While securityManager is on for sure. 
  And running action just by ``run()`` method will just invoke it as usual method without any additional privileges. 
  And SecurityException might not be caught when using  AccessController because in that case user can supply proper privileges to be provided to action and thus SecurityException won't be thrown. On the other hand when you just run the action and exception occurs there is no way for privileges to be supplied to it so it will always fail and application will always use default collision buffer.

senivam (changes requested):
see note in the source code

senivam (approved):
LGTM

jansupol (approved):

Issue #4655, opened 2020-12-12
org.glassfish.jersey.containers:jersey-container-jetty-http:3.0.0 Unusable on JDK11+


birdflyer-lszo:
I am using the org.glassfish.jersey.containers:jersey-container-jetty-http:3.0.0 dependency to setup my Jetty container. While on version 2.x this worked just fine, version 3.0.0 does no longer seem to provide a JettyHttpContainerFactory class.

In the source repository there are multiple source directories for different JDKs, but the build does not seem to build one for JDK 11 onward. The JAR file published on https://repo1.maven.org/maven2/org/glassfish/jersey/containers/jersey-container-jetty-http/3.0.0/ appears to be the one for JDK8. According to the POM, JDK11+ modules are only built when a specific profile is active.

The repository does not seem to include a JDK11+ build of the jersey-container-jetty-http artifact. Is there a different repository that contains the JDK11 artifacts?

jansupol:
The goal was to create a multirelease artifact that works for JDK 8 (well Jetty needs JDK 11, so that version just informs that JDK 11 is required) as well as for JDK 11. It looks like JDK 11 classes were not included during the build :(

senivam:
@birdflyer-lszo the release process of jakartified Jersey includes compilation of modules by different JDKs. Since several modules do not support JDKs prior to 11 we shall provide some stubs for JDK 1.8 which will inform user of in-availability of a particular module for used JDK. And that particular module shall support multi-release jar in order to provide expected functionality for supported JDKs. 

Jetty related modules where adjusted in that way, that is why you can see several profiles inside the pom.xml each of them is activated by different JDK. 

The order of release is: compile module using higher JDK to generate required functionality for supported JDK, then compile the same module using lower JDK to provide not-supported stub and pick up previously generated functionality from higher JDK.  And then bundle all of this into the final JAR and release to central. 

I can admit that there was an error inside releasing job which did not properly generate functionality for 11+ JDKs and so only stubs where included into 3.0.0 release. Thus I can say that Jetty is not supported in that release due to that pity error. Profiles in POM files are however correct and there is no need to adjust them. 

The issue is already fixed and future releases of Jersey 3.x will be published properly. 

I will inform you (in this issue) when next release is available for public usage. 

And thank you for finding this out and reporting that. I apologise for that mistake which in fact had caused total unavailability of Jetty for the release of Jersey 3.0.0.

birdflyer-lszo:
[...]

PR #4709, opened 2021-02-02
Copy JDK 11 Jetty classes to JDK 8 target

senivam:


jansupol (approved):

darmbrust:
Has https://github.com/eclipse-ee4j/jersey/issues/4655#issuecomment-777870120 been resolved with this patch?

PR #4710, opened 2021-02-02
Adjusting Jersey archetypes

senivam:

### DIFF
--- a/archetypes/jersey-quickstart-grizzly2/src/main/resources/archetype-resources/src/main/java/Main.java
+++ b/archetypes/jersey-quickstart-grizzly2/src/main/resources/archetype-resources/src/main/java/Main.java
@@ -37,7 +37,7 @@ public class Main {
     public static void main(String[] args) throws IOException {
         final HttpServer server = startServer();
         System.out.println(String.format("Jersey app started with endpoints available at "
-                + "%s\nHit Ctrl-C to stop it...", BASE_URI));
+                + "%s%nHit Ctrl-C to stop it...", BASE_URI));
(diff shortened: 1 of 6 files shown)


archetypes/jersey-quickstart-grizzly2/src/main/resources/archetype-resources/src/main/java/Main.java
```diff
@@ -36,8 +36,8 @@ public static HttpServer startServer() {
      */
     public static void main(String[] args) throws IOException {
         final HttpServer server = startServer();
-        System.out.println(String.format("Jersey app started with WADL available at "
-                + "%sapplication.wadl\nHit enter to stop it...", BASE_URI));
+        System.out.println(String.format("Jersey app started with endpoints available at "
+                + "%s\nHit Ctrl-C to stop it...", BASE_URI));
```
aimiebell, line 40: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [FS: Format string should use %n rather than n (VA_FORMAT_STRING_USES_NEWLINE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#fs-format-string-should-use-n-rather-than-n-va-format-string-uses-newline).
This format string includes a newline character (\n). In format strings, it is generally  preferable to use %n, which will produce the platform-specific line separator.

jansupol (approved):

Verdent (approved):

PR #4714, opened 2021-02-04
update maven-antrun-plugin to 3.0.0

senivam:


senivam:
update the whole ant run plugin - instead of bumping its ant version dependency at #4713 
Version 3.0.0 includes by default ant of version 1.9.14

jansupol (approved):

Verdent (approved):

Issue #4636, opened 2020-11-10
ParameterConverter throwing IllegalArgumentException MUST result in 400 or 404


mkarg:
When a ParameterConverter is throwing IllegalArgumentException, Jersey currently is silently assuming the default value (or null) as the converter's outcome due to [this codeline](https://github.com/eclipse-ee4j/jersey/blob/c1263e09ad867e88cf0596930d1e81917f997a48/core-server/src/main/java/org/glassfish/jersey/server/internal/inject/SingleValueExtractor.java#L65). This is wrong according to [this JAX-RS discussion](https://github.com/eclipse-ee4j/jaxrs-api/issues/915). Instead Jersey MUST use a 400 or 404 result code.

PR #4737, opened 2021-03-02
ParameterConverter throwing IllegalArgumentException MUST result in 404

jbescos:
Relates to https://github.com/eclipse-ee4j/jersey/issues/4636 and https://github.com/eclipse-ee4j/jaxrs-api/issues/915

### DIFF
--- a/tests/e2e-server/src/test/java/org/glassfish/jersey/tests/e2e/server/IllegalArgumentExceptionTest.java
+++ b/tests/e2e-server/src/test/java/org/glassfish/jersey/tests/e2e/server/IllegalArgumentExceptionTest.java
@@ -1,0 +1,124 @@
+/*
+ * Copyright (c) 2021 Oracle and/or its affiliates. All rights reserved.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0, which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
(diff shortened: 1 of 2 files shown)


tests/e2e-server/src/test/java/org/glassfish/jersey/tests/e2e/server/IllegalArgumentExceptionTest.java
```diff
@@ -0,0 +1,80 @@
...
+    public static class TestResource {
+        @GET
+        @Path("1")
+        public String get1(@QueryParam(PARAM_NAME) CustomObj value) {
+            return "ok";
+        }
+    }
+
+    @Test
+    public void illegalArgumentExceptionWith404() {
```
mkarg, line 55: It might be a good idea to add a second test which asserts that `IllegalArgumentException` results in `400 Bad Request` (not `404 Not Found`) for `@HeaderParam`.

mkarg (commented):
LGTM in general, but the second case (400 Bad Request) should be tested, too, IMHO.

tests/e2e-server/src/test/java/org/glassfish/jersey/tests/e2e/server/IllegalArgumentExceptionTest.java
jbescos: Thanks for the suggestion. It has been added.

jansupol (approved):

senivam (approved):

PR #4740, opened 2021-03-04
GrizzlyHttpServerTest - testing stability with HTTP, HTTPS and HTTP/2

dmatej:
- parallel execution on random free ports
- disabled for jdk8
- possibility to use three different clients
  - jersey-client - but it doesn't support HTTP/2
  - jdk11+ HttpClient - but it is not available in older JDK versions
  - jetty-client - but version compatible with jdk8 uses different classes than version compatible with jdk11
- each test creates a configured server, then creates several clients
  which are sending GET requests for given time.
- all responses must be HTTP 200
- after the test then number of processed requests is written to the STDOUT
- if there will be just one non compliant response, all clients are stopped  and the test fails.
- originally created to reproduce the issue https://github.com/eclipse-ee4j/grizzly/issues/2125 of the Grizzly project,  which used Jersey and Grizzly together.

### DIFF
--- a/containers/grizzly2-http/pom.xml
+++ b/containers/grizzly2-http/pom.xml
@@ -17,7 +17,8 @@
 
 -->
 
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
     <modelVersion>4.0.0</modelVersion>
 
     <parent>
@@ -32,6 +33,14 @@
 
     <description>Grizzly 2 Http Container.</description>
 
+    <properties>
+        <!-- alternatives: JerseyHttpClientThread, JettyClientThread in the same package -->
+        <client>JdkHttpClientThread</client>
+        <clientImplPackage>org.glassfish.jersey.grizzly2.httpserver.test.tools</clientImplPackage>
+        <clientImpl>${clientImplPackage}.${client}</clientImpl>
+        <testMemory>-Xms160m -Xmx160m -Xss512k</testMemory>
+    </properties>
+
     <dependencies>
         <dependency>
             <groupId>jakarta.inject</groupId>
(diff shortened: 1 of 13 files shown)


jansupol:
BouncyCastle requires CQ, but it has a proprietary BouncyCastle license. Do we need the BouncyCastle?

dmatej:
> BouncyCastle requires CQ, but it has a proprietary BouncyCastle license. Do we need the BouncyCastle?

License should not be a problem: https://www.bouncycastle.org/fr/licence.html

dmatej:
But maybe I can try to simplify it, basically I reused my older code.
Also newest JDK updates have some issues with keystores as they removed some ciphers, so it would be safer if I would review my code once more.

dmatej:
Aaargh, I deleted wrong repository by mistake ... so I have to create a new PR ... I think there is no other way to fix it.

dmatej:
So, current state:

1. Please tell me if I should create new duplicit PR, because I deleted the original fork by mistake. Now I created new fork, based directly on the Eclipse's instead of Payara's fork. I am not sure if Github can merge it even when the repo vanished.
2. **I need some navigation for that CQ**, I "googled" just this [link](https://wiki.eclipse.org/Project_Management_Infrastructure/Creating_A_Contribution_Questionnaire), but I cannot find anything like that.
3. Alternative for the BouncyCastle exists, but it is [proprietary](https://www.oracle.com/java/technologies/faq-sun-packages.html) too - it is the keytool implementation under sun.security.x509.* package, which requires additional [instructions](https://stackoverflow.com/questions/61149455/access-sun-security-x509-in-jdk-11-without-modules) for the compiler. I expect this package will be completely replaced in future JDK releases, also I'm not sure if it is present in all current JDK distributions with the same api (but maybe is). But I would expect that the certificate management will significantly change in comming years, so ... from my point of view replacing BC dependency with some "future standard" is easier than having to mess with the compiler settings and all changes one by one (I still remember [issues](https://github.com/eclipse-ee4j/grizzly/pull/2112) with ALPN).

arjantijms:
> BouncyCastle requires CQ, but it has a proprietary BouncyCastle license. Do we need the BouncyCastle?

To create a self-signed cert, I think it's one of the only viable options. I typically use it too in tests like e.g. this:

```java
private static X509Certificate createSelfSignedCertificate(KeyPair keys) {
        try {
            Provider provider = new BouncyCastleProvider();
            Security.addProvider(provider);
            return new JcaX509CertificateConverter()
                    .setProvider(provider)
                    .getCertificate(
                            new X509v3CertificateBuilder(
                                    new X500Name("CN=lfoo, OU=bar, O=kaz, L=zak, ST=lak, C=UK"),
                                    ONE,
                                    Date.from(now()),
                                    Date.from(now().plus(1, DAYS)),
                                    new X500Name("CN=lfoo, OU=bar, O=kaz, L=zak, ST=lak, C=UK"),
                                    SubjectPublicKeyInfo.getInstance(keys.getPublic().getEncoded()))
                    .build(
                            new JcaContentSignerBuilder("SHA256WithRSA")
                                .setProvider(provider)
                                .build(keys.getPrivate())));
        } catch (CertificateException | OperatorCreationException e) {
            throw new IllegalStateException(e);
        }
    }
```

As it's only used for testing, it would concern a "works with" CQ, which is far more relaxed.

arjantijms:
A small ping here. @jansupol @dmatej any progress?

dmatej:
> A small ping here. @jansupol @dmatej any progress?

I thought you will create the request to approve using BouncyCastle in test dependencies ;)
Even your example uses the BouncyCastleProvider - there are no reliable alternatives, because the sun.* implementation in JDK which can do that too changes API between JDK versions. BC not.

senivam:
CQ#: [23715](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23715)
CQ#: [23714](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23714)

senivam (approved):
CQs are RESOLVED, LGTM

dmatej:
Seems there is some bug on GitHub - I just rebased the branch to current eclipse/3.x branch, but now GitHub says that the repository was deleted, which is not true. There were no conflicts, I verified even that the version is still the same. Is it still possible to merge it, or should I create a new PR, same as this one? 
I don't have write access here, I would try that otherwise.
https://github.com/dmatej/jersey/tree/http2-test-3.x

senivam:
@dmatej as from my point of view (at GH) the PR still exists and can be merged (w/o conflicts), so no reason to change anything.

Issue #3719, opened 2017-11-17
JerseyEventSink continues to write to Response after the Container has completed the async context


jerseyrobot:
Our product offers different REST APIs based on Jersey 2.26 running on Tomcat 8.5.

We are facing problems with asynchronous requests (using javax.ws.rs.sse.SseEventSink) which are closed by the client. If the async context is completed on server side, then it works as expected.

But when a client closes the connection, then the responses are recycled by Tomcat and reused for other requests. Tomcat will call onError method of any registered AsyncListener in AsyncContext. If none of the registered listeners are completing the AsyncContext, it will be done by the container. As far as I could see, Jersey does not register any listener to AsyncContext.
The asynchronous thread will just continue to write to the Response. If the Response is already reused by another request, then the SSE Event Sink is writing to this other Response. This leads to non-deterministic behaviour.
We have seen a wide range of unexpected resulte on the Client:
* Empty Response
* Truncated Responses
* CORS Problems (because response headers could not be written to a committed response)
* ...

As a workaround we had to disable the caching of http processors in Tomcat

See also the issue I have created for tomcat:
https://bz.apache.org/bugzilla/show_bug.cgi?id=61768

BCSol:
This is still an issue. The described workaround of setting `processorCache=0` did not work for our Ajp connector.

I was able to hotfix it by implementing `ConnectionCallback` and `CompletionCallback` in `ChunkedOutput` and registering itself as callback in the `AsyncContext`. The idea was to close the `ChunkedOutput`, when tomcat decides to complete the `AsyncContext` on connection loss.
Although this solution is working, it doesn't feel like the proper solution to the bug, especially because my experience with the internal framework and the Servlet 3.0 specification is pretty limited. So some feedback would be appreciated a lot. :)

PR #4741, opened 2021-03-08
WIP: Close ChunkedOuput onComplete in AsyncContext

BCSol:
@see https://github.com/eclipse-ee4j/jersey/issues/3719
@see https://bz.apache.org/bugzilla/show_bug.cgi?id=61768

On client disconnect JerseyEventSink is able to write to responses
after the container has completed. This causes non-deterministic
behaviour:
 - empty responses
 - truncated responses
 - sse event strings ending up in responses from other requests

The proposed fix is to register `ChunkedOuput` to the `AsyncContext` as
`CompletionCallback`, closing the resource.

We might want to consider calling `onClose` in `onComplete` as well.

### DIFF
--- a/core-server/src/main/java/org/glassfish/jersey/server/ChunkedOutput.java
+++ b/core-server/src/main/java/org/glassfish/jersey/server/ChunkedOutput.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2012, 2020 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2012, 2021 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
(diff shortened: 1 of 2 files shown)


BCSol:
The failing CI (dependency resolution) is most likely not caused by this merge request, since previous commits have the same issue.

BCSol:
wont fix

BCSol:
The issue causing this behaviour was somewhere else after all.
Even though the proposed fixes helped with the frequency of the problem, it was not able to fix it though.

In the faulty setup we connected apache2 with an ajp connector with tomcat8.
One servlet filter in our chain did not support async processing, this produced an error in the catalina log and seemed to have caused the issue of sse events showing up in other responses.

Issue #4034, opened 2019-01-22
Jersey Logging filter cuts entity - placeholder for #3306 and #3281


jansupol:
#3281 (Jersey Logging filter cuts entity)
#3306 (Please reopen the task)

dkBrazz:
I made a fix - #4753
Is there any chance that it will be merged into master?

senivam:
@dkBrazz yes, sure, we will review it soon. Thanks for the PR, it's appreciated.

PR #4753, opened 2021-03-22
Proper handling of chunked input streams in LoggingInterceptor

dkBrazz:
Fixes unexpected entity cutting in logs
#4034, #3306 and #3281

### DIFF
--- a/core-common/pom.xml
+++ b/core-common/pom.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
-    Copyright (c) 2010, 2020 Oracle and/or its affiliates. All rights reserved.
+    Copyright (c) 2010, 2021 Oracle and/or its affiliates. All rights reserved.
 
     This program and the accompanying materials are made available under the
     terms of the Eclipse Public License v. 2.0, which is available at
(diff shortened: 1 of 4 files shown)


jansupol (approved):

jansupol:
The PR looks good. What it does, though, enables to read all the chunks from the chunked stream. That is assumingly the preferred behaviour. I can imagine a case where people do not want to log all the chunks and want to log just the first one, the way it worked now. Perhaps for performance reasons, or to keep the log small. 
@senivam wdyt? Should we add a possibility to turn off logging all the chunks by the new LoggingFeature builder you created?

dkBrazz:
> a case where people do not want to log all the chunks and want to log just the first one, the way it worked now. Perhaps for performance reasons, or to keep the log small.  

I think existing [maxEntitySize](https://eclipse-ee4j.github.io/jersey.github.io/apidocs/latest/jersey/org/glassfish/jersey/logging/LoggingFeature.html#LoggingFeature-java.util.logging.Logger-java.lang.Integer-) should be enough to cover it

senivam (approved):
tests, provided in the PR prove that maxEntitySize works as designed, so, functionality with restricted entity size is not affected. For me it looks good.

Issue #4651, opened 2020-12-04
Add a ParamConverterProvider for java.util.Optional parameters


Gaff:
Would it be possible to include a `ParamCovnerterProvider` that supports `Optional<X>` types?

I notice DropWizard have some: https://github.com/dropwizard/dropwizard/blob/master/dropwizard-jersey/src/main/java/io/dropwizard/jersey/optional/OptionalParamBinder.java - but presumably this would be a generally useful feature!

Gaff:
@jansupol wow thanks for doing this! Much appreciated.

joschi:
@jansupol @jbescos Thanks for implementing this upstream!

Too bad you didn't implement providers for the other optional classes such as `OptionalDouble`, `OptionalInt`, and `OptionalLong` too, but we've got you covered: [`io.dropwizard.jersey.optional`](https://javadoc.io/static/io.dropwizard/dropwizard-jersey/2.0.21/io/dropwizard/jersey/optional/package-summary.html). 😉 

In the upstream implementation, I think there was an unfortunate decision to return `null` if the value couldn't be converted into the type wrapped by the `Optional<T>` container:
https://github.com/eclipse-ee4j/jersey/blob/52c8aa4d57f35ae2f93b81225cf271864e7a91cd/core-common/src/main/java/org/glassfish/jersey/internal/inject/ParamConverters.java#L282-L286

This means that users will have to check all `Optional<T>` parameters for `null` before interacting with them, otherwise they'll get a `NullPointerException` and clients would receive a response with status code 500 (Internal Server Error).

In my humble opinion this is counter-intuitive to the idiomatic usages of the `Optional<T>` type.

Unfortunately the test cases don't cover the erroneous cases:
https://github.com/eclipse-ee4j/jersey/blob/52c8aa4d57f35ae2f93b81225cf271864e7a91cd/tests/e2e-server/src/test/java/org/glassfish/jersey/tests/e2e/server/OptionalParamConverterTest.java#L79-L87

In order to properly cover the error cases, the resource method would have to be changed as follows:

Original:
https://github.com/eclipse-ee4j/jersey/blob/52c8aa4d57f35ae2f93b81225cf271864e7a91cd/tests/e2e-server/src/test/java/org/glassfish/jersey/tests/e2e/server/OptionalParamConverterTest.java#L47-L51

Changed version with correct error handling:
```java
@GET
@Path("/fromInteger")
[...]

PR #4783, opened 2021-04-26
Support more optionals

jbescos:
Adding support for OptionalInt, OptionalDouble and OptionalLong.

This was mentioned by @joschi here: https://github.com/eclipse-ee4j/jersey/issues/4651


senivam (approved):
this does make sense, but we have to deal with another PR on the same topic which is #4799 thus we will somehow merge those 2 PRs after fixing conflicts which would appear after one or another PR is merged. So, wait with merging but legitimate.

jbescos:
@joschi I just merged your changes with mine. Would you mind to review it please?.

core-common/src/main/java/org/glassfish/jersey/internal/inject/ParamConverters.java
```diff
@@ -305,6 +309,125 @@ public String toString(T value) throws IllegalArgumentException {
...
+
+            OPTIONAL(Optional.class) {
+                @Override
+                Object empty() {
+                    return Optional.empty();
+                }
+                @Override
+                Object of(Object value) {
+                    return Optional.of(value);
+                }
```
joschi, line 380: Do we need to cover this case? I think `Optional<T>` would already be handled by `OptionalCustomProvider` before `OptionalProvider` gets invoked.

tests/e2e-server/src/test/java/org/glassfish/jersey/tests/e2e/server/OptionalParamConverterNoProviderTest.java
```diff
@@ -0,0 +1,150 @@
...
+
+        @GET
+        @Path("/fromList")
+        public Response fromList(@QueryParam(PARAM_NAME) List<Optional<Integer>> data) {
+            StringBuilder builder = new StringBuilder("");
+            for (Optional<Integer> val : data) {
+                builder.append(val.orElse(0));
+            }
+            return Response.ok(builder.toString()).build();
+        }
```
joschi, line 64: Would it make sense to have a corresponding test for `List<OptionalInt>>`, `List<OptionalLong>>`, and `List<OptionalDouble>>`?

core-common/src/main/java/org/glassfish/jersey/internal/inject/ParamConverters.java
```diff
@@ -305,6 +309,125 @@ public String toString(T value) throws IllegalArgumentException {
...
+                            T val = aggregated.getConverter((Class<T>) type, type, annotations).fromString(value.toString());
+                            if (val != null) {
+                                return (T) optionals.of(val);
+                            } else {
+                                /*
+                                 *  In this case we don't send Optional.empty() because 'value' is not null.
+                                 *  But we return null because the provider didn't find how to parse it.
+                                 */
+                                return null;
+                            }
```
joschi, line 350: If `Optionals` covered only `OptionalInt`, `OptionalLong`, and `OptionalDouble` we wouldn't need to cover this case (which is already covered in `CustomOptionalProvider`).

core-common/src/main/java/org/glassfish/jersey/internal/inject/ParamConverters.java
jbescos: Right, it is not necessary

tests/e2e-server/src/test/java/org/glassfish/jersey/tests/e2e/server/OptionalParamConverterNoProviderTest.java
jbescos: Done

core-common/src/main/java/org/glassfish/jersey/internal/inject/ParamConverters.java
jbescos: Right, it is not necessary

joschi (approved):
LGTM. 👍

jansupol (approved):

PR #4785, opened 2021-04-27
JDK16 Support

jansupol:

### DIFF
--- a/ext/cdi/jersey-cdi1x-validation/src/main/java/org/glassfish/jersey/ext/cdi1x/validation/internal/CdiInterceptorWrapperExtension.java
+++ b/ext/cdi/jersey-cdi1x-validation/src/main/java/org/glassfish/jersey/ext/cdi1x/validation/internal/CdiInterceptorWrapperExtension.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015, 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2015, 2021 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
@@ -40,6 +40,7 @@ import javax.enterprise.inject.spi.InjectionTarget;
 import javax.enterprise.util.AnnotationLiteral;
 import javax.interceptor.Interceptor;
 
+import org.glassfish.jersey.ext.cdi1x.internal.JerseyVetoed;
 import org.glassfish.jersey.internal.util.collection.Cache;
 import org.glassfish.jersey.server.model.Resource;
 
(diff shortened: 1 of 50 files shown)


jansupol:
Required CQs for:

- org.ops4j.pax.url:pax-url-mvn:1.3.7
- org.apache.felix:org.apache.felix.framework:7.0.0

senivam (approved):

ext/cdi/jersey-cdi1x-validation/src/main/java/org/glassfish/jersey/ext/cdi1x/validation/internal/CdiInterceptorWrapperExtension.java
```diff
@@ -171,4 +176,12 @@ public void destroy(ValidationInterceptor instance,
             }
         });
     }
+
+    /* package */ Cache<Class<?>, Boolean> getJaxRsResourceCache() {
+        /* CDI injection hack */
+        if (jaxRsResourceCache == null) {
```
jbescos, line 182: Should we synchronize this or are we sure only one thread can invoke it the first time?. Something like:
```
    /* package */ Cache<Class<?>, Boolean> getJaxRsResourceCache() {
        /* CDI injection hack */
        if (jaxRsResourceCache == null) {
           synchronize(this) {
                if (jaxRsResourceCache == null) {
                      jaxRsResourceCache = new Cache<>(clazz -> Resource.from(clazz) != null);
                }
           }
        }
        return jaxRsResourceCache;
    }
```

jbescos (approved):
I wrote a small comment you to double check.

ext/cdi/jersey-cdi1x-validation/src/main/java/org/glassfish/jersey/ext/cdi1x/validation/internal/CdiInterceptorWrapperExtension.java
jansupol: For the sake, it better be initialized in a constructor of the `CdiInterceptorWrapper` to be called by just a single thread.

jbescos: I don't know all the details about how is going to be used, but imagine that there is a singleton that invokes that method and it saves it as a field. So next time it wants to use the cache, it read it from the field (instead of getting it from getJaxRsResourceCache()).

It will not harm to have that lock because it will only lock at the beginning and we get rid of possible memory issues.

jbescos (approved):

ext/cdi/jersey-cdi1x-validation/src/main/java/org/glassfish/jersey/ext/cdi1x/validation/internal/CdiInterceptorWrapperExtension.java
jansupol: It happens that all the `if (jaxRsResourceCache == null) {` block is unnecessary, so we could have dropped it

jbescos (approved):

Issue #4773, opened 2021-04-19
NullPointerException in HeaderUtils.getPreferredCookie


chilloutman:
Introduced in: #4390 (which addressed #3433)

Cookie paths can be null. This is not handled in HeaderUtils.java:326, which causes a NullPointerException to be thrown.

```java
return first.getPath().length() > second.getPath().length() ? first : second;
```

---

The problem occurs when the client receives something like the following, where it did not in versions prior to #4390 being implemented.

```
Set-Cookie: A=B,A=C
```

```java
public static void main (String[] args) {
	var a = new NewCookie("NAME", "VALUE");
	var b = new NewCookie("NAME", "VALUE");
	try {
		HeaderUtils.getPreferredCookie(a, b);
	} catch (NullPointerException e) {
		e.printStackTrace();
	}
}
```

PR #4795, opened 2021-05-12
Fix issue NullPointerException in HeaderUtils.getPreferredCookie #4773

tvallin:

### DIFF
--- a/core-common/src/main/java/org/glassfish/jersey/message/internal/HeaderUtils.java
+++ b/core-common/src/main/java/org/glassfish/jersey/message/internal/HeaderUtils.java
@@ -311,21 +311,11 @@ public final class HeaderUtils {
      *              - if equal, compare path length.
      */
     public static NewCookie getPreferredCookie(NewCookie first, NewCookie second) {
-
-        if (first == null) {
-            return second;
-        } else if (second == null) {
-            return first;
-        }
-
-        if (first.getMaxAge() != second.getMaxAge()) {
-            return Comparator.comparing(NewCookie::getMaxAge).compare(first, second) > 0 ? first : second;
-        } else if (first.getExpiry() != null && second.getExpiry() != null && !first.getExpiry().equals(second.getExpiry())) {
-            return Comparator.comparing(NewCookie::getExpiry).compare(first, second) > 0 ? first : second;
-        } else if (first.getPath() != null && second.getPath() != null) {
-            return first.getPath().length() > second.getPath().length() ? first : second;
-        }
-        return first;
+        return Comparator.nullsFirst(
+                Comparator.comparingInt(NewCookie::getMaxAge)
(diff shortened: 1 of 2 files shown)


core-common/src/main/java/org/glassfish/jersey/message/internal/HeaderUtils.java
```diff
@@ -303,12 +303,12 @@ public static void checkHeaderChanges(final Map<String, String> headersSnapshot,
...
+     * @param first     NewCookie to be compared.
+     * @param second    NewCookie to be compared.
      * @return the preferred NewCookie according to rules :
      *              - the latest maxAge.
-     *              - if equal, compare the expiry date
-     *              - if equal, compare name length
+     *              - if equal, compare the expiry date.
+     *              - if equal, compare path length.
      */
     public static NewCookie getPreferredCookie(NewCookie first, NewCookie second) {
```
senivam, line 313: I would advise to rewrite the whole method into something like:
```
public static NewCookie getPreferredCookie(NewCookie first, NewCookie second) {

        return Comparator.nullsFirst(
                Comparator.comparingInt(NewCookie::getMaxAge)
                        .thenComparing(NewCookie::getExpiry, Comparator.nullsLast(Comparator.naturalOrder()))
                        .thenComparing(NewCookie::getPath, Comparator.nullsLast(Comparator.comparing(String::length))))
                        .compare(first, second) > 0 ? first : second;
    }
```
because it starts to remind a nightmare with all those conditions.

senivam (changes requested):

core-common/src/main/java/org/glassfish/jersey/message/internal/HeaderUtils.java
tvallin: Good idea !

tests/e2e-core-common/src/test/java/org/glassfish/jersey/tests/e2e/common/message/internal/HeaderUtilsTest.java
```diff
@@ -227,4 +227,22 @@ public void testgetPreferredCookie(){
...
+        NewCookie first = new NewCookie("NAME", "VALUE");
+        NewCookie second = new NewCookie("NAME", "VALUE");
+        NewCookie returnedCookie = HeaderUtils.getPreferredCookie(first, second);
+
+        assertEquals(first, returnedCookie);
+    }
+
+    @Test
+    public void testGetPreferredCookieWithNullInput() {
+        NewCookie first = null;
```
senivam, line 241: it could be transformed into
```
assertNull(HeaderUtils.getPreferredCookie(null, null));
```

senivam (changes requested):

senivam (approved):

jansupol (approved):

Issue #3087, opened 2015-03-04
Add support for Optional from Java 8 into Jersey


jerseyrobot:
We're currently able to support some of the types from Java 8 in Jersey (extension module). Create infrastructure for this extension module and add support for Optional (entity, param).
#### Affected Versions
[2.16]

joschi:
This has been addressed via https://github.com/eclipse-ee4j/jersey/issues/4651 and https://github.com/eclipse-ee4j/jersey/pull/4690.

PR #4799, opened 2021-05-23
Support custom parameter types with `Optional<T>`

### LINKED ISSUE #4798: `ParamConverter` for `Optional<T>` doesn't support custom types
The `ParamConverter` implementation for the Java `Optional<T>` container which was introduced in Jersey 2.34 (and Jersey 3.0.1) doesn't support handling custom types.

With `Optional<T>` being a container similar to `Collection<T>` (and its derivates), I would expect it to support custom types for which a `ParamConverterProvider` and `ParamConverter` exist.

Unfortunately this is not the case. `org.glassfish.jersey.internal.inject.ParamConverters.OptionalProvider` only supports the other types defined in `org.glassfish.jersey.internal.inject.ParamConverters.AggregatedProvider` but not any other type of parameters for which a `ParamConverterProvider` and `ParamConverter` have been registered.

Here's an example (using the Jersey testing framework) with a custom `ParamConverterProvider` and `ParamConverter` for `java.time.Instant`:
```java
import org.glassfish.jersey.internal.inject.ExtractorException;
import org.glassfish.jersey.server.ResourceConfig;
import org.glassfish.jersey.test.JerseyTest;
import org.junit.Test;

import javax.ws.rs.*;
import javax.ws.rs.core.*;
import javax.ws.rs.ext.*;
import java.lang.annotation.Annotation;
import java.lang.reflect.Type;
import java.time.Instant;
import java.util.Optional;

import static org.junit.Assert.assertEquals;

public class CustomParamConverterOptionalTest extends JerseyTest {
    @Override
    protected Application configure() {
        return new ResourceConfig(TestResource.class)
                .register(InstantParamConverterProvider.class);
    }

    @Test
    public void testInstant() {
        String paramMissing = target("/instant")
                .request().get(String.class);
        assertEquals("null", paramMissing);

        String optionalParamMissing = target("/optional")
                .request().get(String.class);
        assertEquals("default", optionalParamMissing);

        String paramProvided = target("/instant").queryParam("i", "2021-05-18T22:49:00Z")
                .request().get(String.class);
        assertEquals("2021-05-18T22:49:00Z", paramProvided);

        String optionalParamProvided = target("/optional").queryParam("i", "2021-05-18T22:49:00Z")
                .request().get(String.class); // 💥 Fails with: javax.ws.rs.NotFoundException: HTTP 404 Not Found
        assertEquals("2021-05-18T22:49:00Z", optionalParamProvided);
    }

    @Produces(MediaType.TEXT_PLAIN)
    @Path("/")
    public static class TestResource {
        @GET
        @Path("/optional")
        public String optionalInt(@QueryParam("i") Optional<Instant> i) {
            if (i == null) {
                return "null";
            }
            return i.map(Instant::toString).orElse("default");
        }

        @GET
        @Path("/instant")
        public String optionalInt(@QueryParam("i") Instant i) {
            if (i == null) {
                return "null";
            }
            return i.toString();
        }
    }

    @Provider
    public static class InstantParamConverterProvider implements ParamConverterProvider {
        @Override
        public <T> ParamConverter<T> getConverter(Class<T> rawType, Type genericType, Annotation[] annotations) {
            if (rawType.equals(Instant.class)) {
                return new ParamConverter<T>() {
                    @Override
                    public T fromString(String value) {
                        if (value == null) {
                            throw new IllegalArgumentException();
                        }
                        try {
                            return rawType.cast(Instant.parse(value));
                        } catch (Exception e) {
                            throw new ExtractorException(e);
                        }
                    }

                    @Override
                    public String toString(T value) {
                        if (value == null) {
                            throw new IllegalArgumentException();
                        }
                        return value.toString();
                    }
                };
            } else {
                return null;
            }
        }
    }
}
```

Refs #4651
Refs #4690
Refs #4790

joschi:
The `OptionalProvider` (`ParamConverterProvider`) introduced in #4690 didn't support custom parameter types wrapped in an `Optional<T>` but only the types implemented in `org.glassfish.jersey.internal.inject.ParamConverters`.

This PR adds support for arbitrary parameter types by looking up the parameter converter providers via `InjectionManager`.

Refs #3087
Refs #4651
Refs https://github.com/dropwizard/dropwizard/issues/3969
Closes #4798

### DIFF
--- a/core-common/src/main/java/org/glassfish/jersey/internal/inject/ParamConverterConfigurator.java
+++ b/core-common/src/main/java/org/glassfish/jersey/internal/inject/ParamConverterConfigurator.java
@@ -32,7 +32,7 @@ public class ParamConverterConfigurator implements BootstrapConfigurator {
     @Override
     public void init(InjectionManager injectionManager, BootstrapBag bootstrapBag) {
         InstanceBinding<ParamConverters.AggregatedProvider> aggregatedConverters =
-                Bindings.service(new ParamConverters.AggregatedProvider())
+                Bindings.service(new ParamConverters.AggregatedProvider(injectionManager))
                     .to(ParamConverterProvider.class);
         injectionManager.register(aggregatedConverters);
     }
--- a/core-common/src/main/java/org/glassfish/jersey/internal/inject/ParamConverters.java
+++ b/core-common/src/main/java/org/glassfish/jersey/internal/inject/ParamConverters.java
@@ -21,11 +21,11 @@ import java.lang.annotation.Annotation;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
-import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.security.AccessController;
 import java.text.ParseException;
 import java.util.Date;
+import java.util.List;
 import java.util.Optional;
 
 import javax.inject.Inject;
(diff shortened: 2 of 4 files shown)


jbescos (approved):
LGTM, thank you

jansupol (approved):

PR #4814, opened 2021-06-25
TimeWindowStatisticsImplTest Fix

jbescos:
In Jenkins the default value of 8 is loaded before the property 3 is set.


jbescos:
I closing it as @senivam found other way.

PR #4831, opened 2021-07-27
Use '; ' as a separator between values of Cookie or Cookie2 request headers instead of ',' ...

plevart:
I propose this patch for issue https://github.com/eclipse-ee4j/jersey/issues/2811

### DIFF
--- a/connectors/jdk-connector/src/main/java/org/glassfish/jersey/jdk/connector/internal/HttpRequestEncoder.java
+++ b/connectors/jdk-connector/src/main/java/org/glassfish/jersey/jdk/connector/internal/HttpRequestEncoder.java
@@ -36,10 +36,13 @@ class HttpRequestEncoder {
 
     private static void appendUpgradeHeaders(StringBuilder request, Map<String, List<String>> headers) {
         for (Map.Entry<String, List<String>> header : headers.entrySet()) {
+            String delimiter = "Cookie".equalsIgnoreCase(header.getKey()) ||
+                               "Cookie2".equalsIgnoreCase(header.getKey())
+                               ? "; " : ",";
             StringBuilder value = new StringBuilder();
             for (String valuePart : header.getValue()) {
                 if (value.length() != 0) {
-                    value.append(",");
+                    value.append(delimiter);
                 }
                 value.append(valuePart);
             }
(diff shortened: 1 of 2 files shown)


plevart:
I have now signed ECA. How to force a re-check above?

plevart:
Just a note that the presented patch solves my problem of correctly interpreting Cookies on the server side.

jansupol:
Note: Cookie2 is from [RFC 2965](https://datatracker.ietf.org/doc/html/rfc2965), obsolete by RFC 6265.

jansupol:
Hi, thank you for your PR.

I am not sure about this. The HTTP headers are delimited by "," separator, whereas ";" separator is used to separate multiple parts of a single HTTP header value. For instance `Accept-Encoding: compress;q=0.5, gzip;q=1.0`. 

The question is how the header values got split into multiple strings in the list when ";" is used. Unfortunately, there are multiple ways to enter the headers into Jersey. A test case would make it clear how was it done.

plevart:
> 
> Unfortunately, there are multiple ways to enter the headers into Jersey. A test case would make it clear how was it done.

These headers were not "entered" into Jersey by the app, but by the JdkConnectorProvider or HttpUrlConnectorProvider as part of processing the Set-Cookie response header which stores cookies that are later added to Cookie request header that follows. So this works automatically without any user code. It mimics what browsers do with cookies.

plevart:
> I am not sure about this. The HTTP headers are delimited by "," separator, whereas ";" separator is used to separate multiple parts of a single HTTP header value. For instance `Accept-Encoding: compress;q=0.5, gzip;q=1.0`.
> 

See: https://datatracker.ietf.org/doc/html/rfc6265#section-5.4

There should only be one Cookie request header (there can be many Set-Cookie response headers), so multiple cookie values (which are pairs of key=value) must be sent with singe Cookie request header. You can view multiple key=value pairs as parts of a single Cookie header value which are delimited with ';'. A single key=value pair will never have ';' in it, so this should be safe.

For examle, taking a look at what Mozilla Firefox does. It sends a single Cookie request header that looks like this:

Cookie: experimentation_subject_id=censored_value; _ga=censored_value; _fbp=censored_value; _rdt_uuid=censored_value

In this example, 4 cookies were being sent to the server.

plevart:
> Note: Cookie2 is from [RFC 2965](https://datatracker.ietf.org/doc/html/rfc2965), obsolete by RFC 6265.

I agree, the patch should only consider "Cookie" as the request header when special-casing the concatenation of parts (values).

PR #4835, opened 2021-08-05
Prevent NoSuchMethodError when used MP Rest Client 1.4 API & CDI

jansupol:


jansupol:
Also a cleanup in mp rest client implementation

senivam (approved):

PR #4837, opened 2021-08-12
add possibility to use entity with http method Options in requests according to the RFC 7231

aserkes:
add possibility to use entity with http method Options in requests according to the RFC 7231


jansupol (approved):

jansupol (approved):
Please update the HttpMethodEntityTest.testOptions test.

senivam (approved):
LGTM

PR #4840, opened 2021-08-16
Shallow cloning on travis

YunLemon:
According to [https://docs.travis-ci.com/user/customizing-the-build#git-clone-depth](url), Travis CI provide a way to shallow clone a repository. This has the obvious benefit of speed, since you only need to download a small number of commits.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,4 +27,4 @@ script:
   - bash etc/travis/travis.sh $TEST_SET
 
 git:
-  depth: false
+  depth: 3


senivam:
you are right, but shallow cloning was turned off on purpose due to copy-right check being applied to (almost) all files in the project. When shallow cloning is used it is not possible to perform copy-right check because files are being cloned with wrong creation date. 

The other thing - Travis builds are not used anymore. Jersey project moved CI jobs to Eclipse's jenkins.

PR #4859, opened 2021-09-16
Implementation of EntityPart API

jansupol:

### DIFF
--- a/core-common/src/main/java/org/glassfish/jersey/internal/AbstractRuntimeDelegate.java
+++ b/core-common/src/main/java/org/glassfish/jersey/internal/AbstractRuntimeDelegate.java
@@ -25,6 +25,7 @@ import java.util.WeakHashMap;
 import jakarta.ws.rs.core.CacheControl;
 import jakarta.ws.rs.core.Configuration;
 import jakarta.ws.rs.core.Cookie;
+import jakarta.ws.rs.core.EntityPart;
 import jakarta.ws.rs.core.EntityTag;
 import jakarta.ws.rs.core.Link;
 import jakarta.ws.rs.core.MediaType;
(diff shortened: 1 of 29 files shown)


core-common/src/main/java/org/glassfish/jersey/internal/AbstractRuntimeDelegate.java
```diff
@@ -50,6 +52,8 @@
 
     private final Set<HeaderDelegateProvider> hps;
     private final Map<Class<?>, HeaderDelegate<?>> map;
+    private static final Object EPB_LOCK = new Object();
+    private static volatile EntityPartBuilderProvider cachedEntityPartBuilderProvider;
```
jbescos, line 56: You can get rid of both variables with an inner static class:
```
private static class Lazy {

    private static final EntityPartBuilderProvider cachedEntityPartBuilderProvider = findEntityPartBuilderProvider();

    /**
     * Obtain a {@code RuntimeDelegate} instance using the method described in {@link #getInstance}.
     *
     * @return an instance of {@code RuntimeDelegate}.
     */
    private static EntityPartBuilderProvider findEntityPartBuilderProvider() {
        for (final EntityPartBuilderProvider entityPartBuilder : ServiceFinder.find(EntityPartBuilderProvider.class)) {
            if (entityPartBuilder != null) {
                return entityPartBuilder;
            }
        }
        throw new IllegalArgumentException(LocalizationMessages.NO_ENTITYPART_BUILDER_FOUND());
    }
}
```

And later get the instance in this way. It will be loaded the first time is used.
`Lazy.cachedEntityPartBuilderProvider`

jansupol: We may utilize Jersey `LazyValue` for this

senivam (approved):

jbescos (approved):

PR #4908, opened 2021-10-26
Default Exception Mapper re-worked

senivam:


core-server/src/main/java/org/glassfish/jersey/server/ServerRuntime.java
```diff
@@ -129,6 +129,11 @@
     /** Resolve relative URIs according to RFC7231 (not JAX-RS 2.0 compliant */
     private final boolean rfc7231LocationHeaderRelativeUriResolution;
 
+    /**
+     * Default exception mapper (@since 3.1.0 according to JAX-RS 3.1 spec)
+     */
+    private static final ExceptionMapper<Throwable> DEFAULT_EXCEPTION_MAPPER = new DefaultExceptionMapper();
```
jansupol, line 135: Now we can make `DefaultExceptionMapper` a package-private class in core-server

core-server/src/main/java/org/glassfish/jersey/server/ServerRuntime.java
```diff
@@ -510,7 +522,7 @@ private Response mapException(final Throwable originalThrowable) throws Throwabl
                             && throwable.getCause() instanceof MessageBodyProviderNotFoundException) {
                         throw throwable;
                     }
-                    Response waeResponse;
```
jansupol, line 513: Now we probably can move the HeaderValueException block back where it was in 3.0.3 so that HeaderValueException can be mappable.

jansupol (changes requested):

jansupol:
Also, I assume `core-server/src/test/java/org/glassfish/jersey/server/AsyncCallbackServerTest.java#testCompletionFail` should work again.

core-server/src/main/java/org/glassfish/jersey/server/ServerRuntime.java
```diff
@@ -441,8 +452,9 @@ public void process(final Throwable throwable) {
                         completionCallbackRunner.onComplete(responseError);
                     }
                 }
+                defaultMapperResponse = processResponseWithDefaultExceptionMapper(responseError, request);
```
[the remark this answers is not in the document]
jansupol: I am thinking about rising the logger level from FINE to INFO. That way the server log will still contain the error information, even though it is not propagated to the underlying container.

jansupol (approved):

PR #4911, opened 2021-11-11
Missing Content-Type header should be application/octet-stream optionally

jansupol:

### DIFF
--- a/etc/jenkins/jenkins_build.sh
+++ b/etc/jenkins/jenkins_build.sh
@@ -2,4 +2,4 @@
 
 export DEBUG=true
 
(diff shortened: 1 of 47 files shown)


etc/jenkins/jenkins_build.sh
```diff
@@ -2,4 +2,4 @@
 
 export DEBUG=true
 
-mvn -V -U -B -e -Psnapshots clean install glassfish-copyright:check -Dcopyright.quiet=false
\ No newline at end of file
+mvn -V -U -B -e -Pstaging clean install glassfish-copyright:check -Dcopyright.quiet=false
```
senivam, line 5: I would advise keeping of both profiles here ``-Pstaging,snapshots``

Issue #3493, opened 2017-01-13
Add BeanParam support in Jersey 2 client


jerseyrobot:
#### Affected Versions
[2.25]

andrei-ivanov:
Any chance to get the PR reviewed and merged? 😕

senivam:
In order for this ( [235](https://github.com/jersey/jersey/pull/235) ) PR to be reviewed some enthusiast shall re-publish it here (to new Jersey) with valid ECA signed. Afterwards it will be reviewed.

jansupol:
The bean params are supported by the MP Rest Client which has similar functionality as the proxy client. The #4086 is using BeanParams. Would it help?

andrei-ivanov:
@lucastsa any chance you would re-publish your PR here?

@jansupol not sure, in theory there's a single place where we create these client proxies, so the changes should be minimal, but I would also have to check if the MP Rest Client requires any other changes (we use the JAX-RS / Jersey / Spring / Spring Boot stack). One other thing is that #4086 seems yet to be merged, so I can't actually test.

derekm:
@jansupol -- WebResourceFactory is still better, more mature, more battle-tested than MP Rest Client in several respects, despite missing `BeanParam`s and `ParamConverterProvider`s, etc.

For example, https://github.com/eclipse/microprofile-rest-client/issues/217

It would be a good idea to continue to maintain WebResourceFactory & have WRF & MP Rest Client use the same backend abstractions/proxies. I'd like to a MP Rest Client move away from the fat, clunky `RestClientBuilder` API and towards the leaner, more advanced WRF API...

hoangbnc:
I saw `@BeanParam` being added to `WebResourceFactory` in `v3`. Could that be backported to `v2` as well?

oramattkosem:
I concur.  This would be a helpful feature addition in v2.

prmx:
This issue could be closed, as BeanParams support already backported to jersey#2.41

senivam:
closed as resolved

PR #4916, opened 2021-11-26
Issue #3493 - Add BeanParam support to WebresourceFactory

robersheimer:
Allows the proxy client holding a resource interface to supply
a bean of the corresponding type for the method fields annotated
with @BeanParam, instead of having to supply the associated
parameters individually. This also works if BeanParams themselves
contain fields annotated with @BeanParam.

Signed-off-by: Richard Obersheimer <richard.obersheimer@anderscore.com>

### DIFF
--- a/ext/proxy-client/src/main/java/org/glassfish/jersey/client/proxy/RequestParameters.java
+++ b/ext/proxy-client/src/main/java/org/glassfish/jersey/client/proxy/RequestParameters.java
@@ -1,0 +1,203 @@
+package org.glassfish.jersey.client.proxy;
+
+import jakarta.ws.rs.BeanParam;
+import jakarta.ws.rs.CookieParam;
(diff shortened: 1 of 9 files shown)


jansupol:
@robersheimer Thanks for the PR. It looks interesting. 

But, I'd like to ask you to provide the PR against 3.1 branch. I know it is confusing, sorry for that. We have 3.1 branch separated because it uses a new API (3.1), which is not in central, yet. But Jakarta EE 10 is really close, and after the release, we fix the branch names. 

Make sure to use -Pstaging profile when building the branch locally. Also, for the PR to pass the CI tests, the copyright headers will need to be present.

robersheimer:
Thank you, I tried again from a branch based on the 3.1 at https://github.com/eclipse-ee4j/jersey/pull/4919

PR #4919, opened 2021-11-29
Issue #3493 - Add BeanParam support to WebResourceFactory

robersheimer:
Allows the proxy client holding a resource interface to supply
a bean of the corresponding type for the method fields annotated
with @BeanParam, instead of having to supply the associated
parameters individually. This also works if BeanParams themselves
contain fields annotated with @BeanParam.

Signed-off-by: Richard Obersheimer <richard.obersheimer@anderscore.com>


jansupol (approved):

jansupol:
Thank you @robersheimer.

PR #4929, opened 2021-12-10
Updated dependency versions

arjantijms:
Signed-off-by: arjantijms <arjan.tijms@gmail.com>

### DIFF
--- a/media/json-jettison/pom.xml
+++ b/media/json-jettison/pom.xml
@@ -53,7 +53,7 @@
                 <extensions>true</extensions>
                 <configuration>
                     <instructions>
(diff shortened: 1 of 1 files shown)


jansupol (approved):
Technically, this is correct, GF 7 will use JAX-B 4.0. But please know that GF is not the only OSGi environment Jersey is used in and for those environments a comment about supporting both JAX-B 3.0 and JAX-B 4.0 would make better sense.

arjantijms:
> a comment about supporting both JAX-B 3.0 and JAX-B 4.0 would make better sense.

I simply copied the comment from the similar updates that were done before in Jersey ;)

PR #4943, opened 2021-12-21
Multipart filename support RFC6266 without language

zapodot:
The RFC6266 does support having a language identifier as part of an encoded filename (content disposition param filename*), however the current Jersey Multipart parsing will not decode the filename if the language identifier is not present.

Example:
```
Content-Disposition: form-data; name=myfile; filename=myfile.pdf; filename*=utf-8''myfile.pdf

<content>
```
Currently this when you bind a FormDataContentDisposition parameter the filename value will be `utf-8''myfile.pdf`


jansupol:
This looks like it makes sense, we do not need to have the charset a part of the header, it is already in the String actual charset.

Issue #4965, opened 2022-01-17
Content-length set to -1 when it is not sent from the server


kjezek:
When a request is triggered to a server endpoint, which does not provide Content-length, the ApacheConnector adds Content-length with the value equals to -1.

The problematic code is here: 
https://github.com/eclipse-ee4j/jersey/blob/e84b5e897a459ee76693498290dade0709e182fd/connectors/apache-connector/src/main/java/org/glassfish/jersey/apache/connector/ApacheConnector.java#L512-L514

It should probably be:

```
                if (headers.get(HttpHeaders.CONTENT_LENGTH) == null && entity.getContentLength() > 0) {
                    headers.add(HttpHeaders.CONTENT_LENGTH, String.valueOf(entity.getContentLength()));
                }
```

The main problem is that the ApacheConnector generates the header value, which is against the specification, i.e. when the content length is not know, this header should not be provided at all, not provided with a negative value.

PR #4967, opened 2022-01-17
extended condition not to add the content-length header negative

kjezek:
This update prevents wrong adding of the Content-Length header with the value -1. Such a value is not defined in the specification. The header should not be present at all when the length is not defined.

### DIFF
--- a/connectors/apache-connector/src/main/java/org/glassfish/jersey/apache/connector/ApacheConnector.java
+++ b/connectors/apache-connector/src/main/java/org/glassfish/jersey/apache/connector/ApacheConnector.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2020 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2010, 2022 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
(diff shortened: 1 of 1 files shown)


kjezek:
This should fix: https://github.com/eclipse-ee4j/jersey/issues/4965

jansupol:
Thank you for the PR. The continuous integration check did fail because the copyright year is not updated to 2022.

The PR does not seem right. Why did you exclude zero-length content? Can you provide a test that demonstrates the use-case this PR helps?

kjezek:
Hi @jansupol 
1. I have fixed the license year issues.

2. The Content-Length header must not be added to the header list when the length is negative. First, it is against the specification - when the length is not know, the header must be omitted, not used with a negative number.  Secondly, the code I have fixed did originally add the wrong Content-Length header when the header was actually not present in the original response - i.e. it modified the response wrongly, adding a wrong header.

jansupol:
I can understand the negative number.
My question was why did you want to remove header for the zero length content, i.e. when entity.getContentLength() == 0?

kjezek:
@jansupol you are right, the length of Content-Length is a valid value.  I have fixed the condition.

jansupol (approved):

jbescos (approved):

PR #4968, opened 2022-01-17
Enable percent encoding servlet context path and servlet path

jansupol:


jbescos (approved):

senivam (approved):

consultantleon:
I see you linked to 2.36 milestone, any forecast when this will be released? I saw it's included in the 3.1.0-M5 milestone but for 2.x there do not seem to be milestone releases?

senivam:
Hi @consultantleon, thank you for the question. The 3.1.0 is the major release so we publish milestones for it. While 2.36 will be a patch release so it will be released as the next in a row. Regarding estimates - as soon as we have all required patches included we will have the release.

PR #4971, opened 2022-01-31
Netty Connector support for TLS over Proxy

olotenko:
1. Support null PROXY_USERNAME

2. Support SSL Hostname verification

3. Update request host and port in SSL Context

4. Correct order of TLS handler vs Proxy handler

### DIFF
--- a/connectors/netty-connector/src/main/java/org/glassfish/jersey/netty/connector/NettyClientProperties.java
+++ b/connectors/netty-connector/src/main/java/org/glassfish/jersey/netty/connector/NettyClientProperties.java
@@ -53,4 +53,6 @@ public class NettyClientProperties {
      *  </p>
      */
     public static final String MAX_CONNECTIONS = "jersey.config.client.maxConnections";
+
+    public static final String ENABLE_SSL_HOSTNAME_VERIFICATION = "jersey.config.client.tls.enableHostnameVerification";
 }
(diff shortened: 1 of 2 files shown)


connectors/netty-connector/src/main/java/org/glassfish/jersey/netty/connector/NettyClientProperties.java
```diff
@@ -53,4 +53,6 @@
      *  </p>
      */
     public static final String MAX_CONNECTIONS = "jersey.config.client.maxConnections";
+
+    public static final String ENABLE_SSL_HOSTNAME_VERIFICATION = "jersey.config.client.tls.enableHostnameVerification";
```
olotenko, line 57: Not sure if there is a common client config setting. This mimics the setting as in Jetty connector.

connectors/netty-connector/src/main/java/org/glassfish/jersey/netty/connector/NettyConnector.java
```diff
@@ -234,9 +232,28 @@ protected void initChannel(SocketChannel ch) throws Exception {
                          final String password = ClientProperties.getValue(
                                  config.getProperties(), ClientProperties.PROXY_PASSWORD, String.class);
 
-                         p.addLast(new HttpProxyHandler(new InetSocketAddress(u.getHost(),
-                                                                              u.getPort() == -1 ? 8080 : u.getPort()),
-                                                        userName, password));
+                         InetSocketAddress proxyAddr = new InetSocketAddress(u.getHost(),
+                                                                             u.getPort() == -1 ? 8080 : u.getPort());
+                         p.addLast(userName == null ? new HttpProxyHandler(proxyAddr)
```
olotenko, line 237: Allow anonymous proxies that don't require username / password

jansupol:
Closing in favor of #5046.

PR #4985, opened 2022-02-16
JPMS for main modules

senivam:


senivam:
closed in favor of #5001

Issue #4984, opened 2022-02-11
InboundMessageContext.getLength() blows up for CONTENT_LENGTH that exceed Integer limits


lalitmathwani:
We have a use case where we are using the jersey client (@ 2.33 currently) to issue a Http HEAD to an endpoint that hosts files for download.
For files sizes that exceed a threshold of 2 Gigs approx - the maximum Integer limit (2147483647) - we start to see NumberFormatExceptions. 

This issue seems to occur during the CONTENT_LENGTH parsing here  :
https://github.com/eclipse-ee4j/jersey/blob/9bb2e8dffe2ee307dde151b35f3b98fc81782d4c/core-common/src/main/java/org/glassfish/jersey/message/internal/InboundMessageContext.java#L435

Caused by: java.lang.NumberFormatException: For input string: \"4294967296\"\n\t
at java.base/java.lang.NumberFormatException.forInputString(Unknown Source)\n\t
at java.base/java.lang.Integer.parseInt(Unknown Source)\n\t
at java.base/java.lang.Integer.parseInt(Unknown Source)\n\t
at org.glassfish.jersey.message.internal.InboundMessageContext$4.apply(InboundMessageContext.java:435)\n\t... 34 common frames omitted\n"

Checking if there a reason why the CONTENT_LENGTH values were constrained to Integers instead of Longs..

jansupol:
HTTP RFC does not have a limit, browsers [do](https://www.motobit.com/help/scptutl/pa98.htm), Servlet [can use longs](https://javaee.github.io/javaee-spec/javadocs/javax/servlet/ServletRequest.html#getContentLengthLong--) since 3.1.

Can Jersey use an additional method (getLengthLong)?

jbescos:
> HTTP RFC does not have a limit, browsers [do](https://www.motobit.com/help/scptutl/pa98.htm), Servlet [can use longs](https://javaee.github.io/javaee-spec/javadocs/javax/servlet/ServletRequest.html#getContentLengthLong--) since 3.1.
> 
> Can Jersey use an additional method (getLengthLong)?

I'm going to check this

jbescos:
Adding stacktrace to see how to reproduce the issue.

It happens when javax.ws.rs.core.Response#getLength is invoked in user code. It looks Jersey is not invoking this method.

[...]

PR #4994, opened 2022-03-01
InboundMessageContext.getLength() blows up for CONTENT_LENGTH that exceed Integer limits

jbescos:
[Issue ](https://github.com/eclipse-ee4j/jersey/issues/4984)

I don't see a reason to not use a long instead of an integer.


jansupol:
@jbescos Can the changed return type of the public method cause compatibility issues?

jansupol:
See for instance ServletRequest#contentLength & ServletRequest#getContentLengthLong

jbescos:
@jansupol right, I just found these issues right now. I am closing this.

PR #4745, opened 2021-03-10
Logging delimiter parametrized

senivam:
Signed-off-by: Maxim Nesen <maxim.nesen@oracle.com>

Issue #4734, opened 2021-03-01
Print request/response logs in a single line


sarun85:
As the LoggingFeature is splitting the request data into several lines, it is difficult to filter the logs based on the thread id. Is it possible to provide an option in the LoggingFeature to pass in a delimiter as an optional parameter that will replace the line terminator "\n". This will help our project to filter the logs of a particular request from the log file using a script.

Thanks.

senivam:
Hi @sarun85, for now line separator is hard-codded into LoggingInterceptor, but I think it won't be a big trouble to make i parametrized in future releases.

PR #5003, opened 2022-03-14
Fix logging delimiter parameterization

nunomsantos:
Hi all,

While reading the classes and tests for the logging feature I noticed that the code introduced in pull request #4745 to add the improvement requested on #4734 is not working as intended.
This has not been noticed by the existing tests because they were using a constructor that would not pass through the problematic code.

Currently, when registering the `LoggingFeature` and setting the `LoggingFeature.LOGGING_FEATURE_SEPARATOR` property, the empty constructor is called, which calls other constructors with nulls and so on.

Eventually we reach the constructor using a `LoggingFeatureBuilder`, where the arguments are set to be passed to the last constructor.
There, the builder is also set with the default separator, which later causes the following code to ignore the configuration properties on `configureBuilderParameters(...)` .
```java
final String filterSeparator = CommonProperties.getValue(/* read config from properties */);
// ...
// Config read from properties is never set
builder.separator = builder.separator == null ? filterSeparator : builder.separator;
```

Since the builder is initially set with the default separator, it is never overridden by the configurations.
The fix is just to remove the line setting the `DEFAULT_SEPARATOR` on the builder before passing to the last constructor.

I guess a conversation could be had about the other configurations: anything set as non-null in the builder cannot be overridden by configurations later; but that's likely okay because those other constructors are only used in tests and the empty constructor just sets everything as null.

### DIFF
--- a/core-common/src/main/java/org/glassfish/jersey/logging/LoggingFeature.java
+++ b/core-common/src/main/java/org/glassfish/jersey/logging/LoggingFeature.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016, 2021 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2016, 2022 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
@@ -222,7 +222,6 @@ public class LoggingFeature implements Feature {
                 .level(level)
                 .verbosity(verbosity)
                 .maxEntitySize(maxEntitySize)
-                .separator(DEFAULT_SEPARATOR)
         );
 
     }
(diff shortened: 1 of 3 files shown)


nunomsantos:
Hi, @jansupol. Not sure if I need to request a review but I don't see how I can do that in the UI.

senivam (approved):

core-common/src/main/java/org/glassfish/jersey/logging/LoggingFeature.java
```diff
@@ -222,7 +223,6 @@ public LoggingFeature(Logger logger, Level level, Verbosity verbosity, Integer m
                 .level(level)
                 .verbosity(verbosity)
                 .maxEntitySize(maxEntitySize)
-                .separator(DEFAULT_SEPARATOR)
```
jansupol, line 225: The removal of the `DEFAULT_SPEARATOR` is fine here, but should the `configureBuilderParameters` be updated from `builder.separator = builder.separator == null ? filterSeparator : builder.separator;` to `builder.separator = builder.separator == null ? filterSeparator : DEFAULT_SEPARATOR`?

nunomsantos: If we look at some lines above, we can see that the default separator is already considered when the configs are read (same behavior as other configs):
https://github.com/eclipse-ee4j/jersey/blob/9bb2e8dffe2ee307dde151b35f3b98fc81782d4c/core-common/src/main/java/org/glassfish/jersey/logging/LoggingFeature.java#L289-L295

Also, notice that with such change we would always discard what was set in the `builder`.

jansupol (approved):

PR #5007, opened 2022-03-18
Fix regression on LoggingFeature's max entity size

nunomsantos:
Hi all,

PR #4531 implemented the performance improvement requested on #4522 but introduced a regression related to the max entity size property of `LoggingFeature`.
After it, logs from outbound streams are not appended with "...more..." when their entity is trimmed.

The reason is that we were stopping writes to a `ByteArrayOutputStream` too early, so we would never know if it was written in excess.
This looks like a sufficient fix, let me know if you can think of other edge cases.


jansupol (approved):
LGTM, `baos.write(ba, off, len)` does all the checks for `len` and `off`

senivam (approved):

PR #5009, opened 2022-03-18
Depepndency on servlet-api updated to version 6.0.0

dmatej:
Depends on Grizzly snapshot, that is why it is just a draft.


jansupol:
I wonder whether you can drop the Servlet 5 dependencies for the Jetty tests.

dmatej:
> I wonder whether you can drop the Servlet 5 dependencies for the Jetty tests.

I had the same question, but as the build uses just the server as it is and uses jetty client, both 11.0.7, it passed all tests to my surprise. But I still need to rerun everything once more, now I'm back on Glassfish using this branch as in GF's branch I enabled some tests and 2 of them which failed, which caused this avalanche of updates.

Jersey tests did not pass with servlet5, when I updated just grizzly dependency.

jansupol (approved):

senivam (approved):

PR #5016, opened 2022-03-29
Updated CDI, BV, and EL.

jansupol:


jansupol:
All the methods with comment ` // @Override - Removed in CDI 4` are kept for compatibility of Jersey with Glassfish 7.0.0.-M1 and later which do not have CDI updated, yet.

senivam (approved):

PR #5018, opened 2022-03-30
Update Xerces to prevent CVEs

jansupol:


senivam:
is there any drawback of putting the latest (2.12.2) version there? And do we need to raise a CQ for this update?

senivam (approved):

jbescos (approved):

PR #5028, opened 2022-04-04
Enhancement for docprocessor

johnmarquez12:
Current implementation of the `DocProcessor` provides access to types that extend the `Element` class such as `TypeElement`, `ExecutableElement` and `VariableElement`. However extracting javadoc from these is not possible without access to a `DocTrees` object, example shown in usages [here](https://openjdk.java.net/groups/compiler/using-new-doclet.html#tags). By providing access to the `DocletEnvironment` to the `DocProcessor` functions, we can get access to objects that can properly evaluate the javadoc of these elements.

### DIFF
--- a/ext/wadl-doclet/src/main/java12/org/glassfish/jersey/wadl/doclet/DocProcessor.java
+++ b/ext/wadl-doclet/src/main/java12/org/glassfish/jersey/wadl/doclet/DocProcessor.java
@@ -60,9 +60,9 @@ public interface DocProcessor {
      * @param classDoc     the class javadoc
      * @param classDocType the {@link ClassDocType} to extend. This will later be processed by the
      *                     {@link org.glassfish.jersey.server.wadl.WadlGenerator}s.
-     * @param docEnv       the doclet environment used to process the classDoc
      */
-    void processClassDoc(TypeElement classDoc, ClassDocType classDocType, DocletEnvironment docEnv);
+    @Deprecated
+    void processClassDoc(TypeElement classDoc, ClassDocType classDocType);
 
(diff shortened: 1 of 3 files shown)


jansupol:
This one is difficult.

On one hand, it breaks backward compatibility. It changes the `DocProcessor` interface which can be implemented by users, and the user implementation will stop working. I'd like to avoid this.

On the other hand, making the change backward compatible is challenging. Jersey made backward-incompatible changes in the past, but it is not ideal.

Will need to give this more thoughts

johnmarquez12:
> The issue can be with users implementing their own DocProcessor, which is being loaded by the `ResourceDoclet`. One option is to deprecate the methods in `DocProcessor` and create the default methods with the third arg there. The default methods just invoke the deprecated methods. The new implementation of course will override the default methods.
> 
> How ugly does that sound?

Would deprecating the old methods and adding new ones with a third arg in `DocProcessor`, then calling the new methods along with the deprecated ones in `ResourceDoclet` be enough? This would at least give users time to move to the new methods.

jbescos:
> The issue can be with users implementing their own DocProcessor, which is being loaded by the `ResourceDoclet`. One option is to deprecate the methods in `DocProcessor` and create the default methods with the third arg there. The default methods just invoke the deprecated methods. The new implementation of course will override the default methods.
> 
> How ugly does that sound?

It looks ugly for a while, but eventually we can remove the deprecated methods and it will look pretty again.

I'm fine with this solution.

jbescos (approved):
LGTM

jansupol (approved):

Issue #4097, opened 2019-04-11
Blocked thread if WebApplicationException is reused.


kingsfleet:
I think this exception has been raise a few times; but it never seems to come with a reproducer so the scenario is that after a while there is a number of stuck threads noted in this case by WLS 12.2.1; but I can see that this code would be a problem in the servers:

"[ACTIVE] ExecuteThread: '48' for queue: 'weblogic.kernel.Default (self-tuning)'" #157 daemon prio=5 os_prio=0 tid=0x00007f1c849e2000 nid=0x56c7 waiting on condition [0x00007f1c58318000]
   java.lang.Thread.State: WAITING (parking)
	at sun.misc.Unsafe.park(Native Method)
	- parking to wait for  <0x00000007b36845c8> (a jersey.repackaged.com.google.common.util.concurrent.AbstractFuture$Sync)
	at java.util.concurrent.locks.LockSupport.park(LockSupport.java:175)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer.parkAndCheckInterrupt(AbstractQueuedSynchronizer.java:836)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer.doAcquireSharedInterruptibly(AbstractQueuedSynchronizer.java:997)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquireSharedInterruptibly(AbstractQueuedSynchronizer.java:1304)
	at jersey.repackaged.com.google.common.util.concurrent.AbstractFuture$Sync.get(AbstractFuture.java:285)
	at jersey.repackaged.com.google.common.util.concurrent.AbstractFuture.get(AbstractFuture.java:116)
	at org.glassfish.jersey.servlet.internal.ResponseWriter.getResponseContext(ResponseWriter.java:299)
	at org.glassfish.jersey.servlet.internal.ResponseWriter.callSendError(ResponseWriter.java:215)
	at org.glassfish.jersey.servlet.internal.ResponseWriter.commit(ResponseWriter.java:194)
	at org.glassfish.jersey.server.ContainerResponse.close(ContainerResponse.java:413)
	at org.glassfish.jersey.server.ServerRuntime$Responder.writeResponse(ServerRuntime.java:784)
	at org.glassfish.jersey.server.ServerRuntime$Responder.processResponse(ServerRuntime.java:444)
	at org.glassfish.jersey.server.ServerRuntime$Responder.process(ServerRuntime.java:490)
[...]

PR #5035, opened 2022-04-11
Blocked thread if WebApplicationException is reused. #4097

jbescos:
Relates to https://github.com/eclipse-ee4j/jersey/issues/4097

To fix the blocked thread issue, it is only necessary the changes in class ResponseWriter

The other changes are related to fail with 500 when the response is reused more than 1 time.


core-server/src/main/resources/org/glassfish/jersey/server/internal/localization.properties
```diff
@@ -71,6 +71,7 @@ error.resources.cannot.merge=Resources do not have the same path and cannot be m
 error.request.abort.in.response.phase=The request cannot be aborted as it is already in the response processing phase.
 error.request.set.entity.stream.in.response.phase=The entity stream cannot be set in the request as it is already in the response processing phase.
 error.request.set.security.context.in.response.phase=The security context cannot be set in the request as it is already in the response processing phase.
+error.response.already.commited=Response was already committed. This response is being reused more than 1 time.
```
jansupol, line 74: The response entity was already committed. The Response entity is being reused more than once.

tests/e2e-server/src/test/java/org/glassfish/jersey/tests/e2e/server/Issue4097Test.java
```diff
@@ -0,0 +1,68 @@
...
+public class Issue4097Test extends JerseyTest {
+
+    @Override
+    protected Application configure() {
+        return new ResourceConfig(Issue4097Resource.class);
+    }
+
+    @Override
+    protected TestContainerFactory getTestContainerFactory() throws TestContainerException {
+        return new JdkHttpServerTestContainerFactory();
```
jansupol, line 43: The default TestContainerFactory seems to work as well.

jansupol (approved):
The test for `context.isCommitted()` may seem to be too much, we may only restrict the responses with entity (`context.hasEntity() == true` ). But then, the same issue can occur with for instance headers, when the header value is an inputstream for instance. Hence restricting all Responses seems like a proper way to go.

senivam (approved):

PR #5047, opened 2022-04-23
Use Jsonb more effectively

dansiviter:
While debugging I also noticed this issue that was originally spotted some time ago (#3702) . My biggest concern is the creation of an intermediate `String` which for large JSON payloads is inefficient.


senivam (approved):
I would agree with the proposed solution.

jansupol (approved):

Issue #5045, opened 2022-04-21
Netty Connector doesn't support Followredirects.


sumitaneja:
Follow redirect(jersey.config.client.followRedirects)  is true but netty connector still doesn't follow redirect and provides 307 to client.

jbescos:
Hi @sumitaneja I have created an [issue](https://github.com/netty/netty/issues/12332) in Netty to support follow redirects.

We will see whether they can implement this, or we have to implement it in Jersey.

jansupol:
hm... I think Netty supports follow-redirect, at least HelidonConnector based on Netty already implements follow-redirect.

jansupol:
it might be possible that Helidon WebClient supports follow-redirect on its own, without Netty, though...

jbescos:
> it might be possible that Helidon WebClient supports follow-redirect on its own, without Netty, though...

I think Helidon manages that in io.helidon.webclient.RedirectInterceptor

jansupol:
Closed by #5048

PR #5048, opened 2022-04-25
Netty Connector doesn't support Followredirects

jbescos:
Relates to https://github.com/eclipse-ee4j/jersey/issues/5045

### DIFF
--- a/connectors/netty-connector/src/main/java/org/glassfish/jersey/netty/connector/NettyClientProperties.java
+++ b/connectors/netty-connector/src/main/java/org/glassfish/jersey/netty/connector/NettyClientProperties.java
@@ -68,4 +68,18 @@ public class NettyClientProperties {
      * @see javax.net.ssl.SSLParameters#setEndpointIdentificationAlgorithm(String)
      */
     public static final String ENABLE_SSL_HOSTNAME_VERIFICATION = "jersey.config.client.tls.enableHostnameVerification";
+
+    /**
+     * The maximal number of redirects during single request.
+     * <p/>
+     * Value is expected to be positive {@link Integer}. Default value is {@value #DEFAULT_MAX_REDIRECTS}.
+     * <p/>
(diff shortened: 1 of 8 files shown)


jansupol (approved):
Looks good. It would even be better if a redirect loop would have been detected, something like org.glassfish.jersey.jdk.connector.internal.RedirectHandler implements.

jbescos:
Thanks for the comments, I will work on that. I was not aware about the redirect handler, and it looks I should use it.

connectors/netty-connector/src/main/java/org/glassfish/jersey/netty/connector/NettyClientProperties.java
```diff
@@ -68,4 +68,18 @@ public class NettyClientProperties {
...
+     * Value is expected to be positive {@link Integer}. Default value is {@value #DEFAULT_MAX_REDIRECTS}.
+     * <p/>
+     * HTTP redirection must be enabled by property {@link org.glassfish.jersey.client.ClientProperties#FOLLOW_REDIRECTS},
+     * otherwise {@code MAX_REDIRECTS} is not applied.
+     *
+     * @since 2.36
+     * @see org.glassfish.jersey.client.ClientProperties#FOLLOW_REDIRECTS
+     * @see org.glassfish.jersey.netty.connector.internal.RedirectException
+     */
+    public static final String MAX_REDIRECTS = "jersey.config.client.NettyConnectorProvider.maxRedirects";
```
jansupol, line 84: Could you please update the [properties documentation](https://github.com/eclipse-ee4j/jersey/blob/master/docs/src/main/docbook/appendix-properties.xml#L1761)?

jbescos: Yes, lets wait for a green build just in case I did something wrong.

PR #5071, opened 2022-05-27
Update Netty proxy settings

jansupol:
Allow to pass headers to the proxy
Allow to set proxy using the JDK -Dhttp.proxyHost setting


senivam (approved):

connectors/netty-connector/src/main/java/org/glassfish/jersey/netty/connector/NettyConnector.java
```diff
@@ -246,8 +256,29 @@ protected void initChannel(SocketChannel ch) throws Exception {
...
                          InetSocketAddress proxyAddr = new InetSocketAddress(u.getHost(),
                                                                              u.getPort() == -1 ? 8080 : u.getPort());
-                         p.addLast(userName == null ? new HttpProxyHandler(proxyAddr)
-                                                    : new HttpProxyHandler(proxyAddr, userName, password));
+                         ProxyHandler proxy = createProxyHandler(jerseyRequest, proxyAddr, userName, password, connectTimeout);
+                         p.addLast(proxy);
+                     } else {
+                         ProxySelector sel = ProxySelector.getDefault();
+                         for (Proxy proxy: sel.select(requestUri)) {
+                             if (proxy.type().equals(Proxy.Type.HTTP)) {
```
jbescos, line 264: I'm checking the class java.net.Proxy and it has a constructor where the type could be null. Does it makes sense to change that line by any of these?:
`proxy.type() == Proxy.Type.HTTP`
or
`Proxy.Type.HTTP.equals(proxy.type())`

connectors/netty-connector/src/main/java/org/glassfish/jersey/netty/connector/NettyConnector.java
```diff
@@ -246,8 +256,29 @@ protected void initChannel(SocketChannel ch) throws Exception {
...
+                                 SocketAddress proxyAddress = proxy.address();
+                                 if (InetSocketAddress.class.isInstance(proxy.address())) {
+                                     InetSocketAddress proxyAddr = (InetSocketAddress) proxyAddress;
+                                     if (proxyAddr.isUnresolved()
+                                             && proxyAddr.getHostName() != null
+                                             && proxyAddr.getHostName().startsWith("http://")) {
```
jbescos, line 270: Is it possible that starts by https?

jbescos (approved):

connectors/netty-connector/src/main/java/org/glassfish/jersey/netty/connector/NettyConnector.java
jansupol: No, JDK does not support HTTPS proxy

connectors/netty-connector/src/main/java/org/glassfish/jersey/netty/connector/NettyConnector.java
jansupol: Updated

jbescos (approved):

Issue #5072, opened 2022-06-01
Version 3.1.0-M5 was the last version passing the TCK with GlassFish


dmatej:
After we updated to M6, 4 JAXRS TCK tests started failing:
```
[runcts] OUT => [javatest.batch] FAILED........com/sun/ts/tests/jaxrs/platform/container/completioncallback/JAXRSClient.java#argumentContainsExceptionInTwoCallbackClassesTest_from_standalone
[runcts] OUT => [javatest.batch] FAILED........com/sun/ts/tests/jaxrs/platform/container/completioncallback/JAXRSClient.java#argumentContainsExceptionInTwoCallbackInstancesTest_from_standalone
[runcts] OUT => [javatest.batch] FAILED........com/sun/ts/tests/jaxrs/platform/container/completioncallback/JAXRSClient.java#argumentContainsExceptionTest_from_standalone
[runcts] OUT => [javatest.batch] FAILED........com/sun/ts/tests/jaxrs/platform/container/completioncallback/JAXRSClient.java#argumentContainsExceptionWhenSendingIoExceptionTest_from_standalone
```

Each test then produces several stacktraces, this is for one of tests:
```
[2022-06-01T22:52:30.481+0200] [glassfish 7.0] [WARNING] [] [org.glassfish.jersey.server.ServerRuntime$Responder] [tid: _ThreadID=56 _ThreadName=http-listener-1(5)] [timeMillis: 1654116750481] [levelValue: 900] [[
An exception mapping did not successfully produce and processed a response. Logging the exception propagated to the default exception mapper.
java.lang.RuntimeException: java.io.IOException: throw exception
at com.sun.ts.tests.jaxrs.platform.container.completioncallback.ExceptionThrowingStringBean.get(ExceptionThrowingStringBean.java:31)
at com.sun.ts.tests.jaxrs.common.provider.StringBeanEntityProvider.writeTo(StringBeanEntityProvider.java:55)
at com.sun.ts.tests.jaxrs.common.provider.StringBeanEntityProvider.writeTo(StringBeanEntityProvider.java:34)
at org.glassfish.jersey.message.internal.WriterInterceptorExecutor$TerminalWriterInterceptor.invokeWriteTo(WriterInterceptorExecutor.java:242)
at org.glassfish.jersey.message.internal.WriterInterceptorExecutor$TerminalWriterInterceptor.aroundWriteTo(WriterInterceptorExecutor.java:227)
[...]

PR #5073, opened 2022-06-02
Issue #5072 Revert "Make sure the default exception mapper is processed"

dmatej:
- This reverts commit 38e5a53a179b4a26885cc2527312e9953ebfabf8.
- the change caused GlassFish stopped passing 4 tests in the JAXRS TCK:
  jaxrs/platform/container/completioncallback/*

### DIFF
--- a/core-server/src/main/java/org/glassfish/jersey/server/ServerRuntime.java
+++ b/core-server/src/main/java/org/glassfish/jersey/server/ServerRuntime.java
@@ -449,10 +449,9 @@ public class ServerRuntime {
                         request.getResponseWriter().failure(responseError);
                     } finally {
                         completionCallbackRunner.onComplete(responseError);
-
-                        defaultMapperResponse = processResponseWithDefaultExceptionMapper(responseError, request);
                     }
 
+                    defaultMapperResponse = processResponseWithDefaultExceptionMapper(responseError, request);
                 }
 
             } finally {


jansupol:
Odd, is it possible the tests are wrong? The Jakarta REST 3.1 made a change there, will need to check.

dmatej:
I have no idea, but if I would do review for your commit, I would approve it too, seems logical :-)
There is yet some comment in the release method, maybe related:
```
                // Commit the container response writer if not in chunked mode
                // responseContext may be null in case the request processing was cancelled.
```

With your change the GF server.log contained many exceptions, see the issue: https://github.com/eclipse-ee4j/jersey/issues/5072

So it seems it has some consequences which need to be resolved too, but it is complicated somehow ...

senivam:
@dmatej, actually, since the Default Exception Mapper is required (Jakarta REST 3.1 spec) we have to adapt our tests to that change. So, probably some work is required on the GF side as well. Because this new requirement really introduces some inconveniences to developers in terms of changing the existing code :)

senivam:
@dmatej, actually, since the Default Exception Mapper is required (Jakarta REST 3.1 spec) we have to adapt our tests to that change. So, probably some work is required on the GF side as well. Because this new requirement really introduces some inconveniences to developers in terms of changing the existing code :)

dmatej:
You mean updating TCK tests? This change breaks 4 of them, I have no idea what can GlassFish do to fix/workaround these failures when failures go directly from Jersey.
See https://github.com/eclipse-ee4j/glassfish/pull/23964 or those stacktraces in the issue: https://github.com/eclipse-ee4j/jersey/issues/5072

arjantijms:
@senivam did you try passing the TCK with 3.1.0-m6+?

jansupol:
The proposal indeed fixes the CTS test, But it is not a proper fix. 

- First, it does not include the `DefaultExceptionMapper` in cases where an `Exception` is thrown in `request.getResponseWriter().failure(responseError);` 
- Next, the issue is caused by the fact the `completionCallbackRunner.onComplete` is called twice, first by `processResponseWithDefaultExceptionMapper` then by `completionCallbackRunner.onComplete(responseError)`. For the test, the order matters. That is why the revert would work, it ensures the order. But the correct fix is to call `onComplete` just once.

jansupol:
I am about to close this and provide a better fix.

dmatej:
I agree and I would be very happy if you can provide a better fix. I can then test it with GlassFish and TCK locally so we can avoid any redundant releases.

Issue #5089, opened 2022-06-20
Support for proxy


jansupol:
1. Jersey supports ClientProperties.PROXY_URI for Apache, Jetty & Netty client.

For HttpUrlConnector, it is not supported, though [there seems to exist a workaround](https://hiwanglong.github.io/2019/06/how-to-set-proxy-by-jersey-client/). 
Tasks:

- HttpUrlConnector shall use it
- 3.1 jnh-connector shall use it
- [Test](https://github.com/eclipse-ee4j/jersey/blob/master/tests/e2e-client/src/test/java/org/glassfish/jersey/tests/e2e/client/connector/proxy/ProxyTest.java)

2. HttpUrlConnector Supports jdk properties :http.proxyHost and http.proxyPort (ignored for localhost), netty supports that too.
Tasks:

- 3.1 jnh-connector shall use it
- Apache connector shall use it.
- [Test](https://github.com/eclipse-ee4j/jersey/blob/master/tests/e2e-client/src/test/java/org/glassfish/jersey/tests/e2e/client/connector/proxy/ProxySelectorTest.java)

3. Jdk-connector supports proxy. Use it in the test, too.

jbescos:
@jansupol there is one complication with system properties in many connectors. See my comment here: https://github.com/eclipse-ee4j/jersey/pull/5091#issuecomment-1162997477

PR #5091, opened 2022-06-22
Support ClientProperties.PROXY_URI & al in HttpUrlConnector

jbescos:
Relates to https://github.com/eclipse-ee4j/jersey/issues/5089

I don't know what is jnh-connector, Jetty?. That connector is probably not implemented in this PR.

### DIFF
--- a/core-client/src/main/java/org/glassfish/jersey/client/internal/HttpUrlConnector.java
+++ b/core-client/src/main/java/org/glassfish/jersey/client/internal/HttpUrlConnector.java
@@ -319,7 +319,7 @@ public class HttpUrlConnector implements Connector {
         }
     }
 
-    private URI getProxyUri(Object proxy) {
+    private static URI getProxyUriValue(Object proxy) {
         if (proxy instanceof URI) {
             return (URI) proxy;
         } else if (proxy instanceof String) {
@@ -329,16 +329,15 @@ public class HttpUrlConnector implements Connector {
         }
     }
 
-    private String getFromConfigOrSystem(ClientRequest request, String clientProperty, String systemProperty) {
-        String result = request.resolveProperty(clientProperty, String.class);
-        if (result == null) {
-            result = System.getProperty(systemProperty);
-        }
-        return result;
-    }
-
-    private URI getProxyUri(ClientRequest request) {
-        Configuration config = request.getConfiguration();
+    /**
+     * Builds an URI from {@link ClientProperties#PROXY_URI}.
+     * In case it is not set, it will build it from System properties http.proxyHost and http.proxyPort.
+     * Otherwise returns null.
+     *
+     * @param config the configuration containing properties
+     * @return the URI or null
+     */
+    public static URI getProxyUri(Configuration config) {
         Object proxyUri = config.getProperty(ClientProperties.PROXY_URI);
         if (proxyUri == null) {
             String proxyHost = System.getProperty("http.proxyHost");
@@ -347,7 +346,7 @@ public class HttpUrlConnector implements Connector {
                 return URI.create(proxyHost + ":" + proxyPort);
             }
         } else {
-            return getProxyUri(proxyUri);
+            return getProxyUriValue(proxyUri);
         }
         return null;
     }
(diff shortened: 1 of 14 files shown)


senivam:
jnh-connector (JavaNetHttp connector) relates to Jersey 3.1.+

jbescos:
There is one issue with the system property 'http.nonProxyHosts' in some connectors.

Internally every connector has its own client (Apache, Jetty, etc), and it is configured in the constructor of the connector. The proxy is part of this configuration. Then the client is re-used in different requests.

'http.nonProxyHosts' depends on runtime data, because we need to check the current URI of the request in order to enable or disable the proxy, but note that the proxy was already configured before in the constructor.

I don't see a good option to address this. In my opinion there are 2:
1 - Do not support any proxy system properties and delegate that on the implementation (Apache, Jetty, etc). In case they don't support it, we don't support it.
2 - In case we really must support this, we could create 2 clients in every connector, one with the proxy and other without proxy. In runtime we select the proper one comparing the host in http.nonProxyHosts with the current request URI.

In my opinion we should go for option 1, because option 2 is memory and performance inefficient, and the different implementations can deal with this with no problem because they have access to system properties.

core-client/src/main/java/org/glassfish/jersey/client/internal/HttpUrlConnector.java
```diff
@@ -314,10 +321,34 @@ protected void secureConnection(final JerseyClient client, final HttpURLConnecti
...
-        uc = this.connectionFactory.getConnection(request.getUri().toURL());
+        Proxy proxy = null;
+        Object proxyUri = request.getConfiguration().getProperties().get(ClientProperties.PROXY_URI);
+        if (proxyUri != null) {
+            URI uri = getProxyUriValue(proxyUri);
+            String username = ClientProperties.getValue(request.getConfiguration().getProperties(),
+                    ClientProperties.PROXY_USERNAME, ExternalProperties.HTTP_PROXY_USER);
+            String password = ClientProperties.getValue(request.getConfiguration().getProperties(),
+                    ClientProperties.PROXY_PASSWORD, ExternalProperties.HTTP_PROXY_PASSWORD);
+            if (username != null && password != null) {
```
senivam, line 344: I think a password is not a must, username presence should be enough.

senivam (changes requested):

tests/integration/externalproperties/src/test/java/org/glassfish/jersey/tests/externalproperties/HttpProxyTest.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2020, 2022 Oracle and/or its affiliates. All rights reserved.
```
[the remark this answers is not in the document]
jbescos: The plugin is forcing me to add it, because in the list of commits I did a modification.

It doesn't matter, lets make the plugin happy.

senivam (approved):

core-client/src/main/java/org/glassfish/jersey/client/internal/HttpUrlConnector.java
```diff
@@ -314,10 +321,37 @@ protected void secureConnection(final JerseyClient client, final HttpURLConnecti
...
+            throw new ProcessingException(LocalizationMessages.WRONG_PROXY_URI_TYPE(ClientProperties.PROXY_URI));
+        }
+    }
+
     private ClientResponse _apply(final ClientRequest request) throws IOException {
         final HttpURLConnection uc;
-
-        uc = this.connectionFactory.getConnection(request.getUri().toURL());
+        Proxy proxy = null;
+        Object proxyUri = request.getConfiguration().getProperties().get(ClientProperties.PROXY_URI);
```
jansupol, line 337: Would it be better to use `request.resolveProperty` to enable proxy settings in the filters?

jbescos: It is done in this way in other connectors too:
`request.getConfiguration().getProperties().get(ClientProperties.PROXY_URI);`
It looks the property could be a String or an URI. I am not sure why but I think it is safer to keep it consistent with other connectors.

PR #5095, opened 2022-07-08
Move Proxy parsing logic from connectors to a common class

jansupol:


jansupol:
- leave the resolution of the request hostname on the proxy
- Put ClientProxy class to innate internal package for the Felix plugin to export the package - makes the package visible to connector modules.
- Add a test for Grizzly connector using a proxy

jbescos (approved):

PR #5102, opened 2022-07-20
FISH-5723 Fixes WebappClassloader memory-leak issue by removing JAXRSResources classes from cache on shutdown event

jGauravGupta:

### DIFF
--- a/core-common/src/main/java/org/glassfish/jersey/internal/util/collection/Cache.java
+++ b/core-common/src/main/java/org/glassfish/jersey/internal/util/collection/Cache.java
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2017, 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2022 Payara Foundation and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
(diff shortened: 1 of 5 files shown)


jansupol:
@jGauravGupta Thank you for the PR. I have two questions regarding the PR (perhaps I have a wrong understanding of `@BeforeShutdown`):

1. The cache purge is done on `@BeforeShutdown`, i.e. when Weld container is being shutdown. Does the Cache gets not garbage collected when shutdown?
2. The `CdiComponentProvider` which holds the `CdiComponentProviderRuntimeSpecifics` is a CDI Extension, and Weld container keeps only a single instance of the class. (Well, there is another instance, not in Weld, but assumingly it has an empty Cache). Why do you purge the cache only for a single thread, instead of `Cache.clear()`?

jGauravGupta:
Hello @jansupol ,

This issue is reproducible for both the Payara and Glassfish servers if the web application contains JAXRS resources.

Yes, the cache is not garbage collected by default on application un-deployment which holds the JAXRS resource classes reference of all deployed applications hence WebappClassloader is not garbage collected. [@BeforeShutdown callback](https://github.com/weld/core/blob/5.0.1.Final/impl/src/main/java/org/jboss/weld/bootstrap/WeldRuntime.java#L77) in the extension is invoked on the un-deployment of each application and removes the specific JAX-RS resource classes which belong to currently undeploying application classloader instead of 'Cache.clear()'.

jansupol (approved):

jansupol:
@jGauravGupta Thanks!

PR #5114, opened 2022-08-04
Change JettyConnector 'readTimeout' behavior to match socket read tim…

daijithegeek:
…eout definition - e.g., ApacheConnector behavior matches it.

* Read timeout: Time on waiting to receive the first data byte.

The `timeout` method timeouts the request even if data were already received, capping the query to a maximum execution time.
This behavior is problematic when streaming data over a prolonged duration; the client has already received data bytes, but data continues to flow.

I provided a jetty specific property (jersey.config.jetty.client.totalTimeout) that configures the 'totalTimeout' when required.

### DIFF
--- a/connectors/jetty-connector/src/main/java/org/glassfish/jersey/jetty/connector/JettyConnector.java
+++ b/connectors/jetty-connector/src/main/java/org/glassfish/jersey/jetty/connector/JettyConnector.java
@@ -314,8 +314,14 @@ class JettyConnector implements Connector {
         request.followRedirects(clientRequest.resolveProperty(ClientProperties.FOLLOW_REDIRECTS, true));
         final Object readTimeout = clientRequest.resolveProperty(ClientProperties.READ_TIMEOUT, -1);
         if (readTimeout != null && readTimeout instanceof Integer && (Integer) readTimeout > 0) {
-            request.timeout((Integer) readTimeout, TimeUnit.MILLISECONDS);
+            request.idleTimeout((Integer) readTimeout, TimeUnit.MILLISECONDS);
         }
+
+        final Object totalTimeout = clientRequest.resolveProperty(JettyClientProperties.TOTAL_TIMEOUT, -1);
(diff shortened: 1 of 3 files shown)


connectors/jetty-connector/src/main/java/org/glassfish/jersey/jetty/connector/JettyConnector.java
```diff
@@ -314,8 +314,14 @@ private Request translateRequest(final ClientRequest clientRequest) {
         request.followRedirects(clientRequest.resolveProperty(ClientProperties.FOLLOW_REDIRECTS, true));
         final Object readTimeout = clientRequest.resolveProperty(ClientProperties.READ_TIMEOUT, -1);
         if (readTimeout != null && readTimeout instanceof Integer && (Integer) readTimeout > 0) {
-            request.timeout((Integer) readTimeout, TimeUnit.MILLISECONDS);
+            request.idleTimeout((Integer) readTimeout, TimeUnit.MILLISECONDS);
```
jbescos, line 317: Javadoc does not say much about this, but I think this does not exactly have the same meaning than read time out. This method should also have impact in writing (not only read).

However there is no such method to set a read timeout, and this is more correct than before, so it looks good to me.

jbescos (approved):
LGTM

connectors/jetty-connector/src/main/java/org/glassfish/jersey/jetty/connector/JettyConnector.java
daijithegeek: > Javadoc does not say much about this, but I think this does not exactly have the same meaning than read time out. > This method should also have impact in writing (not only read).

You are right. the timeout may fire when there is a long enough pause during data transfer. I don't know to be honest.
As you say, the situation seems better like this anyway.

jansupol (approved):
Looks ok. The "idleTimeout" does not seem to be exactly the same as "readTimeout", but the behaviour looks similar.

PR #5124, opened 2022-08-10
Support JDK19

jbescos:
```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  53:02 min
[INFO] Finished at: 2022-08-10T13:31:12+02:00
[INFO] ------------------------------------------------------------------------

jbescos@jbescos-TECRA-X40-E:~/workspace/jersey$ mvn --version
Apache Maven 3.6.3 (cecedd343002696d0abb50b32b541b8a6ba2883f)
Maven home: /home/jbescos/programs/apache-maven-3.6.3
Java version: 19-ea, vendor: Oracle Corporation, runtime: /home/jbescos/programs/java/jdk-19
Default locale: en, platform encoding: UTF-8
OS name: "linux", version: "5.4.0-107-generic", arch: "amd64", family: "unix"
```

### DIFF
--- a/tests/integration/property-check/src/test/java/org/glassfish/jersey/internal/config/SystemPropertiesConfigurationModelTest.java
+++ b/tests/integration/property-check/src/test/java/org/glassfish/jersey/internal/config/SystemPropertiesConfigurationModelTest.java
@@ -44,7 +44,6 @@ import org.glassfish.jersey.server.ServerProperties;
 import org.glassfish.jersey.server.oauth1.OAuth1ServerProperties;
 import org.glassfish.jersey.servlet.ServletProperties;
 import org.glassfish.jersey.test.TestProperties;
-import org.junit.Ignore;
 import org.junit.Test;
 
 public class SystemPropertiesConfigurationModelTest {
@@ -78,7 +77,6 @@ public class SystemPropertiesConfigurationModelTest {
     }
 
     @Test
-    @Ignore("System.setSecurityManager is deprecated in JDK 19 and there is no replacement")
     public void propertyLoadedWhenSecurityException() {
         final String TEST_STRING = "test";
         SecurityManager sm = System.getSecurityManager();
--- a/tests/integration/property-check/src/test/java/org/glassfish/jersey/internal/config/SystemPropertiesConfigurationModelTest.java
+++ b/tests/integration/property-check/src/test/java/org/glassfish/jersey/internal/config/SystemPropertiesConfigurationModelTest.java
@@ -44,6 +44,7 @@ import org.glassfish.jersey.server.ServerProperties;
 import org.glassfish.jersey.server.oauth1.OAuth1ServerProperties;
 import org.glassfish.jersey.servlet.ServletProperties;
 import org.glassfish.jersey.test.TestProperties;
+import org.junit.Ignore;
 import org.junit.Test;
 
 public class SystemPropertiesConfigurationModelTest {
@@ -77,6 +78,7 @@ public class SystemPropertiesConfigurationModelTest {
     }
 
     @Test
+    @Ignore("System.setSecurityManager is deprecated in JDK 19 and there is no replacement")
     public void propertyLoadedWhenSecurityException() {
         final String TEST_STRING = "test";
         SecurityManager sm = System.getSecurityManager();
--- a/etc/jenkins/Jenkinsfile_ci_build
+++ b/etc/jenkins/Jenkinsfile_ci_build
@@ -32,12 +32,12 @@ pipeline {
                             '''
                     }
                 }
-                stage('JDK 16 ') {
+                stage('JDK 19 ') {
                     agent {
                         label 'centos-7'
                     }
                     tools {
-                        jdk 'openjdk-jdk16-latest'
+                        jdk 'openjdk-jdk19-latest'
                         maven 'apache-maven-latest'
                     }
                     steps {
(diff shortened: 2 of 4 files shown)


tests/integration/property-check/src/test/java/org/glassfish/jersey/internal/config/SystemPropertiesConfigurationModelTest.java
```diff
@@ -77,6 +78,7 @@ public void allPropertyClassLoaded() throws IOException {
     }
 
     @Test
+    @Ignore("System.setSecurityManager is deprecated in JDK 19 and there is no replacement")
```
senivam, line 81: good, but this can be ignored in a profile (pom.xml) designed solely for JDK 19 (while still working for other JDKs).

senivam:
org.apache.groovy:groovy-all: [CQ24207](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=24207)

tests/integration/property-check/src/test/java/org/glassfish/jersey/internal/config/SystemPropertiesConfigurationModelTest.java
jansupol: Yep, this should have been excluded just for JDK 19+

jbescos: Right, fixed

jansupol:
My JDK:
```java -version
java version "19.0.1"
Java(TM) SE Runtime Environment (build 19.0.1+5-12)
Java HotSpot(TM) 64-Bit Server VM (build 19.0.1+5-12, mixed mode, sharing)
```
And the result:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.0:testCompile (default-testCompile) on project jersey-container-grizzly2-http: Compilation failure: Compilation failure:
[ERROR] /jersey/containers/grizzly2-http/src/test/java/org/glassfish/jersey/grizzly2/httpserver/test/tools/JerseyHttpClientThread.java:[53,15] java.lang.Thread.Builder is a preview API and is disabled by default.
[ERROR]   (use --enable-preview to enable preview APIs)
[ERROR] /jersey/containers/grizzly2-http/src/test/java/org/glassfish/jersey/grizzly2/httpserver/test/tools/JerseyHttpClientThread.java:[53,45] incompatible types: jakarta.ws.rs.client.Invocation.Builder cannot be converted to java.lang.Thread.Builder
[ERROR] /jersey/containers/grizzly2-http/src/test/java/org/glassfish/jersey/grizzly2/httpserver/test/tools/JerseyHttpClientThread.java:[54,42] cannot find symbol
[ERROR]   symbol:   method get()
[ERROR]   location: variable builder of type java.lang.Thread.Builder
```

Is it an odd JDK bug?

jbescos:
> /jersey/containers/grizzly2-http/src/test/java/org/glassfish/jersey/grizzly2/httpserver/test/tools/JerseyHttpClientThread.java

There is no such `/jersey/containers/grizzly2-http/src/test/java/org/glassfish/jersey/grizzly2/httpserver/test/tools/JerseyHttpClientThread.java` file, right?. Is it a local change that you could have?.

senivam:
@jbescos, it appears since 3.x - added a year ago by Payara guys. But when your PR is merged into the 3.x, this will become an issue :) however, this is very probably a bug in JDK not in your (or theirs) code.

jbescos:
> @jbescos, it appears since 3.x - added a year ago by Payara guys. But when your PR is merged into the 3.x, this will become an issue :) however, this is very probably a bug in JDK not in your (or theirs) code.

It looks adding this --enable-preview should pass. We can also wait for a final JDK19 version to try it.

jansupol:
JerseyHttpClientThread has:
```
        final WebTarget path = client.target(getSettings().targetUri.toString());
        final Builder builder = path.request();
```

But the compiler for some reason thinks the Builder is `Thread.Builder` instead of `Invocation.Builder`....

jbescos:
> JerseyHttpClientThread has:
> 
> ```
>         final WebTarget path = client.target(getSettings().targetUri.toString());
>         final Builder builder = path.request();
> ```
> 
> But the compiler for some reason thinks the Builder is `Thread.Builder` instead of `Invocation.Builder`....

Yes, I noticed that and it makes no sense. There is no point to fail with that because the signature of the method clearly returns `Invocation.Builder`. That must be a bug.

We can wait for a final JDK19 release and compile it. Is there a release date for that?.

jansupol:
After a discussion with the JDK team: The issue is that `JerseyHttpClientThread` extends `Thread` and hence the `Builder` is the `Thread.Builder`. So it is enough to explicitly say:
```
 final Invocation.Builder builder = path.request();
```

jbescos:
> After a discussion with the JDK team: The issue is that `JerseyHttpClientThread` extends `Thread` and hence the `Builder` is the `Thread.Builder`. So it is enough to explicitly say:
> 
> ```
>  final Invocation.Builder builder = path.request();
> ```

I see. Anyway that is for 3.x branch because in master there is no JerseyHttpClientThread class.

I can create a separate PR to make that change to prevent this issue will happen once we migrate master changes to 3.x:
`final Invocation.Builder builder = path.request();`

jansupol (approved):

senivam (approved):

PR #5137, opened 2022-08-29
Fix test issues

jbescos:
I was not able to reproduce it, but it looks some users are having issues with a couple of tests.

@jansupol it would be good if you can run both tests and provide me feedback about the results.


tests/integration/jersey-4697/src/test/java/org/glassfish/jersey/tests/integration/jersey4697/MonitoringEventListenerTest.java
```diff
@@ -16,13 +16,17 @@
 
 package org.glassfish.jersey.tests.integration.jersey4697;
 
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.when;
+import static org.junit.Assert.assertEquals;
```
senivam, line 19: In the latest version of the PR there are many redundant imports. Could you please clean them out? 

And those imports, which are already in place (before any changes) should not be reformatted because when it comes to merges, reformatting brings unnecessary conflicts.

senivam (changes requested):
see notes in the code

tests/integration/jersey-4697/src/test/java/org/glassfish/jersey/tests/integration/jersey4697/MonitoringEventListenerTest.java
jbescos: Good point

senivam (approved):

jansupol (approved):

PR #5143, opened 2022-09-01
Merge 3.0->3.1 (Jdk 19 + cdi-rs-inject OSGi)

jansupol:


pom.xml
```diff
@@ -2237,6 +2237,7 @@
         <gf.impl.version>7.0.0-M2</gf.impl.version>
         <!-- Jakartified -->
         <cdi.api.version>4.0.1</cdi.api.version>
+        <cdi.osgi.version>javax.enterprise.*;version="[3.0,5)"</cdi.osgi.version>
```
senivam, line 2240: jakarta.enterprise

senivam (changes requested):

pom.xml
senivam: BTW, javax.annotation.osgi.version in the incubator/cdi-inject-weld/pom.xml shall be jakarta.annotation.osgi.version as well

senivam: Unfortunately, this comes from the master through 3.x to 3.1 thus shall be fixed in the 3.x as well...

senivam (approved):

PR #5160, opened 2022-09-15
Fixes for: API docs, user-guide, examples bundle

senivam:

### DIFF
--- a/connectors/helidon-connector/pom.xml
+++ b/connectors/helidon-connector/pom.xml
@@ -42,7 +42,7 @@
         <dependency>
             <groupId>io.helidon.jersey</groupId>
             <artifactId>helidon-jersey-connector</artifactId>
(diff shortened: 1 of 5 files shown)


connectors/helidon-connector/pom.xml
```diff
@@ -42,7 +42,7 @@
         <dependency>
             <groupId>io.helidon.jersey</groupId>
             <artifactId>helidon-jersey-connector</artifactId>
-            <version>3.0.0-M1</version>
+            <version>3.0.1</version>
```
jansupol, line 45: We do not have CQ for 3.0.1

senivam: [CQ24261](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=24261)

jansupol (approved):

PR #5163, opened 2022-09-19
Update JdkHttpServerFactory.java

### LINKED ISSUE #3383: jersey-container-jdk-http ignores host and always binds to wildcard address
When using the jersey-container-jdk-http to create a HttpServer, the configuration is passed via an URI, from which protocol, port and path is extracted and used.
Yet, it is not possible to bind the server to the localhost/loopback interface. The HttpServer is always bound to the wildcard address via <tt>HttpServer.create(new InetSocketAddress(port), 0)</tt>

I would prefer to be able to manually bind the HttpServer at a later time, there should be no need to bind the HttpServer at construction. For the constructor, isHttps and path should be sufficient.

If this change is too complicated, it would help to actually use the host part of the URI for the InetSocketAddress
#### Affected Versions
[2.22.2]

Matthiasvanderhallen:
The hostname part of the URI is ignored by JDKHttpServerFactory. Fixes #3383


jansupol:
Please update copyright year

jansupol (approved):

Issue #5169, opened 2022-09-23
Create HTTPS tests with JNH connector


jansupol:
The current stack trace is 
java.lang.NullPointerException: Cannot invoke "java.io.InputStream.read(byte[], int, int)" because "this.input" is null

	at org.glassfish.jersey.message.internal.EntityInputStream.read(EntityInputStream.java:79)
	at org.glassfish.jersey.message.internal.ReaderInterceptorExecutor$UnCloseableInputStream.read(ReaderInterceptorExecutor.java:273)
	at java.base/sun.nio.cs.StreamDecoder.readBytes(StreamDecoder.java:270)
	at java.base/sun.nio.cs.StreamDecoder.implRead(StreamDecoder.java:313)
	at java.base/sun.nio.cs.StreamDecoder.read(StreamDecoder.java:188)
	at java.base/java.io.InputStreamReader.read(InputStreamReader.java:177)
	at java.base/java.io.Reader.read(Reader.java:248)
	at org.glassfish.jersey.message.internal.ReaderWriter.readFromAsString(ReaderWriter.java:152)
	at org.glassfish.jersey.message.internal.ReaderWriter.readFromAsString(ReaderWriter.java:137)
	at org.glassfish.jersey.message.internal.AbstractMessageReaderWriterProvider.readFromAsString(AbstractMessageReaderWriterProvider.java:94)
	at org.glassfish.jersey.message.internal.StringMessageProvider.readFrom(StringMessageProvider.java:54)
	at org.glassfish.jersey.message.internal.StringMessageProvider.readFrom(StringMessageProvider.java:36)
	at org.glassfish.jersey.message.internal.ReaderInterceptorExecutor$TerminalReaderInterceptor.invokeReadFrom(ReaderInterceptorExecutor.java:233)
	at org.glassfish.jersey.message.internal.ReaderInterceptorExecutor$TerminalReaderInterceptor.aroundReadFrom(ReaderInterceptorExecutor.java:212)
	at org.glassfish.jersey.message.internal.ReaderInterceptorExecutor.proceed(ReaderInterceptorExecutor.java:132)
	at org.glassfish.jersey.message.internal.MessageBodyFactory.readFrom(MessageBodyFactory.java:1072)
	at org.glassfish.jersey.message.internal.InboundMessageContext.readEntity(InboundMessageContext.java:888)
	at org.glassfish.jersey.message.internal.InboundMessageContext.readEntity(InboundMessageContext.java:822)
[...]

PR #5170, opened 2022-09-27
Create HTTPS tests with JNH connector #5169

jbescos:
Relates to https://github.com/eclipse-ee4j/jersey/issues/5169

Test passes successfully.


jansupol (approved):

jbescos:
I think I need to rebase it from master and change some junit imports

jansupol:
There would be more PRs like that, we can fix it later, or you can fix it if you want

jbescos:
I will do it in all my PRs now, so there will be less work to do later

jbescos:
I am going to leave it as it is, because there are many conflicts coming from classes that I didn't touch

PR #5176, opened 2022-10-12
Fix support @Inject for JAX-RS & Servlet beans

jansupol:


jansupol:
- Prevent the wrong CDI Extension execution order of the CdiComponentProvider extension and InjectExtension
- Provide `@JerseyContext` qualifier for CDI beans for Servlet classes to be used with `@Inject`

jansupol:
replaced by #5178

PR #5194, opened 2022-11-14
Fix optionals when empty

jbescos:
There was one issue with optionals when the parameter was empty.


jansupol:
It would be interesting to test `@NotNull Integer nullable` and `@NotNull Optional<Integer> nullable`. While the first should fail with bean validation, I'd say the second should not.

jbescos:
> It would be interesting to test `@NotNull Integer nullable` and `@NotNull Optional<Integer> nullable`. While the first should fail with bean validation, I'd say the second should not.

Added in this commit: https://github.com/eclipse-ee4j/jersey/pull/5194/commits/84198e39cbd6d06ded4f44c4cbf12292a54a795e

jansupol (approved):

PR #5198, opened 2022-11-16
SSLHandshake No appropriate protocol (protocol is disabled or cipher suites are inappropriate)

jbescos:
TLS 1.0 and TLS 1.1 are deprecated, and in some updated JDKs tests are failing.


senivam:
I would advise checking if those protocols are supported in the given environment, something like 
```
final SSLContext context = SSLContext.getInstance("TLS");
        context.init(null, null, null);
        final String[] supportedProtocols = context.getDefaultSSLParameters().getProtocols();
        if (Arrays.toString(supportedProtocols).contains("TLSv1.1")) {
           System.setProperty("jdk.tls.server.protocols", "TLSv1.1");
           System.setProperty("jdk.tls.client.protocols", "TLSv1.1");
        }
```

jbescos:
> I would advise checking if those protocols are supported in the given environment, something like
> 
> ```
> final SSLContext context = SSLContext.getInstance("TLS");
>         context.init(null, null, null);
>         final String[] supportedProtocols = context.getDefaultSSLParameters().getProtocols();
>         if (Arrays.toString(supportedProtocols).contains("TLSv1.1")) {
>            System.setProperty("jdk.tls.server.protocols", "TLSv1.1");
>            System.setProperty("jdk.tls.client.protocols", "TLSv1.1");
>         }
> ```

Done

senivam (approved):

jansupol (approved):

PR #5202, opened 2022-11-22
Brotli support

tvallin:
Draft PR because brotli is not released yet.


jansupol:
I believe we cannot make Brotli a default dependency. Rather, it would reside in a separate module. While brotli impl is in a testing phase, I'd see the module under the incubator, until some maturity is reached.
Then we would move it under `media`.
The package name ought to be the same under incubator and media, so please consider it while naming the package.

incubator/brotli/src/main/java/org/glassfish/jersey/message/BrotliEncoder.java
```diff
@@ -0,0 +1,37 @@
+package org.glassfish.jersey.message;
```
jansupol, line 1: I do not believe it is a good idea to have the same package name in two modules, and `org.glassfish.jersey.message` is in core-common already.

PR #5205, opened 2022-11-30
Add caching and improve performance

jansupol:


jansupol:
Compared to #5204, the performance tests proved the changes in 5204 MediaTypeProvider as contra productive and the reworked HttpHeaderReader improves the performance even better.

senivam (approved):

PR #5208, opened 2022-12-02
Fix FORM_PARAM_CONSUMED warning

dosy4ev:
When request `Content-type` header equals to
`application/x-www-form-urlencoded` and request contains query params
then query params will be treated as form params.  And jersey will
print a FORM_PARAM_CONSUMED warning log message.

This behavior should be configured by
`jersey.config.servlet.form.queryParams.disabled` property.

With this property set to `true` query params are not converted to form
params but due to bug FORM_PARAM_CONSUMED warning is still printed.

### DIFF
--- a/containers/jersey-servlet-core/src/main/java/org/glassfish/jersey/servlet/WebComponent.java
+++ b/containers/jersey-servlet-core/src/main/java/org/glassfish/jersey/servlet/WebComponent.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2012, 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2012, 2022 Oracle and/or its affiliates. All rights reserved.
  *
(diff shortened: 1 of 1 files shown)


jansupol (approved):

jbescos (approved):

EddyKr:
We're seeing the warning when we think we should not see it.
There is an endpoint that only contains @FormParam annotations and we don't send in a request body. It's not clear what are we doing wrong as we would like to get rid of this message, besides just filtering it out.
There is a workaround that if we would add any query parameter and enable the `jersey.config.servlet.form.queryParams.disabled ` then the message would not be printed because of `final boolean keepQueryParams = queryParamsAsFormParams || queryParams.isEmpty();` since then the query params would not be empty.
Maybe we are doing something wrong and the warning message could be avoided. Any kind of guidance would be appreciated.

PR #5209, opened 2022-12-07
Merge JUnit 4 -> 5 bump into origin/3.x

senivam:


jansupol (approved):
I have checked the result and it looks good.
The difference I found is that org.glassfish.jersey.server.ResourceConfigBuilderTest
did not make it from 2.x to 3.x some time ago

PR #5237, opened 2023-01-20
Don't Specify Wildcard Package OSGi Version for MicroProfile Rest Client

Pandrex247:
Not sure if this is the current "HEAD" branch or not - I'm trying to patch Jersey 3.1.

The OSGi package versions for MicroProfile Rest Client don't conform to the project version.
In the Rest Client 3.0 API jar, the packages are exported as such:
* `org.eclipse.microprofile.rest.client;version="2.0"`
* `org.eclipse.microprofile.rest.client.annotation;version="1.1.1"`
* `org.eclipse.microprofile.rest.client.ext;version="2.0"`
* `org.eclipse.microprofile.rest.client.inject;version="2.0"`
* `org.eclipse.microprofile.rest.client.spi;version="1.1.1"`

And Jersey with its current configuration imports the following:
* `org.eclipse.microprofile.rest.client;version="[3.0,4)"`
* `org.eclipse.microprofile.rest.client.annotation;version="[3.0,4)"`
* `org.eclipse.microprofile.rest.client.ext;version="[3.0,4)"`
* `org.eclipse.microprofile.rest.client.inject;version="[3.0,4)"`
* `org.eclipse.microprofile.rest.client.spi;version="[3.0,4)"`

This change should make it so the imports are as follows:
* `org.eclipse.microprofile.rest.client;version="[2.0,3)"`
* `org.eclipse.microprofile.rest.client.annotation;version="[1.1,2)"`
* `org.eclipse.microprofile.rest.client.ext;version="[2.0,3)"`
* `org.eclipse.microprofile.rest.client.inject;version="[2.0,3)"`
* `org.eclipse.microprofile.rest.client.spi;version="[1.1,2)"`

Signed-off-by: Andrew Pielage <pandrex247@hotmail.com>

### DIFF
--- a/ext/microprofile/mp-rest-client/pom.xml
+++ b/ext/microprofile/mp-rest-client/pom.xml
@@ -114,7 +114,6 @@
                         <Import-Package>
                             ${cdi.osgi.version},
                             jakarta.decorator.*;version="[3.0,5)",
-                            org.eclipse.microprofile.rest.client.*;version="[3.0,4)",
                             org.eclipse.microprofile.config.*;version="!",
                             *
                         </Import-Package>
--- a/ext/microprofile/mp-rest-client/pom.xml
+++ b/ext/microprofile/mp-rest-client/pom.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
-    Copyright (c) 2019, 2022 Oracle and/or its affiliates. All rights reserved.
+    Copyright (c) 2019, 2023 Oracle and/or its affiliates. All rights reserved.
 
     This program and the accompanying materials are made available under the
     terms of the Eclipse Public License v. 2.0, which is available at


jansupol:
This should have been fixed in the rest client API I believe. But until then, you are right the felix plugin should set the version the API provides. It is doing so when the line is removed (as you did in the PR). But, If they ever fix that in the MP Rest Client API, the module won't work with it.

So I would rather set `org.eclipse.microprofile.rest.client.*;version="[2.0,4)"`, what do you think?

pdudits:
Last time I was questioning that in 2019 in eclipse/microprofile-rest-client#177. I understood they want to retain per-package semantic version, so we should consider that a feature of the API.

Pandrex247:
For reference, when they released Rest Client 3.0 for Jakarta EE they made the effort to increment all of the package versions with major or minor increments to the above so it's not an accident that they are what they are - it seems that they wish to continue versioning packages separately. 

[Reference Commit](https://github.com/eclipse/microprofile-rest-client/commit/a1011eced8f02c25726de04493de22a23ca6d2a0#diff-e3338953951a6735a2fba4ecc85e800b6735c2ab486dfdd636f00df90b0f416a)

jansupol (approved):
The package-versioning sounds like extra confusion for the customers, but I agree with letting the versioning be setup automatically by the Felix plugin

PR #5240, opened 2023-01-24
Proxy client enhancements

zUniQueX:
This PR provides several enhancements to the `proxy-client` module:
 - address deprecation of `AccessController`: The `AccessController` class is deprecated since Java 17 and marked for removal. The proposed change tries to use an `AccessController` if possible and uses a fallback otherwise.
 - parameterize raw generic types: Some generic types were used without generic type arguments and therefore produce warnings. Adding `<?>` as generic type argument fixes this problem.
 - change proxy library: `java.lang.reflect.Proxy` can proxy simple interfaces only. ByteBuddy provides a more sophisticated proxying mechanism and can proxy arbitrary types. This allows for example the execution of custom code in the interface's default methods since these aren't forwarded to the proxy any more.
 - add client instrumentations: Currently the client builds its request without the possibility to modify it. Sometimes, there is the need to modify the request (e.g. for adding dynamic query parameters). The `InstrumentedProxyClient` interface provides a SPI to modify the `WebTarget` and the `Invocation.Builder` if necessary.

### DIFF
--- a/ext/proxy-client/pom.xml
+++ b/ext/proxy-client/pom.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 <!--
 
-    Copyright (c) 2012, 2022 Oracle and/or its affiliates. All rights reserved.
+    Copyright (c) 2012, 2023 Oracle and/or its affiliates. All rights reserved.
 
     This program and the accompanying materials are made available under the
     terms of the Eclipse Public License v. 2.0, which is available at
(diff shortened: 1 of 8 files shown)


jansupol:
Thank you for the PR. I do not think we should need ByteBuddy dependency, though. Can it do something the JDK Proxy is not capable of?

jansupol:
I am really not sure what this PR tries to accomplish. It looks like the primary reason is to support `InstrumentedProxyClient`. Jersey contains other mechanisms to accomplish what this PR tries to solve, I believe. It is the [InvocationBuilderListener](https://eclipse-ee4j.github.io/jersey.github.io/documentation/latest/client.html#d0e5745) and [ClientBuilderListener](https://github.com/eclipse-ee4j/jersey/blob/master/core-client/src/main/java/org/glassfish/jersey/client/spi/ClientBuilderListener.java). 

If these two won't suit your needs, we can discuss some solutions, but preferably it should be a solution that not only the proxy client benefits from, but also the regular client would potentially use it.

I do not feel `ByteBuddy` is something that should have been brought to Jersey at the moment. We already have ASM, we have javassist, we have HK2 proxies, and we have CDI/Weld proxies. I do not feel a strong reason for bringing in another guy. No matter the latest Weld relies on it.

zUniQueX:
My main goal with this PR was to support instrumentation for the `WebTarget`. As far as I can see, the Jersey SPI doesn't support something for this at the moment.

The main problem to solve is that some APIs provide some kind of 'context-dependent' query parameters. That means you can't use a `@QueryParam` annotation with a fixed key, but rather need a way of adding a dynamic query parameter at runtime. For the classic client API, this isn't a problem, as you have direct access to the `WebTarget`, where you can manually add the query param. On the server side however, there isn't the need for an extension either, since the server can utilize an `UriInfo` to retrieve all query parameters as a `MultivaluedMap`. Only the proxy client can't use a map to set the query parameters and therefore one would have to modify the global `WebTarget` it is initially created with. So my main goal was to allow the customization of the `WebTarget` for each request individually.

I don't think it's a good idea to provide an SPI interface for this case, since the `WebTarget` is newly created on each modification. As you've already pointed out, the instrumentation handler could be registered in the `WebResourceFactory`. This would also make it obsolete to use an other proxy library.

PR #5241, opened 2023-01-25
Add SNI Support based on Host header

jansupol:

### DIFF
--- a/connectors/apache-connector/src/main/java/org/glassfish/jersey/apache/connector/ApacheConnector.java
+++ b/connectors/apache-connector/src/main/java/org/glassfish/jersey/apache/connector/ApacheConnector.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2022 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2010, 2023 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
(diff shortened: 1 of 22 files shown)


connectors/apache-connector/src/main/java/org/glassfish/jersey/apache/connector/ApacheConnector.java
```diff
@@ -821,4 +829,55 @@ protected OutputStream createOutputStream(final long len, final SessionOutputBuf
...
+
+        @Override
+        public Socket createLayeredSocket(
+                final Socket socket,
+                final String target,
+                final int port,
+                final HttpContext context) throws IOException {
+            httpContexts.set(context);
+            Socket superSocket = super.createLayeredSocket(socket, target, port, context);
+            httpContexts.remove();
```
jbescos, line 864: Better to use finally block here in case there is an exception?

jansupol: Why not. The cleanup is not necessary, though. The information won't be used by any other thread, and the failing thread will reset the value with a new one for the next connection.

jbescos: Ok, I thought the thread came from a thread pool

senivam (approved):

jbescos (approved):

Verdent (approved):

PR #5244, opened 2023-01-26
No longer mark ByteArrayInputStream in LoggingInterceptor

alexsuter:
If the entity has been buffered before with (response.bufferEntity), we'll receive a ByteArrayInputStream here. Calling response.readEntity(...) will always return null, because the marking of a ByteArrayInputStream behaves different then for other streams.


alexsuter:
feedback would be appreciated!

jansupol:
I guess this is good, but it would be better to have a test proving the code is correct. I assume setting ByteArrayInputStream using `ClientResponseContext#setEntityStream` and checking the response would do.

alexsuter:
Thanks for the feedback: @jansupol 
I'll try to implement a test in the next days.

alexsuter:
Hey @jansupol 
I've added a test. I think the change would be good, but I'm able to handle this now also in my client code, by reseting the entity stream on my own. e.g.:

```java
response.bufferEntity();
response.readEntity(String.class)
response.getEntityStream().reset();
```

Then the mark does not have any bad effect on the given `ByteArrayInputStream`

alexsuter:
any feedback?

jansupol:
Sorry, we are preparing for 2.40 release and are reviewing the PRs. We missed this earlier :(

The test shows the reason for your requiring the change, which I did not understand at first. The reason seems to be that there is data consumed before the LoggingInterceptor consumes the data. But I am not convinced the LogginInterceptor should do the reset, Perhaps the data are consumed on purpose. So really the logging interceptor should reset only the data read by the interceptor and not the other.

PR #5270, opened 2023-02-28
allow custom Content-Length for HEAD method

PeterRK:
In some cases, enity size can be figured out without reading all data. When I want to response such size info for HEAD request, I find that Content-Length header is always set 0. Here is a fix.


jansupol:
Yep, this sounds right. 

Please register by the Eclipse Foundation to be able to provide fixes there. See [Missing ECA](https://github.com/eclipse-ee4j/jersey/wiki/Missing-ECA,-Missing-Sign-off-by) for more details.

jansupol:
Please update the copyright year to pass the tests. Feel free to add a test for this functionality.

senivam (approved):

PR #5282, opened 2023-03-20
Get media type fix

jansupol:


core-common/src/main/java/org/glassfish/jersey/internal/util/collection/GuardianStringKeyMultivaluedMap.java
```diff
@@ -170,14 +170,17 @@ public Set<String> getGuards() {
      */
     public boolean isObservedAndReset(String key) {
         Boolean observed = guards.get(key);
-        guards.put(key, false);
+        if (observed != null) {
+            guards.put(key, false);
+        }
```
jbescos, line 175: We could do it in one line with putIfAbsent

jbescos (approved):

senivam (approved):

PR #5301, opened 2023-03-31
merge of the actual 3.x into 3.1 branch

senivam:


core-server/src/main/java/org/glassfish/jersey/server/internal/routing/AbstractMethodSelectingRouter.java
```diff
@@ -399,6 +399,7 @@ private List<Router> getMethodRouter(final RequestProcessingContext context) {
 
         final List<ConsumesProducesAcceptor> satisfyingAcceptors = new LinkedList<>();
         final Set<ResourceMethod> differentInvokableMethods = Collections.newSetFromMap(new IdentityHashMap<>());
+        final MediaType requestContentType = request.getMediaType();
         for (ConsumesProducesAcceptor cpi : acceptors) {
             if (cpi.isConsumable(request)) {
```
jansupol, line 404: The `cpi.isConsumable(request)` is not expanded to `cpi.isConsumable(ContainerRequest requestContext, MediaType contentType)` as is in 3.x. (`MethodSelectingRouter`)

jansupol: See [here](https://github.com/eclipse-ee4j/jersey/blob/master/core-server/src/main/java/org/glassfish/jersey/server/internal/routing/MethodSelectingRouter.java#L216)

senivam: unfortunately since [3.0.4](https://github.com/eclipse-ee4j/jersey/tree/3.0.4/core-server/src/main/java/org/glassfish/jersey/server/internal/routing) there is no MethodSelectingRouter neither in 3.x nor in 3.1. However, it exists in the [3.0.3](https://github.com/eclipse-ee4j/jersey/tree/3.0.3/core-server/src/main/java/org/glassfish/jersey/server/internal/routing). Do you recall any reason why it was removed?

jansupol: It has been split into two (OctetStreamMethodSelectingRouter, WildcardMethodSelectingRouter), and the `MethodSelectingRouter` is `AbstractMethodSelectingRouter` now. So it has been wrongly merged to 3.x, too....

jansupol (approved):
So let's fix this in 3.x and merge to 3.1 then

PR #5319, opened 2023-04-28
Additional logging for SNI

jansupol:


jansupol:
I have dropped the localization messages as they are part of jdk.connector now.

senivam (approved):

PR #5326, opened 2023-05-11
Return null instead of throwing exception when querystring parameter is missing

rutterpaul-personal:
See https://github.com/eclipse-ee4j/jersey/issues/5260

Instead of throwing IllegalArgumentException when querystring parameter value is null (e.g. not present), just return null.
Throwing exceptions is expensive, especially when it's done for every querystring parameter that has no @DefaultValue (and is not present in the request querystring).

I'm still looking for a good place to add a unit test. Would ParamConverterInternalTest be it?

### DIFF
--- a/core-common/src/main/java/org/glassfish/jersey/internal/inject/ParamConverters.java
+++ b/core-common/src/main/java/org/glassfish/jersey/internal/inject/ParamConverters.java
@@ -60,7 +60,7 @@ public class ParamConverters {
         @Override
         public T fromString(final String value) {
             if (value == null) {
-                throw new IllegalArgumentException(LocalizationMessages.METHOD_PARAMETER_CANNOT_BE_NULL("value"));
+                return null;
             }
             try {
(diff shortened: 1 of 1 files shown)


rutterpaul-personal:
@jansupol i tried to supply a new PR under my personal account, where i signed the ECA.
But it doesn't seem to accept it, although the github account is properly matching the one of this PR.
What did i miss?

Is it because of the commit message not being in the format defined in https://www.eclipse.org/projects/handbook/#resources-source?

paulrutter:
At least the build succeeded and the ECA is ok.
What else do we need to get this into a release?

Any tips on where we would add a unit test?

jansupol:
Sorry for the delay, still planning this for review for 2.40...

jansupol:
The test...it depends, it could be in `ParamConverterInternalTest`, or for end-to-end tests in `ParamConverterTest`.

senivam (approved):

PR #5331, opened 2023-05-12
allow for resource methods to return <? extends XYZ>

jansupol:


mkarg:
Thank you for fixing this! :+1:

mkarg (approved):
Thanks! :-)

senivam (approved):

mkarg:
You are awesome, guys! Really begging to get this feature ASAP! :-)

mkarg:
What is the schedule for getting this change into a release?

mkarg:
Thank you Maxin, for the outlook, but actually I need to know the schedule for 3.x. I assume that would happen shortly after?

senivam:
usually yes, but it depends on the number of changes to be released.

mkarg:
In Jersey 2.x, when an InterceptorWriter was replacing a Collection's **content** (e. g. from `Collection<Apple>` to `Collection<Banana>`) *solely setting the entity* by `context.setEntity(myCollection)` (hence, *no* `context.setGenericType()` was used), then the type of the objects inside of the collection was correctly detected to be `Banana`, not `Apple` anymore. Since Jersey 3.x, Jersey forwards the *old* type, so `Apple` is still expected, leading to Yasson being unable to cast `Banana` to `Apple`.

Will this PR *also* resolve *that* problem, too, or is the WriterInterceptor required to *also* `context.setGenericType()`?

jbescos (approved):

ar30067:
in class ResourceMethodValidator.java Line 134 
`
 final Type responseType = invocable.getResponseType();
        if (!isConcreteType(responseType)) {
            Errors.warning(invocable.getHandlingMethod(),
                    LocalizationMessages.**TYPE_OF_METHOD_NOT_RESOLVABLE_TO_CONCRETE_TYPE**(responseType,
                            invocable.getHandlingMethod().toGenericString()));
        }
`
is this warning related to this issue ?

jansupol:
@ar30067 It may be in case your resource methods return <? extends XYZ>  or may not if the cause is different.

PR #5341, opened 2023-05-20
Improving I/O Efficience and Performance using NIO API instead of IO API

mkarg:
# Target
The aim of this PR is to improve…
* …🌳 Efficiency: Consume less power, produce less lost heat and CO₂.
* …🏎 Performance: Transfer data in less time.

# Justification
Jersey is a key component of tens of thousands of servers operated world-wide, hence it should perform data transfers as quickly as possible, with least resource consumption as possible.

# Description
This PR replaces usage of IO API by NIO API in some parts of Jersey.

# Non-Targets
* It is *not* a target of this PR to replace IO API *completely*. I only applied NIO API *where needed* for the aim of this PR (plus some locations I came across incidentally while coding this PR).
* It is *not* a target of this PR to *ultimately* reduce resource consumption / maximize performance. This PR shall only be a *starter* for introducing *OS-offloading* capabilities into Jersey 3.2.
* It is *not* a target of this PR to backport changes into Jersey 3.1 or 2.x, as performance optimizations are like new features IMHO, so they should not pop up in maintenance-only branches.
* It is *not* a target of this PR to *completely overhaul* all existing source code. Wherever I optimized or simplified the source code, it was only done in files or code lines that are to be touched by this PR *anyways*.

## Background Information
Since Java 7 (and later, Java 9 and 11) there is a new NIO API which allows the JRE to do internal "tricks":
* Offload transfer of data to optimized algorithms inside of the JRE (e. g. many implementations of `transferTo` recently got hand-optimized, to work faster than typical custom application code and/or skip execution if possible and/or offload to wrapped streams, etc.).
* Offload transfer of data to lower leves of the technology stack (e. g. stream-to/from-file and stream-to/from-socket is offloaded to the operating system, i. e. bypasses the Java heap, hence no time is squandered by useless native-to/from-Java-RAM transfers).
* Use JRE-internal short-cuts, which are not visible at the public API level (e. g. the JRE internally maintaines reusable buffers, while applications have to recreate buffers always).
* Use JRE-internal optimizations basing on envidence instead of hope (e. g. the JRE's byte buffer size was recently optimized using intensive JMH benchmarks on several technology stacks, so it outperforms *in all cases* any kind of custom application code).
* etc.

*Note: I deliberately did not squash the commits, so reviewers get a better understanding of the particular reason for each change.*

*Note: I am the original author of several of the mentioned JRE-internal optimizations, and my target is to help downstream projects to adopt them, hence to make most of them. If there are any questions regarding these internals, feel free to directly ask me. NB: One of the projects I already helped to optimize was Maven, which benefits from the changes I made since several months.*


core-common/src/main/java/org/glassfish/jersey/message/internal/AbstractMessageReaderWriterProvider.java
```diff
@@ -42,18 +43,24 @@
...
      * @param in  the input stream to read from.
      * @param out the output stream to write to.
      * @throws IOException if there is an error reading or writing bytes.
+     *
+     * @deprecated use {@code in.transferTo(out)} instead.
      */
+    @Deprecated
     public static void writeTo(InputStream in, OutputStream out) throws IOException {
-        ReaderWriter.writeTo(in, out);
+        in.transferTo(out);
```
jansupol, line 63: In general, I am not in favor of replacing `ReaderWriter.writeTo`. While transferTo seems to do exactly the same as `writeTo`, Jersey has a mechanism to extend the buffer size used.

core-common/src/main/java/org/glassfish/jersey/message/internal/ReaderWriter.java
```diff
@@ -157,80 +156,6 @@ public static String readFromAsString(Reader reader) throws IOException {
...
-     * Attempts to allocate larger arrays may result in
-     * OutOfMemoryError: Requested array size exceeds VM limit
-     */
-    private static final int MAX_BUFFER_SIZE = Integer.MAX_VALUE - 8;
-
-    /**
-     * Java 9+ InputStream::readAllBytes
-     * TODO Replace when Java 8 not any longer supported (3.1+)
-     */
-    private static byte[] readAllBytes(InputStream inputStream) throws IOException {
```
jansupol, line 172: This method is actually good. While it copies the code from the JDK, one can compare the method in JDK 11 and 17, and see some differences. So this method uses the tweaked code from JDK 17.

The method also contains an important fix for streams that have issues with reading 0 bytes from them.

core-common/src/main/java/org/glassfish/jersey/message/internal/AbstractMessageReaderWriterProvider.java
mkarg: In a JMH test campaign at OpenJDK we have empirically *proven* that it makes *no sense* for any application to use any other buffer sizes than the JRE's default size. In particular we have proven that it makes *no sense* to use buffers smaller than 8K (JRE's previous default) / larger than 16K (JRE's current default). Keeping that configuration option will *in virtually all* cases result in (very) worse efficience and performance compared to using `transferTo`. Hence I do not see what your target is by *not* using `transferTo`. Using `transferTo` is the *sole* key point of this PR, and is the *sole proven* driver for optimum efficience and performance. So if you decision is final, we can drop this PR -- but this means that Jersey will always have bad efficency and performance.

mkarg: BTW, transferTo is definitively *not* doing the same as Jersey's custom code in recent JREs. In fact, modern JREs offload the data transfer to the OS, which leads to dramatic performance and efficiencecy gains. My PR targets on moden JREs, not on JRE 11. That's why I opened it for 3.1, not for 3.x / master. Please check the actual source code of JRE 21-SNAPSHOT.

mkarg: A compromise could be that we keep calling `ReaderWriter.transferTo` but inside of that method we add an option that is calling the old code *if explicitly wanted*. For example, we could replace the default buffer size 8192 by -1. For -1 we call transferTo. For any other number (hence explicitly set 8192 or any other value) we call the original code. WDYT?

core-common/src/main/java/org/glassfish/jersey/message/internal/ReaderWriter.java
mkarg: I do not see the point. This PR is intended for Jakarta EE 11 which will have JDK 17 or even 21 as its minimum, which brings the code you want by default. So why keeping this method?

core-common/src/main/java/org/glassfish/jersey/message/internal/AbstractMessageReaderWriterProvider.java
mkarg: @jansupol I think this is what we could agree upon: c118b65a235d9a2d4ef9ada743312998a2ce04b0. It preserves the option to explicitly set the buffer size, but unless one does so, it let's the JRE do its job. So user who don't care get JRE optimizations out-of-the-box, and users who want to get the old code, can explicity set 8192. This is the best of all worlds.

mkarg: NB: User should abstain from actually setting *particularly* the original value of 8192. If one really has tested different buffer sizes carefully (like we did at OpenJDK), he will notice that **16384** *always* outperfoms 8192, while bigger buffers do not provide *considerable* additional gain (but cost a lot of RAM in sum on production servers with massive parallel requests). Due to that, modern JREs use 16384 (at least for byte buffers at time of writing).

jansupol: @mkarg ReaderWriter's BUFFER_SIZE is a public property with a default value of 8192. This must not be changed, think about compatibility.

I do not see much difference in my early copy of JDK 21. But I agree, the JDK internals can change. This is seen by extending the buffer values in JDK 21. While the tests with JDK may show the large buffer brings a slight performance boost, the customer may judge otherwise, based on their use case. 

You are right, however, that by default the buffer size is not changed by the majority of customers. Hence, what we can do is forward to JDK's transfer method for cases where the customer does not change the BUFFER_SIZE to a value different from 8192.

core-common/src/main/java/org/glassfish/jersey/message/internal/ReaderWriter.java
jansupol: The change was in the code, these lines:
```
                if (nread == BUFFER_SIZE) { // This differs from JDK version
                    break;                  // prevents a bug (See ReaderWriterTest)
                }
```
The JDK's code reads in a loop until the buffer is full and then the loop asks the stream to read again. That makes the read of zero bytes confuse some streams and return -1. While the stream's behavior is arguably incorrect, it would be a sudden change in Jersey that would make many years of running code fail miserably for a large customer. 

Is there a strong reason for dropping the code?

core-common/src/main/java/org/glassfish/jersey/message/internal/AbstractMessageReaderWriterProvider.java
mkarg: @jansupol I see we come nearer to a final solution. :-)

What we measured at OpenJDK is that the performance / efficiency benefit strongly depends greatly on the actual *platform stack*, less of the actual *use case*, which is why we still have decided for a *one size fits it all* buffer size even in JDK 21. I do not see any evidence that this experience should not fit to particularly Jersey. Anyways I have no objection against keeping the IO_BUFFER_SIZE option available for those rare cases that I might not see but possibly exist (do you actually know some?).

The actual code changes BTW can be found here: https://github.com/openjdk/jdk/pulls?q=is%3Apr+is%3Aclosed+author%3Amkarg. They (mostly) are effective *solely* when using transferTo, in particular when using transferTo in conjunction with JRE (non-custom) streams (hence the lots of changed implementations) - as you can see, over the past years, we changed and discussed a lot of JRE code, and that for a good reason (to make the most of modern server hardware / to not stand in the way of OS-internal optimizations). 

Regarding the default size: I do not think that anybody really relies on particularly the value 8192 being the default for all times, but people instead rely on "BUFFER_SIZE == IO_DEFAULT_BUFFER_SIZE", and they rely on IO_DEFAULT_BUFFER_SIZE telling them the *actual* value at any time they can trust upon *to be the default*, so I think it would be OK if we tell them that IO_DEFAULT_BUFFER_SIZE == -1 from the next minor or major release. IMHO existing applications that did *not* check IO_DEFAULT_BUFFER_SIZE are simply *wrong* because that is the reason why the docs do *not* say "8192" but say "IO_DEFAULT_BUFFER_SIZE", right?

Only enabling the fallback code in case the application is setting a value *other than 8192* sounds rather odd and ambiguous. What if a user *wants* to really use just 8192 bytes, because he does not want to squander the new JRE default of 16384 (or possibly even more bytes in JRE 22 or 23)? So I really would recommend to go with IO_DEFAULT_BUFFER_SIZE == -1 instead, to make the actual behavior change *visible* (if we trust our users to read the release notes), or (for maximum backwards compatibility) to *actually detect **any** change* by using a Jersey-internal flag which is set to true as soon as IO_BUFFER_SIZE >= 0 (i. e. Jersey *internally* ignores BUFFER_SIZE, but externally the BUFFER_SIZE backward-compatibly reflects IO_DEFAULT_BUFFER_SIZE == 8192).

**Having said that, in what direction shall I effectively migrate this PR now?**

core-common/src/main/java/org/glassfish/jersey/message/internal/ReaderWriter.java
mkarg: Yes, there is a strong reason: Future JRE-internal optimizations will not be effective in Jersey (e. g. hypothetically we might decide to replace Java completely by C code).

Certainly the investments done in OpenJDK SHOULD (MUST?) reflect its benefits to Jersey users, too. So that situation would be a major drawback for *all* Jersey users! In an open source project it simply MUST NOT happen that one single party (read: customer) prevents *all other* parties from progressing. I think we all agree on this, particularly now that we are hosted at the EF.

BTW, the same problem could arise (for the same customer and any other users) at any time in *any other* application, as it is not limited to Jersey, but limited to the particular stream implementation. So if such a fix is *essentially needed*, it has to go into the JRE, to be on the safe side, instead of limiting the fix particularly to Jersey. But if the fix is *not needed* (hence was only a nice gift to the customer), the customer would be really on the safer side when he fixes his custom stream instead (which I assume, but don't know, to be the case here).

Having said that, a compromise could be:
* Let's keep the code *for 3.x* so the customer / we gain more time for a real fix *at the right place*.
* Let me check with my OpenJDK peers if we add the fix to JDK 21 / 22. If they refuse to accept the change proposal, the customer should get contacted that the nice gift will go away eventually.
* Let's drop the code in Jersey 4.

I think that would be fair to *all* parties.

core-common/src/main/java/org/glassfish/jersey/message/internal/AbstractMessageReaderWriterProvider.java
mkarg: @jansupol "ping" - I need your decision regarding above proposal. Thanks.

core-common/src/main/java/org/glassfish/jersey/message/internal/ReaderWriter.java
jansupol: Yep, keeping it for 3.x sounds like a good compromise. It could have been revisited once the JDK really brings in an internal optimization. Right now, keeping the method as optimized as possible from the newest JDK would bring positives to people using old Java (11).

core-common/src/main/java/org/glassfish/jersey/message/internal/AbstractMessageReaderWriterProvider.java
jansupol: Really it is not a good practice to remove or change public properties, and values. 

You have a point that people might require a lower buffer than the one in JDK on purpose. However, since the change comes in JDK 21 which is not released yet, it is not very likely people will want to override that value, and it makes more sense to me to live with overriding this value than changing public constants. 

If you want to be precise, you can check for JDK version (using the Jersey JdkVersion class) and if JDK is 21, keep Jersey routine for BUFFER_SIZE 16384. But then, is it not the goal to use JDK code instead of Jersey? While I can imagine the poor guy spending days on testing various BUFFER_SIZE with JDK 11, I can hardly imagine that guy doing so with beta JDK 21.

jansupol:
Please update the copyright year in modified files, the CI tests fail otherwise.

core-common/src/main/java/org/glassfish/jersey/message/internal/ReaderWriter.java
mkarg: Per our compromise I reintroduced the original code in 6b5d1948d0ea8f147b1bf01c65f6b20ef0386388, so the workaround still exists until Jersey 4. I have updated the TODO so nobody gets confused by the original justification. I also filed [an issue](https://github.com/eclipse-ee4j/jersey/issues/5344) so we do not forget to fix this in Jersey 4.

JDK 21 is already in Rampdown Phase, so *even if* Team OpenJDK adops the workaround (which is not decided yet) it will not make it into any JDK earlier than 22. It might make sense to contact the customer right now, as chances to get the workaround into JDK are actually rather low.

mkarg:
> > java.lang.NegativeArraySizeException: -1
> > at org.glassfish.jersey.message.internal.ReaderWriter.readAllBytes(ReaderWriter.java:193)

I know - this is because I did not finish work, because the discussion whether we go with -1 or not is still open.

jansupol:
I believe we need to keep all the properties (including IO_DEFAULT_BUFFER_SIZE) and not make any troubling changes.

Since the changes are JDK-related, I would suggest opening the PR against the master branch so that each version of Jersey can benefit from your updates.

mkarg:
> I believe we need to keep all the properties (including IO_DEFAULT_BUFFER_SIZE) and not make any troubling changes.

I will do as you say.

> Since the changes are JDK-related, I would suggest opening the PR against the master branch so that each version of Jersey can benefit from your updates.

The master branch uses Java 8 while my changes rely on Java 11, so I do not see a way to make that work.

mkarg:
@jansupol I have reintroduced IO_DEFAULT_BUFFER_SIZE in 2750a6faa898fa5e196b22c53f802238fea03814 as you asked for. In that revision the user can even explicitly demand a static 8k buffer if he really wants to. Unless the buffer size is explicitly set, the JRE will decide on its own. WDYT?

tests/e2e-core-common/src/test/java/org/glassfish/jersey/tests/e2e/common/message/internal/ReaderWriterTest.java
```diff
@@ -29,7 +29,7 @@ public class ReaderWriterTest {
     @Test
     public void testToNeverAskToReadZeroBytes() throws IOException {
         // Unnamed app server bug test
-        int size = ((ReaderWriter.BUFFER_SIZE + 1000) / 10) * 10;
+        int size = (((ReaderWriter.BUFFER_SIZE) + 1000) / 10) * 10;
```
jansupol, line 32: Copyright Year

pom.xml
```diff
@@ -961,8 +933,8 @@
                                         <excludes>
                                             <exclude>module-info.java</exclude>
                                         </excludes>
-                                        <source>1.8</source>
-                                        <target>1.8</target>
+                                        <source>${java.version}</source>
+                                        <target>${java.version}</target>
```
jansupol, line 937: This probably is wrong. It does not need to be 1.8, it can be 11, but it should not be higher. When building Jersey, we use about three versions of JDK to produce proper multi-release jars.

jansupol:
> @jansupol I have reintroduced IO_DEFAULT_BUFFER_SIZE in [2750a6f](https://github.com/eclipse-ee4j/jersey/commit/2750a6faa898fa5e196b22c53f802238fea03814) as you asked for. In that revision the user can even explicitly demand a static 8k buffer if he really wants to. Unless the buffer size is explicitly set, the JRE will decide on its own. WDYT?

This part looks good now.

jansupol:
Have you thought about putting the changes to 2.x? Most of it would work there (but the NullOutputStream)

pom.xml
mkarg: Actually this line is correct. `${java.version}` is used all over this file, and is *not* the JDK version, but actually is a *constant* defined in line 2157 of this file which explicitly locks to the particular value `11`, but not higher - just as you want it to be. :-)

mkarg:
> Have you thought about putting the changes to 2.x? Most of it would work there (but the NullOutputStream)

As I wrote earlier, this PR will not build *unchanged* on Jersey 2.x, as 2.x is based on Java 8. Besides NullOutputStream, InputStream::transferTo did not exist prior to Java 9, and Reader::transferTo did not exist prior to Java 10. So what I can do (with acceptable effort) is: Once *this* PR is merged into the 3.1 branch, I can open a second PR for the 2.x branch with backports of *just those (few)* parts that actually will work in Java 8.

mkarg:
> This part looks good now.

Thank you, Jan. I have switch this PR from draft to ready status, so Github allows to merge it. I propose you apply Github's own squash button (so I abstain from squash-and-force-push), so we do not lose the conversion history in Github.

mkarg:
> Have you thought about putting the changes to 2.x? Most of it would work there (but the NullOutputStream)

@jansupol I have opened the separate PR https://github.com/eclipse-ee4j/jersey/pull/5350 which contains all JRE 8 compatible changes found in *this* PR.

mkarg:
This PR exists since May 20. Today it is November 12. It would be nice to get told why not integrating it for so long. Thank you.

jansupol:
I pretty much was under the impression you are aware. Mostly because the tests never passed, for the reasons we discussed in the PR for 2.x. The stuff was merged in 2.x and now it propagated to 3.1 branch. Some parts of this PR could be added to the propagated changes, but not in the form as it is right now, as again, it breaks the build.

jansupol:
I am sorry, I was wrong. The changes of yours are yet to be propagated, it is a [WIP](https://github.com/eclipse-ee4j/jersey/pull/5461)

mkarg:
@jansupol Can you please point me to the discussion regarding the failed builds?

jansupol:
I think it is [this one](https://github.com/eclipse-ee4j/jersey/pull/5350#discussion_r1260763422). 

The PR also suffered from JDK issues on Windows, and I needed to create a [fix for the NIO](https://github.com/eclipse-ee4j/jersey/pull/5392) changes. [I filed a bug](https://bugs.openjdk.org/browse/JDK-8314511) on JDK but the behavior won't be changed.

I would welcome it if you could continue with this PR, build atop the merged 3.0 (once done), and add just what's relevant for this JDK11-related 3.1 branch.

mkarg:
> I think it is [this one](https://github.com/eclipse-ee4j/jersey/pull/5350#discussion_r1260763422).

I don't think so, as that one was fixed, see https://github.com/eclipse-ee4j/jersey/pull/5350#discussion_r1261500842.

> The PR also suffered from JDK issues on Windows, and I needed to create a [fix for the NIO](https://github.com/eclipse-ee4j/jersey/pull/5392) changes. [I filed a bug](https://bugs.openjdk.org/browse/JDK-8314511) on JDK but the behavior won't be changed.

Thank you for the pointer! From my understanding I need to agree with Alan that there is actually no bug in the JDK but the problem comes from the way Jersey uses paths. I will look into your fix to check if I can optimize it without using `File`, or if we have to finally stick with it as-is.

> I would welcome it if you could continue with this PR, build atop the merged 3.0 (once done), and add just what's relevant for this JDK11-related 3.1 branch.

I will overhaul this issue ASAP and drop a note when I'm done. :-)

jansupol:
Ok, so your PR to 2.x is now propagated to 3.1 as well.

mkarg:
> Ok, so your PR to 2.x is now propagated to 3.1 as well.

@jansupol Sorry for the delay, I was booked out with other projects. Nevertheless, I am done with this PR. I have squashed it into one single commit and rebased ontop of current 3.1. It passes the build tests (but Jenkins fails with OutOfMemory), so I think it now is in a state where it can be merged. :-)

mkarg:
@jansupol This PR is IMHO ready for merge. Regarding the problems on Windows, we are working on an improvement, see https://bugs.openjdk.org/browse/JDK-8321591.

mkarg:
@jansupol Is there anything I have to do before this PR is getting merged?

mkarg:
@jansupol "ping"? ;-)

jansupol:
@mkarg The tests are failing with this PR. Have you checked the reason?

mkarg:
> @mkarg The tests are failing with this PR. Have you checked the reason?

Thank you Jan. I will look into the failing tests.

mkarg:
> @mkarg The tests are failing with this PR. Have you checked the reason?

I could not find any failing tests but only two outdated Copyright notes. After fixing them CI was happy: https://ci.eclipse.org/jersey/job/ci-cd-integration/job/PR-5341/37/

mkarg:
@jansupol IMHO this PR is ready to merge now.

mkarg:
@jansupol Removal of JDK 8 in 3.1 induced merge conflicts. I fixed them, so this branch passes all tests now, hence is ready-to-merge again. :-)

pizzi80:
When this PR will be merged?

mkarg:
> When this PR will be merged?

This is a really good question. Indeed I do wonder what @jansupol is waiting for, actually! 🤔

wendigo:
Can we get this into 3.1.7 @jansupol ?

jansupol:
Thanks @mkarg

mkarg:
Thank you, @jansupol, for accepting my changes!

jansupol:
@mkarg Sorry it took so long

mkarg:
> @mkarg Sorry it took so long

No worries! :-) If there is anything volunteers could help in Jersey, just let me know.

PR #5350, opened 2023-06-18
Using Java7+ NIO API for improved performance

mkarg:
This PR is a partial backport of https://github.com/eclipse-ee4j/jersey/pull/5341 as proposed by @jansupol in https://github.com/eclipse-ee4j/jersey/pull/5341#issuecomment-1589610557.

Compared to PR 5341 it only contains those changes which are compatible with Java 8. In particular this means, it mostly consists of using Java 7+ NIO API plus some more improvements, while the actual benefit (preventing data transmission via the JVM's RAM implemented by `ChannelInputStream::transferTo` in recent JREs) is missing due to the fact that `InputStream::transferTo` was not added to the JRE before Java 9.

### DIFF
--- a/core-common/src/main/java/org/glassfish/jersey/message/internal/ReaderWriter.java
+++ b/core-common/src/main/java/org/glassfish/jersey/message/internal/ReaderWriter.java
@@ -245,9 +245,9 @@ public final class ReaderWriter {
      * @throws IOException in case of a write failure.
      */
     public static void writeToAsString(String s, OutputStream out, MediaType type) throws IOException {
-        try (final Writer osw = new OutputStreamWriter(out, getCharset(type))) {
-            osw.write(s);
-        }
+        Writer osw = new OutputStreamWriter(out, getCharset(type));
+        osw.write(s);
+        osw.flush();
     }
 
     /**
--- a/core-common/src/main/java/org/glassfish/jersey/message/internal/ReaderWriter.java
+++ b/core-common/src/main/java/org/glassfish/jersey/message/internal/ReaderWriter.java
@@ -19,12 +19,12 @@ package org.glassfish.jersey.message.internal;
 import java.io.Closeable;
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.InputStreamReader;
 import java.io.OutputStream;
 import java.io.OutputStreamWriter;
 import java.io.Reader;
 import java.io.Writer;
 import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
 import java.security.AccessController;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -54,8 +54,11 @@ public final class ReaderWriter {
     private static final Logger LOGGER = Logger.getLogger(ReaderWriter.class.getName());
     /**
      * The UTF-8 Charset.
+     *
+     * @deprecated use {@code StandardCharsets.UTF_8} instead
      */
-    public static final Charset UTF8 = Charset.forName("UTF-8");
+    @Deprecated(forRemoval = true)
+    public static final Charset UTF8 = StandardCharsets.UTF_8;
     /**
      * The buffer size for arrays of byte and character.
      */
@@ -167,7 +170,7 @@ public final class ReaderWriter {
 
     /**
      * Java 9+ InputStream::readAllBytes
-     * TODO Replace when Java 8 not any longer supported (3.1+)
+     * TODO Replace in Jersey 4.0, as the sole difference to OpenJDK is working around a bug in the input stream.
      */
     private static byte[] readAllBytes(InputStream inputStream) throws IOException {
         List<byte[]> bufs = null;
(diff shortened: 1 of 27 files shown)


mkarg:
@jansupol You asked me to provide these changes explicitly for the master branch, so why not merging it? The PR is open since three weeks, and it solely contains backports of the already reviewed original PR from the 3.x branch.

core-common/src/main/java/org/glassfish/jersey/message/internal/ReaderWriter.java
```diff
@@ -242,9 +245,9 @@ private static byte[] readAllBytes(InputStream inputStream) throws IOException {
      * @throws IOException in case of a write failure.
      */
     public static void writeToAsString(String s, OutputStream out, MediaType type) throws IOException {
-        Writer osw = new OutputStreamWriter(out, getCharset(type));
-        osw.write(s, 0, s.length());
-        osw.flush();
+        try (final Writer osw = new OutputStreamWriter(out, getCharset(type))) {
```
jansupol: I do feel this to be a source of issues. While the try with resources closes the OutputStream, the original flush did not close it.

mkarg: Good catch! I will modify this code so it does not close the stream but only flush it (in fact I wonder why you flush it at all, as this should be the *caller's* job, but anyways, it then does what it did before).

mkarg: @jansupol I fixed it in https://github.com/eclipse-ee4j/jersey/pull/5350/commits/85a1e3d9f85e002798032d1e34103e4da19d648f.

tests/performance/tools/src/main/java/org/glassfish/jersey/tests/performance/tools/TestDataGeneratorApp.java
```diff
@@ -134,7 +133,7 @@ public static void generateFile(final String resourceRelativeUrl, final int mini
 
         final File file = new File(fileName);
         final BufferedWriter writer = new BufferedWriter(new OutputStreamWriter(
-                new FileOutputStream(file), Charset.forName("UTF-8")));
+                Files.newOutputStream(file.toPath()), Charset.forName("UTF-8")));
```
jansupol, line 137: You need to add import for Files:
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.9.0:compile (default-compile) on project performance-test-tools: Compilation failure
/home/jenkins/agent/workspace/ci-cd-integration_PR-5350/tests/performance/tools/src/main/java/org/glassfish/jersey/tests/performance/tools/TestDataGeneratorApp.java:[136,17] cannot find symbol
  symbol:   variable Files
  location: class org.glassfish.jersey.tests.performance.tools.TestDataGeneratorApp

mkarg: Fixed in https://github.com/eclipse-ee4j/jersey/pull/5350/commits/0803e1454860b0cde0b6075aac61108b279b7b8d.

mkarg: (The reason why I could not detect that earlier is that I cannot build from scratch: `mvn compile test-compile` on a fresh clone fails with lots of compilation failures outside of my actual changes).

mkarg:
@jansupol I think the communication between Github and Jenkins is broken. I fixed another missing import last night and now Github says it is waiting for the build, but Jenkins says no builds are queued. Is there a way to manually trigger the build?

senivam:
@mkarg, the build is triggered manually, but it does not pass compilation on JDK 1.8 (that thread has already failed).

mkarg:
@senivam Thank you for triggering the build. Let's wait for it to finish. If there are still bugs in the PR I will fix them ASAP once reported.

Edit: Ah, I see. The `@forRemoval` didn't exist on Java 8. Backporting the original PR is more tricky than expected. ;-)

mkarg:
@senivam Maxim, it seems there are no more Java problems in this PR other than the ones produced by Github/Jenkins themselves. In particular, I assume this is unrelated to my work:
```
Creating placeholder flownodes because failed loading originals.
ERROR: Cannot resume build because FlowNode 41 for FlowHead 1 could not be loaded. This is expected to happen when using the PERFORMANCE_OPTIMIZED durability setting and Jenkins is not shut down cleanly. Consider investigating to understand if Jenkins was not shut down cleanly or switching to the MAX_SURVIVABILITY durability setting which should prevent this issue in most cases.

GitHub has been notified of this commit’s build result

Finished: FAILURE
```

senivam:
I've run the build again, the previous attempt even did not start properly (it's kind of known Eclipse's CI issue).

mkarg:
> I've run the build again, the previous attempt even did not start properly (it's kind of known Eclipse's CI issue).

Thank you. Meanwhile your manually triggered build terminated successfully, so this PR is (at last!) ready to merge. 🥳

mkarg:
@jansupol Is there something else you want me to fix in this PR? I have not seen open comments, and the last status update is ~~two~~ three weeks ago. Did I miss to fix something?

mkarg:
@senivam Maxim I wonder why there is no progress with this PR. It fulfils all changes requested by Jan since weeks. Jan did not answer, so maybe you could share some thoughts what I need to do to finally get this PR merged? Thanks a lot! :-)

jansupol (approved):
@mkarg No need to do anything, this looks good. Thanks.

senivam (approved):
LGTM

Issue #4815, opened 2021-06-29
Delayed synchronization SSL context initialization in HttpUrlConnector affects functionality


2368001770:
Is there a bug with jersey2.34？I found that “DEFAULT_SSL_SOCKET_FACTORY.get()” and "HttpsURLConnection.getDefaultSSLSocketFactory()" are not  always equal.
![image](https://user-images.githubusercontent.com/44369334/123797811-28643e00-d919-11eb-8e83-bd581496499a.png)

jansupol:
There is #4757 that requests the context to be initialized lazily.
There is #4566 that requests the context to be initialized statically.
There is an issue with the native image that requests the context to be initialized lazily.

There seem to be a lot of contra requirements.

senivam:
error message for static (not lazy) default SSL socket factory initialization when used with GraalVM native image:
```
com.oracle.svm.core.util.UserError$UserException: No instances of sun.security.provider.NativePRNG are allowed in the image heap as this class should be initialized at image runtime. To see how this object got instantiated use --trace-object-instantiation=sun.security.provider.NativePRNG.
```

jansupol:
#4332

dtbaum:
This issue is fixed in  [2.41](https://github.com/eclipse-ee4j/jersey/releases/tag/2.41) (see PR: #5359)

Issue #4757, opened 2021-03-29
SSLContext gets initialized too early


bomgar:
Hi,

I have an application that modifies the java Security stack during runtime. This includes the SSL setup. Before this happens I use Jersey to make some http (not https) calls. This causes the SSLContext to initialize

The issue is the static final field here: `org.glassfish.jersey.client.internal.HttpUrlConnector`
```java
    private static final SSLSocketFactory DEFAULT_SSL_SOCKET_FACTORY = HttpsURLConnection.getDefaultSSLSocketFactory();

```

This initializes the SSLContext even if the client is only used for http. 

So this is called:
```java
    public static synchronized SSLContext getDefault()
            throws NoSuchAlgorithmException {
        if (defaultContext == null) {
            defaultContext = SSLContext.getInstance("Default");
        }
        return defaultContext;
    }

```
The default context does not get created again after that even if i modify the Security settings. 

This should only happen when SSL is actually used for the first time.

jansupol:
This seems to go right against #4566.

jansupol:
We may use `LazyValue` that uses the double-checked-locking. The LazyValue ought to be better for the native image, too.

PR #5359, opened 2023-07-07
Race condition

### LINKED ISSUE #5358: Race condition
When the first GET request is in progress, all parallel requests from other Jersey clients, in other threads, fail with **SSLHandshakeException: PKIX path building failed**. 
Once the first GET request is completed,  all subsequent requests work without error.
Here is the code which demonstrates this issue:
https://github.com/dtbaum/jersey-bug-report

The fix in HttpUrlConnector.java solved the issue, see pull request https://github.com/eclipse-ee4j/jersey/pull/5359.
Affected versions: at least 3.1.2 & OpenJDK17 and 2.37 & OpenJDK 8

dtbaum:
Fixed race condition, see:
 https://github.com/eclipse-ee4j/jersey/issues/5358


jansupol:
@dtbaum Can you please elaborate on how the race condition happens?

dtbaum:
@jansupol The race condition occurs when multiple independent Jersey client instances, running in parallel threads, make a GET request. 

When the first GET request is in progress, all parallel requests from other Jersey clients (other client instances, but same SSLContext and same target), in other threads, fail with SSLHandshakeException: PKIX path building failed.
The first caller uses the correct trust store (which is defined in used SSLContext), the parallel callers use the java default trusstore.. The problem is not working `if (DEFAULT_SSL_SOCKET_FACTORY.get() == suc.getSSLSocketFactory())` condition. For subsequent requests it returns suc.getSSLSocketFactory() default factory, BUT another object reference to default factory, therefore the reference equality doesn't work here. A deep equality check, like working .equals() or removing the condition would solve it. 
Once the first GET request is completed, all subsequent requests work without error.
Here is the code which demonstrates this issue:
https://github.com/dtbaum/jersey-bug-report

If you get any issues getting the example to run, please let me know. I would add it as a broken JUnit test on this jersey branch.

dtbaum:
@jansupol: Did you succeed in reproducing the problem?
The reason for the exception is following: Due to mentioned reference equality check,  the "parallel" threads do not use the specified SSLContext, but the default SSLContext  instead. 
Consequently, the server certificate validation occurs **not against the specified truststore**, but against the default truststore, leading to the mentioned exception.

jansupol:
Thank you for the reproducer @dtbaum. The fix breaks `SslHttpUrlConnectorTest.testSSLWithCustomSocketFactory`, but we'll check if there is any possible fix for it.

jansupol:
Related #4815 

This bug has a long history. Basically, there is a bug in JDK that we try to workaround, but there seem to be multiple requirements that go against each other.

dtbaum:
@jansupol:  As much as I understand (which isn't much :-) ) a deep, reliable equality check, like working .equals() instead of object reference equality, would help, without violating other requirements?

dtbaum:
@jansupol:  I think, I have a fix for the race condition, without violating #4757/#4566.  
I have made the method HttpUrlConnectorProvider.ConnectionFactory.getConnection() thread-safe: 
This ensures that parallel initialization in the HttpsURLConnection.getDefaultSSLSocketFactory()  is avoided.
See the screenshot below for call hierarchy:
image

jansupol (approved):

senivam (approved):

fwippe:
@senivam @jansupol @dtbaum Please note that this change has been having a drastic impact on performance, in particular when using Jersey client for multiple services: Only one connection can be created at a time, across all routes. If the intention had been to just synchronize `HttpsURLConnection.getDefaultSSLSocketFactory()`, as an alternative you could have introduced a lazy one-off supplier initializing the socket factory. See #5738 for performance impact in practice.

PR #5372, opened 2023-07-19
Jetty 12 multirelease

zUniQueX:
Supersedes #5342 

This PR provides an early preview of the Jetty 12 upgrade with Jetty 12.0.0.beta4.

Jersey 3.1.x currently has compatibility issues with Jetty. Jersey 3.1 is compatible with Jakarta EE 10 and therefore uses `servlet-api` 6. But the current Jetty version 11.0.x isn't compatible with `servlet-api` 6.

Although the upgrade to Jetty 12 is a major upgrade, this can be released for the servlet compatibility in Jersey 3.1.x IMHO.

Notable changes:
 - The `servlet-api` isn't integrated into Jetty's core classes any more and rather moved out to a separate module. For servlet classes the correct servlet module has to be included (e.g. `jetty-ee10-servlet` for Jersey 3.1).
   The `JettyHttpContainer` is rewritten by removing all servlet related classes. This prevents incompatibilities with future servlet versions.
 - The `Handler` interface has changed and now uses a `Callback` class to signal output completion. The `handled` state is set with the method's return value.
 - `JettyHttpContainer`: The `SecurityContext` is set via the `AuthenticationState`. In the current implementation no `SecurityHandler` is provided and therefore the `AuthenticationState` will always be `null`. Maybe the `JettyHttpContainer` should implement the `Handler.Singleton` interface to let users specify a `SecurityHandler`.


zUniQueX:
The final stage of Jetty 12.0.0 was released yesterday. So I've updated the PR to the released version. Since beta4 no significant changes were introduced.

zUniQueX:
@senivam Thanks for re-triggering the Jenkins build. Unfortunately, looks like the OSGI test is unstable.

senivam:
@zUniQueX I think that's not directly OSGi, but connectivity between ci.eclipse.org/jersey JIPP and eclipse maven repo. It always fails for connectivity reasons.

zUniQueX:
@senivam Thanks for clarification :+1:

senivam:
@zUniQueX you are welcome :) actually, I've retriggered the build again. But it waits in the queue because several builds are already running and this one will run after them. Let's hope it will finish OK.

jansupol:
@zUniQueX Thank you for this effort. Glad to see Jetty 12 is final, yet.

Pros: We definitely want this. ASAP.
Cons: No JDK 11 support. Different Jetty package names.

Open question: Can we do something for people who want latest Jersey 3.1, but JDK 11? Perhaps clone the existing Jetty module into some Jetty-Jdk11 or Jetty11 module?

zUniQueX:
@jansupol I don't know, if that's a good option. Then we would need a downgrade of the servlet api to 5.x because Jetty 11 is EE 9 only. This may cause compatibility issues with other components.

We could check, if Jersey's tests pass in such a module to support _common cases_, but in some edge cases this will probably fail.

jansupol:
@zUniQueX My concern was mainly regarding the client - connector. Jetty Client is more widely used than Jetty Server.

Similarly, we do have Apache connector and Apache5 connector. 
We will need to create Netty 5 connector, soon.
Perhaps we should have Jetty12 connector. The difference, though, is that Jetty version is bound with Jakarta EE versions. Jetty 12 is EE 10 Jetty, but Jetty is behind EE10 release, and people are using Jetty 11 client with Jersey. I am not sure whether we should have Jetty11 or Jetty12 module, but I am positive there should be a module supporting Jetty 11 client. Will see what other people think about it.

zUniQueX:
@jansupol That makes sense to me.

I'd suggest to use the BOM-managed versions in the modules without suffixes, so I'd go for a `jersey-jetty11-connector` module with the current code.

zUniQueX:
@jansupol I've added a module for the Jetty 11.x client. I think that's the better idea than shipping with Jetty 11 in the default Jetty connector module.

To add this connector properly I had to switch to the Grizzly2 container to prevent artifact clashes. That's also the reason why I've skipped adding the new connector to the e2e tests.

One test fails with the Grizzly2 container but I can't understand why. A higher timeout fixes the issue. However, the current timeout should be high enough. I've temporarily disabled the test since we know the connector works because I've just copied the existing code.

jansupol (approved):
LGTM, thanks!

senivam (approved):

zUniQueX:
@jansupol Do you already have plans for the next release? Our EE10 upgrade at dropwizard is currently blocked by a release of this PR.

jansupol:
We are finishing, but it won't happen this week.

zUniQueX:
Okay, thanks :+1:

PR #5436, opened 2023-10-12
Support multipart by Jetty & Netty

jansupol:


senivam (approved):

connectors/netty-connector/src/main/java/org/glassfish/jersey/netty/connector/NettyConnector.java
```diff
@@ -468,9 +469,23 @@ public void run() {
                     }
                 });
 
-                if (entityWriter.getType() != NettyEntityWriter.Type.DELAYED) {
-                    entityWriter.flush();
+                headersSet.await();
```
jbescos, line 472: As we have to wait anyway, do we really need the executorService?. Maybe we can get rid of it and also the CountDownLatch.

jbescos: Ignore it, the countdownlatch is different.

jbescos (approved):

PR #5461, opened 2023-11-09
Merge remote-tracking branch 'upstream/3.0' into 'upstream/3.1' 

senivam:


docs/src/main/docbook/modules.xml
```diff
@@ -116,6 +116,14 @@
 </row>
 <row>
 <entry>
+<link xlink:href="https://eclipse-ee4j.github.io/jersey.github.io/project-info/&version;/jersey/project/jersey-container-jetty-http2/dependencies.html">
```
jansupol, line 119: Double check this: Jetty HTTP/2 for Jetty 12

docs/src/main/docbook/migration.xml
```diff
@@ -28,6 +28,18 @@
          xml:id="migration">
     <title>Migration Guide</title>
 
+    <section xml:id="mig-3.0.11">
```
jansupol, line 31: We should update this for 3.1

ext/bean-validation/pom.xml
```diff
@@ -149,4 +149,16 @@
         </dependency>
     </dependencies>
 
+    <profiles>
```
jansupol, line 152: We do not have JDK 8 in 3.1

inject/cdi2-se/pom.xml
```diff
@@ -132,4 +132,15 @@
             </plugin>
         </plugins>
     </build>
+    <profiles>
```
jansupol, line 135: We do not support JDK 8 in 3.1

tests/e2e/pom.xml
```diff
@@ -217,6 +217,15 @@
                 </plugins>
             </build>
         </profile>
+        <profile>
```
jansupol, line 220: We do not support JDK 8 in 3.1

jansupol (approved):

PR #5466, opened 2023-11-16
Drop 11 for Jetty11 classnames

jansupol:
Support multipart by Jetty 12 connector

The reasoning behind this PR is that Jetty package remains the same. The customer then can just change the module and Jetty dependencies, but not the Jersey app when transitioning from Jetty 11 to Jetty 12


senivam (approved):
LGTM, passes JPMS validation

PR #5467, opened 2023-11-16
Contextualized constructors POC

### LINKED ISSUE #5465: Wrong constructor invocation
I am testing Helidon (that contains Weld as CDI impl) with restull-tcks and there is one test failing `JAXRSClientIT.visibleTest`. I tried to reproduce it in Jersey as a test and the issue does not happen, but this is because the execution flow is different.

The issue is that the resource has various constructors and it is invoking the wrong one. To make it simple imagine the next constructors:
```
        public Resource() {
        }

        public Resource(@Context HttpHeaders headers) {
          this.headers = headers;
        }

```

The TCK test expects that the second will be invoked but it is invoked the first one. I was checking the stack traces and it is related to `jersey/ext/cdi/jersey-cdi1x/src/main/java/org/glassfish/jersey/ext/cdi1x/internal/AbstractCdiBeanSupplier`:
```
    /*
     * Let CDI produce the InjectionTarget. If the constructor contains @Context Args CDI won't be able to produce it.
     * Let the HK2 try to produce the target then.
     */
    private static <T> T produce(InjectionTarget<T> target, CreationalContext<T> ctx, InjectionManager im, Class<T> clazz) {
        try {
            return target.produce(ctx);
        } catch (Exception e) {
            LOGGER.fine(LocalizationMessages.CDI_FAILED_LOADING(clazz, e.getMessage()));
            try {
                return im.create(clazz);
            } catch (RuntimeException re) {
                StringWriter sw = new StringWriter();
                PrintWriter pw = new PrintWriter(sw);
                e.printStackTrace(pw);
                LOGGER.warning(LocalizationMessages.CDI_FAILED_LOADING(clazz, sw.toString()));
                throw re;
            }
        }
    }
```
As there is an empty constructor, the line `target.produce(ctx);` is able to create the resource instance. But in this scenario, the instance should be created by `im.create(clazz);`, because it knows about the Context.

Helidon reproducer here:
https://github.com/jbescos/helidon/blob/tckDisabled/microprofile/server/src/test/java/io/helidon/microprofile/server/JaxRsApplicationVisibilityTest.java

senivam:
The attempt to distinguish constructors with and w/o Context annotation while injecting. 

The code can be re-written with lambdas for better readability.

### DIFF
--- a/ext/cdi/jersey-cdi1x/src/main/java/org/glassfish/jersey/ext/cdi1x/internal/AbstractCdiBeanSupplier.java
+++ b/ext/cdi/jersey-cdi1x/src/main/java/org/glassfish/jersey/ext/cdi1x/internal/AbstractCdiBeanSupplier.java
@@ -19,6 +19,7 @@ package org.glassfish.jersey.ext.cdi1x.internal;
 import java.io.PrintWriter;
 import java.io.StringWriter;
 import java.lang.annotation.Annotation;
+import java.lang.reflect.Constructor;
 import java.util.Iterator;
 import java.util.NoSuchElementException;
 import java.util.Set;
@@ -118,7 +119,8 @@ public abstract class AbstractCdiBeanSupplier<T> implements DisposableSupplier<T
      */
     private static <T> T produce(InjectionTarget<T> target, CreationalContext<T> ctx, InjectionManager im, Class<T> clazz) {
         try {
-            return target.produce(ctx);
+            return hasContextAnnotatedConstructors(clazz) ? im.create(clazz)
+                    : target.produce(ctx);
         } catch (Exception e) {
             LOGGER.fine(LocalizationMessages.CDI_FAILED_LOADING(clazz, e.getMessage()));
             try {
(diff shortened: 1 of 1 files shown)


senivam:
As per CDI spec is not required, closing

ext/cdi/jersey-cdi1x/src/main/java/org/glassfish/jersey/ext/cdi1x/internal/AbstractCdiBeanSupplier.java
```diff
@@ -118,7 +119,8 @@ public void preDestroy(final T instance) {
      */
     private static <T> T produce(InjectionTarget<T> target, CreationalContext<T> ctx, InjectionManager im, Class<T> clazz) {
         try {
-            return target.produce(ctx);
+            return hasContextAnnotatedConstructors(clazz) ? im.create(clazz)
```
jbescos, line 122: We should also change what is inside the Exception block, that currently executes always `im.create(clazz);`, even if it was executed before.

But anyway, it is closed. We will discuss it next Monday probably.

PR #5473, opened 2023-11-21
SameSite is capital first letter only.

jansupol:


jansupol:
See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#samesitesamesite-value

senivam (approved):

PR #5481, opened 2023-11-29
[3.1] Jetty 12 HTTP/2 support

senivam:


connectors/jetty-http2-connector/src/main/java11/org/glassfish/jersey/jetty/http2/connector/JettyHttp2ClientSupplier.java
```diff
@@ -0,0 +1,60 @@
+/*
```
jansupol, line 1: java11 folder does not seem necessary. It can be put among the other files in the java folder. That would make the build (pom file) simpler.

connectors/jetty-http2-connector/src/main/resources/org/glassfish/jersey/jetty/http2/connector/localization.properties
```diff
@@ -0,0 +1,21 @@
...
+# https://www.gnu.org/software/classpath/license.html.
+#
+# SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+#
+
+# {0} - HTTP method, e.g. GET, DELETE
+method.not.supported=Method {0} not supported.
+invalid.configurable.component.type=The supplied component "{0}" is not assignable from Jersey11Client or JerseyWebTarget.
+expected.connector.provider.not.used=The supplied component is not configured to use a Jetty11ConnectorProvider.
+not.supported=Jetty connector is not supported on JDK version less than 11.
```
jansupol: less than 17

connectors/jetty-http2-connector/src/main/resources/org/glassfish/jersey/jetty/http2/connector/localization.properties
```diff
@@ -0,0 +1,21 @@
...
+# Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
+# version 2 with the GNU Classpath Exception, which is available at
+# https://www.gnu.org/software/classpath/license.html.
+#
+# SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+#
+
+# {0} - HTTP method, e.g. GET, DELETE
+method.not.supported=Method {0} not supported.
+invalid.configurable.component.type=The supplied component "{0}" is not assignable from Jersey11Client or JerseyWebTarget.
```
jansupol: There is no Jersey11Client or Jetyy11ConnectorProvider any longer

test-framework/providers/jetty-http2/src/main/java11/org/glassfish/jersey/test/jetty/http2/JettyHttp2TestContainerFactory.java
```diff
@@ -0,0 +1,36 @@
+/*
```
jansupol, line 1: again java11 folder

jansupol (commented):
Can the bom.pom be also updated?

connectors/jetty-http2-connector/src/main/java11/org/glassfish/jersey/jetty/http2/connector/JettyHttp2ClientSupplier.java
senivam: I don't think it will work as expected. Because in the case of the `JettyHttp2ClientSupplier.java` in the separate src folder the resulting jar will look like:

```META-INF/versions/<JDK-version>/<package>/JettyHttp2ClientSupplier.java```

 and for related JDK a proper version will be picked up. However, if the `JettyHttp2ClientSupplier.java` would be in the main src folder it will be located directly in the compiled classes folder `org.glassfish.jersey.jetty.http2.container` (without version separation), and JDK probably won't be able to overwrite it by picking another `JettyHttp2ClientSupplier.java` for a particular JDK version (17+ in our case)

test-framework/providers/jetty-http2/src/main/java11/org/glassfish/jersey/test/jetty/http2/JettyHttp2TestContainerFactory.java
senivam: see my first comment

connectors/jetty-http2-connector/src/main/java11/org/glassfish/jersey/jetty/http2/connector/JettyHttp2ClientSupplier.java
jansupol: That's not how it works. The classloader first checks the META-INF versions and the classes afterwards. See for instance core-common in 2.x. It works for JDK 11, there is no META-INF/versions/8.

senivam: actually there is:

https://github.com/eclipse-ee4j/jersey/tree/2.x/core-common/src/main/java8/org/glassfish/jersey/internal/jsr166

senivam: Yeah, I've got why it shall be here - if we compile 2 similar classes within the same cycle. The compile phase will fail. It is required to exclude from compiling either one or another version of the same class file. It's not possible to compile both files within the same name in a similar compile phase.

senivam: In case of JDK 17+ and no version separation we will have 2 similar class names in one compile phase.

senivam: actually, the resulting jar `jersey-jetty-http2-connector.jar` does not have the `versions/11` folder as well, only `versions/17` but sources structure shall be kept for compilation reasons.

jansupol (approved):

PR #5489, opened 2023-12-06
JerseyEventSink#send throwing non IOExceptions

jbescos:
Relates to https://github.com/jakartaee/rest/issues/1196

Maybe we should catch Throwable instead of Exception.

I don't know what was the reason to catch only IOExceptions, but in the case of Helidon a runtime exception is thrown:
`java.io.UncheckedIOException: java.net.SocketException: Broken pipe`

Anyway, in reactive approach I think we should not throw exceptions in `JerseyEventSink#send`


jbescos:
Wrong branch

PR #5513, opened 2024-01-22
Next iteration towards supporting HK2 & Weld injection

jansupol:


senivam (approved):

paulrutter:
@jansupol i'm not clear on how 4.0.0 now works with injection. Does it still support using HK2 or is Weld now required?

jansupol:
@paulrutter Still HK2, that's the default

PR #5514, opened 2024-01-23
ASCII encode ContentDisposition file name

senivam:


media/multipart/src/main/java/org/glassfish/jersey/media/multipart/ContentDisposition.java
```diff
@@ -211,6 +209,24 @@ protected void addLongParameter(final StringBuilder sb, final String name, final
...
 
+    protected String encodeAsciiFileName(String fileName) {
+        if (fileName == null
+                || (fileName.indexOf('"') == -1
+                && fileName.indexOf('\\') == -1)) {
+            return fileName;
+        }
+
+        final StringBuilder encodedBuffer = new StringBuilder();
+        fileName.chars().forEach(x -> {
```
jansupol: chars() is since JDK 9

senivam: https://docs.oracle.com/javase/8/docs/api/java/lang/CharSequence.html#chars--

jansupol: Interesting, https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/String.html#chars()

jansupol: This would cause the same issue as Buffer#flip(). Once built with JDK 8, executed with JDK 11 it will throw an exception, because JDK 8 String does not have the method. To prevent that, either cast `fileName` to CharSequence, or just do not use the method.

jansupol: See https://jira.mongodb.org/browse/JAVA-2559

senivam: I think that's not the case. The given [JAVA-2559](https://jira.mongodb.org/browse/JAVA-2559) describes differences in return types for the original and overridden methods. When the original ``Buffer#flip`` returns an instance of the ``Buffer`` the overridden one returns an instance of the ``ByteBuffer``. In the case of ``CharSequence#chars`` and ``String#chars``  they both return ``IntStream`` so, this won't cause a failure in the described situation.

senivam: Anyways, reworked to use simpler tooling to be compliant with all cases.

jansupol (approved):

jbescos (approved):

PR #5525, opened 2024-02-04
Adding support for private fields in `@BeanParam`

### LINKED ISSUE #5526: IllegalAccessException in jersey-client-proxy when `@BeanParam` annotated class contains private fields
Getting IllegalAccessException in jersey-client-proxy when `@BeanParam` annotated class contains private fields.

Consider the following example where,
Given that I have a POJO for employees as below:
Given a POJO representing an employee:
```java
public class Employee {
    private String id;
    private String firstName;

    public Employee() {
    }

    public String getId() {
        return id;
    }

    public void setId(String id) {
        this.id = id;
    }

    public String getFirstName() {
        return firstName;
    }

    public void setFirstName(String firstName) {
        this.firstName = firstName;
    }
}
```
We can use the above class as @BeanParam in our service interface as below
```java
    @GET
    @Path("employees")
    List<Employee> findEmployee(@BeanParam FindEmployeeRequest findEmployeeRequest);
```
The above works fine when used in jersey server application, however, in client application it runs into error with following error:
```log
java.lang.IllegalAccessException: class org.glassfish.jersey.client.proxy.RequestParameters cannot access a member of class org.glassfish.jersey.client.proxy.MyBeanParamWithPrivateField with modifiers "private"
	at java.base/jdk.internal.reflect.Reflection.newIllegalAccessException(Reflection.java:394)
	at java.base/java.lang.reflect.AccessibleObject.checkAccess(AccessibleObject.java:714)
	at java.base/java.lang.reflect.Field.checkAccess(Field.java:1156)
	at java.base/java.lang.reflect.Field.get(Field.java:441)
	at org.glassfish.jersey.client.proxy.RequestParameters.addBeanParameter(RequestParameters.java:160)
	at org.glassfish.jersey.client.proxy.RequestParameters.addParameter(RequestParameters.java:139)
	at org.glassfish.jersey.client.proxy.WebResourceFactory.invoke(WebResourceFactory.java:203)
	... 5 more
```

### Expectation
The API models and service interfaces should be compatible with both the jersey based server application and jersey based client proxy so as to enable sharing of common codebase among different projects using the same API.

divyanshshekhar:
### Purpose
Added support for private fields in POJOs used as `BeanParam`

### Issue / Need
The jersey server already supports classes with private fields annotated QueryParam, HeaderParam etc to be used as BeanParam in the JAX-RS API interface. e.g.
Given a POJO representing an employee:
```java
public class Employee {
    private String id;
    private String firstName;

    public Employee() {
    }

    public String getId() {
        return id;
    }

    public void setId(String id) {
        this.id = id;
    }

    public String getFirstName() {
        return firstName;
    }

    public void setFirstName(String firstName) {
        this.firstName = firstName;
    }
}
```
We can use the above class as `@BeanParam` in our service interface as below
```java
    @GET
    @Path("employees")
    List<Employee> findEmployee(@BeanParam FindEmployeeRequest findEmployeeRequest);
```
The above works with Jersey server.
However, the same class and interface cannot be used with `jersey-proxy-client` as of now because the `jersey-proxy-client`, upon encountering fields with private fields having getters failes with following error:
```
java.lang.IllegalAccessException: class org.glassfish.jersey.client.proxy.RequestParameters cannot access a member of class com.dsgaur.example.jersey.api.FindEmployeeRequest with modifiers "private"
	at java.base/jdk.internal.reflect.Reflection.newIllegalAccessException(Reflection.java:394)
	at java.base/java.lang.reflect.AccessibleObject.checkAccess(AccessibleObject.java:714)
	at java.base/java.lang.reflect.Field.checkAccess(Field.java:1156)
	at java.base/java.lang.reflect.Field.get(Field.java:441)
	at org.glassfish.jersey.client.proxy.RequestParameters.addBeanParameter(RequestParameters.java:160)
	at org.glassfish.jersey.client.proxy.RequestParameters.addParameter(RequestParameters.java:139)
	at org.glassfish.jersey.client.proxy.WebResourceFactory.invoke(WebResourceFactory.java:203)
```

This breaks the compatibility of having a common API module with JAX-RS annotated service interfaces and POJOs shared by the server application and the various client applications which might be using the same service. This is a typical usecase.

### Why it fails
It is not checking if the field is accessible before calling `Field.get(beanParam)`, which causes the above error. This statement is reached when a class being used as `@BeanParam` has annotations on the field declaration but the field is private.

### Fix
Check if the field is accessible in the current beanParam object before calling get(). If the field is accessible, get the value, otherwise use the getter for that field if present.

### Affected Versions:
3.x
4.x

Fixed #5526 

Signed-off-by: Divyansh Shekhar Gaur <divyanshshekhar@users.noreply.github.com>

### DIFF
--- a/ext/proxy-client/src/main/java/org/glassfish/jersey/client/proxy/RequestParameters.java
+++ b/ext/proxy-client/src/main/java/org/glassfish/jersey/client/proxy/RequestParameters.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2012, 2023 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2012, 2024 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
--- a/ext/proxy-client/src/main/java/org/glassfish/jersey/client/proxy/RequestParameters.java
+++ b/ext/proxy-client/src/main/java/org/glassfish/jersey/client/proxy/RequestParameters.java
@@ -156,7 +156,7 @@ class RequestParameters {
                 anns.put(ann.annotationType(), ann);
             }
 
-            if (hasAnyParamAnnotation(anns)) {
+            if (field.canAccess(beanParam) && hasAnyParamAnnotation(anns)) {
                 value = field.get(beanParam);
             } else {
                 // get getter annotations if there are no field annotations
--- a/ext/proxy-client/src/test/java/org/glassfish/jersey/client/proxy/MyBeanParamWithPrivateField.java
+++ b/ext/proxy-client/src/test/java/org/glassfish/jersey/client/proxy/MyBeanParamWithPrivateField.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright (c) 2023 Oracle and/or its affiliates. All rights reserved.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0, which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception, which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ */
+
+package org.glassfish.jersey.client.proxy;
+
+import jakarta.ws.rs.*;
+
+/**
+ * @author Divyansh Shekhar Gaur
+ */
+public class MyBeanParamWithPrivateField {
+
+    @QueryParam("privateFieldParam")
+    private String privateFieldParam;
+
+    public MyBeanParamWithPrivateField() {}
+
+    public String getPrivateFieldParam() {
+        return privateFieldParam;
+    }
+
+    public void setPrivateFieldParam(String privateFieldParam) {
+        this.privateFieldParam = privateFieldParam;
+    }
+}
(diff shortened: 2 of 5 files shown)


jansupol (approved):

jansupol:
A universal solution would likely be to set the `field.setAccesible(true)` so that the getter method is not required for the functionality. That's what all the reflection frameworks do.

senivam (approved):

divyanshshekhar:
@jansupol I thought of that as well. However, if we do that, then if someone has defined the annotations on both the field and the getter, then the annotations on the getter would not be detected as it wouldn't go into `else` block.

Let me know your thoughts on this.

jansupol:
You want to override the field value by the getter. But it is inconsistent, the getter overrides the value only for the private fields, and not for the public ones.

divyanshshekhar:
I think if the POJO classes are written as per convention, there should either be public fields or getters but not both. In that case this would work as expected.

ivan-slavchev:
Hi @divyanshshekhar, just FYI, `AccessibleObject.canAccess` throws exception when there are static fields inside the POJO with the current implementation. I don't know if this is a valid case, but maybe it is worth considering. The reason for the exception is that the _obj_ must be null in case of static fields. From the javadoc of the method:

```
     * Test if the caller can access this reflected object. If this reflected
     * object corresponds to an instance method or field then this method tests
     * if the caller can access the given {@code obj} with the reflected object.
     * For instance methods or fields then the {@code obj} argument must be an
     * instance of the {@link Member#getDeclaringClass() declaring class}. For
     * static members and constructors then {@code obj} must be {@code null}.
```

Here is a part of the stack trace for reference:

```
java.lang.IllegalArgumentException: non-null object for public static final int **************
	at java.base/java.lang.reflect.AccessibleObject.canAccess(AccessibleObject.java:494) ~[?:?]
	at de.bonprix.jersey.lib.RequestParameters.addBeanParameter(RequestParameters.java:159)
```

jansupol:
@ivan-slavchev Would you care to create a patch?

ivan-slavchev:
@jansupol here is a patch, I hope it helps:
[use_only_non-static_fields_in_bean_param.patch](https://github.com/eclipse-ee4j/jersey/files/14575363/use_only_non-static_fields_in_bean_param.patch)

PR #5542, opened 2024-02-22
Propagate Future.cancel() to connectors

jansupol:


senivam (approved):

connectors/netty-connector/src/main/java/org/glassfish/jersey/netty/connector/JerseyClientHandler.java
```diff
@@ -157,6 +158,10 @@ protected void notifyResponse() {
 
     @Override
     public void channelRead0(ChannelHandlerContext ctx, HttpObject msg) {
+        if (jerseyRequest.isCancelled()) {
```
jbescos, line 161: Netty javadoc does not provide much details about how often is this method invoked:
https://netty.io/4.0/api/io/netty/channel/SimpleChannelInboundHandler.html

Keep in mind that if this method is invoked once a response is completed, the future will be cancelled, but you will have to wait for the whole time till the response comes, which does not make much sense because the idea of cancelling is to stop execution immediately.

jbescos (approved):
LGTM but I am not sure whether cancelling the request will happen immediately.

PR #5550, opened 2024-03-05
Comply with 4.0 API

jansupol:


bundles/jaxrs-ri/pom.xml
```diff
@@ -314,7 +314,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-shade-plugin</artifactId>
-                <version>3.2.4</version>
+                <version>3.5.2</version>
```
senivam: there is ``${shade.mvn.plugin.version}`` property. Please update it with the current version and use it here or remove the version line from here, it will be taken from the parent.

senivam (approved):

PR #5551, opened 2024-03-05
Update EL impl

jansupol:


senivam:
I think the [expressly-5.0.0.jar](https://repo1.maven.org/maven2/org/glassfish/expressly/expressly/5.0.0/expressly-5.0.0.jar) does not contain ``META-INF/services`` which does not open its ``org.glassfish.expressly.ExpressionFactoryImpl`` as an implementation of ``jakarta.el.ExpressionFactory`` which is critical for us. So, I've submitted a [PR](https://github.com/eclipse-ee4j/expressly/pull/14) there for the issue to be fixed.

senivam (approved):

PR #5589, opened 2024-04-02
Fix thread leak/invasion

### LINKED ISSUE #5588: Each HTTP request unnecessarily starts a new thread, resource leak.
Each instance of `org.glassfish.jersey.jetty.JettyHttpContainer.ResponseWriter` implicitly starts a new thread, because each request creates a new instance of `java.util.Timer` - and these timer threads do not stop. A big request rate can easily start _thousands_ of timer threads.

From the javadoc of `java.util.Timer`:
```
Implementation note: All constructors start a timer thread.
```
and:
```
After the last live reference to a Timer object goes away and all outstanding tasks have completed execution, the timer's task execution thread terminates gracefully (and becomes subject to garbage collection). However, this can take arbitrarily long to occur.
```

So I think that creating a new `j.u.Timer` instance for every request is not a great solution. A better option would be to have only one timer per container.

snazy:
Jersey/Jetty, at least in the 3.1 version line, creates one thread for each HTTP request. This behavior was introduced with #5372 and seems not present in the 2.x or 3.x versions of Jersey.

From the javadoc of `java.util.Timer`:
```
Implementation note: All constructors start a timer thread.
...
After the last live reference to a Timer object goes away and all outstanding tasks have completed execution, the timer's task execution thread terminates gracefully (and becomes subject to garbage collection). However, this can take arbitrarily long to occur.
```
It is fair to assume that "arbitrarily long" may also mean _never_, in case GC never runs.

This change replaces the timer & thread per request with a `ScheduledExecutorService` instance per `JettyHttpContainer`.

Also changed the set-timeout mechanism to use `System.nanoTime()` instead of `System.currentTimeMillis()`, because the latter is prone to wall-clock drift and can result into wrong timeout values.

Fixes #5588

### DIFF
--- a/containers/jetty-http/src/main/java17/org/glassfish/jersey/jetty/JettyHttpContainer.java
+++ b/containers/jetty-http/src/main/java17/org/glassfish/jersey/jetty/JettyHttpContainer.java
@@ -23,10 +23,13 @@ import java.net.URISyntaxException;
 import java.security.Principal;
 import java.util.List;
 import java.util.Map;
-import java.util.Timer;
-import java.util.TimerTask;
 import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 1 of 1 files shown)


snazy:
(Signed the ECA, but it doesn't show up yet?)

senivam (approved):

wendigo:
@senivam is this going to land in the next 3.1.x release?

jansupol (approved):

senivam:
@wendigo, it's heading there

PR #5597, opened 2024-04-09
merge of the actual 3.1 into 4.0

senivam:


bundles/apidocs/pom.xml
```diff
@@ -294,12 +294,14 @@
                             <sourceFileExcludes>
                                 <fileExclude>META-INF/versions/21/org/glassfish/jersey/helidon/connector/*.java</fileExclude>
                                 <fileExclude>org/glassfish/jersey/helidon/connector/*.java</fileExclude>
+                                <fileExclude>org/glassfish/jersey/jetty/*.java</fileExclude>
```
jansupol: Is this correct? Jetty is no longer multirelease here

senivam: fixed. Excludes which were used due to multi-release jars from previous releases are removed.

jansupol (approved):

PR #5604, opened 2024-04-11
Prevent Jackson failing while loading Modules (classloader issues)

jansupol:

PR #5608, opened 2024-04-15
ObjectMapper.findModules throws Error

jbescos:
We need to catch Throwable because it will throw `java.util.ServiceConfigurationError`

Related to https://github.com/eclipse-ee4j/jersey/pull/5604


jansupol (approved):
Nice catch!

senivam (approved):

PR #5610, opened 2024-04-16
Wiremock version update

senivam:


jansupol (approved):

jansupol:
Why was it not put to 2.x?

senivam:
I've tried, but it requires jakartified servlet

senivam:
Actually, Jetty 11 (so, JDK 11+) and jakartified servlet. It's there from the Wiremock 3.0 (the very beginning of the new Wiremock). Here is [my attempt](https://ci.eclipse.org/jersey/job/ci-cd-integration/job/PR-5609/) to use it with the 2.x

PR #5620, opened 2024-04-22
Add ability to configure the queue capacity for ChunkedOutput

paulrutter:
Adding constructor which sets the queue capacity.

This allows for calling write from different threads until the queue is full. Subsequent write calls will block until the queue is emptied.

This change allows to prevent memory issues in case of slow clients. Basically, a backpressure mechanism based on the queue size.

This change provides a way to call write on ChunkedOutput from multiple threads to improve throughput, while preventing memory issues in case of slow clients. The queue is bounded and thus cannot grow endlessly.


paulrutter:
I will put in a new PR under a different account for the ECA.

paulrutter:
Replaced by https://github.com/eclipse-ee4j/jersey/pull/5621

PR #5638, opened 2024-05-03
Set reasonable timeout and disable concurrent builds

fredg02:
Apply Jenkins/CI best practices and avoid builds hanging for days.

### DIFF
--- a/etc/jenkins/Jenkinsfile_ci_build
+++ b/etc/jenkins/Jenkinsfile_ci_build
@@ -2,7 +2,8 @@ pipeline {
     agent none
 
     options {
-          timeout(time: 30, activity: true, unit: 'HOURS')
+          timeout(time: 10, activity: true, unit: 'HOURS')
+          disableConcurrentBuilds()
     }
 
     stages {


jansupol:
@fredg02 Why do we want to disable the concurrent builds? Does it mean that even if we have the available worker, the available worker won't perform the build on another JDK?

senivam:
`disableConcurrentBuilds()` means that if a job is already running (due to a PR for example), another job won't be triggered until the actual job is finished. It can be multithreaded. All workers will be devoted to that actual job run. Other jobs will wait in the queue.

senivam:
However, I do not see anything bad if a part of a build in the queue starts before the whole previous build run is finished. This reduces timeout time for waiting in the queue. 30 hours is set because there may be many pushes/pull requests into a branch and all of them are preferable to be processed. If timeout triggers, the last job in the queue will be killed even before it has a chance to be started. Recently I've checked that if the build hung in the queue for some reason (run out of executors for example) it gets aborted even without timeout triggered. And if there are no executors and builds hung, I would prefer to have them in the queue just to notice something is wrong and report an issue to Eclipse. Otherwise, I'm not able to track if all is going well.  30 hours is enough for me to check the pipeline.

PR #5646, opened 2024-05-16
merge of the actual 3.0 into the 3.1

senivam:


jbescos (approved):

pom.xml
```diff
@@ -2272,7 +2176,7 @@
         <xerces.version>2.12.2</xerces.version>
```
jansupol, line 2176: <wiremock.version>3.5.2</wiremock.version> did not make it to 3.1?

senivam: Unfortunately, when it comes to the 3.1, the latest wiremock does not work with MP TCKs. In the scope of the current merge, I was not able to figure out what is the reason, but the fact remains that only Tom Akehurst's version works properly in that configuration. So, I had to keep it as is.

jansupol (approved):

PR #5649, opened 2024-05-21
MAX_HEADER_SIZE for the Netty connector

senivam:


connectors/netty-connector/src/main/java/org/glassfish/jersey/netty/connector/NettyClientProperties.java
```diff
@@ -129,4 +129,56 @@ public class NettyClientProperties {
...
+            MAX_CHUNK_SIZE = "jersey.config.client.netty.maxChunkSize";
+
+    /**
+     * Default chunk size for Netty Connector.
+     * Taken from {@link io.netty.handler.codec.http.HttpClientCodec#HttpClientCodec(int, int, int)}
+     *
+     * @since 2.44
+     */
+    public static final Integer
+        DEFAULT_CHUNK_SIZE = 8192;
```
jansupol, line 182: Does it make sense to introduce a new property name which we have MessageProperties.IO_BUFFER_SIZE for already?

jansupol: Maybe `MessageProperties.IO_BUFFER_SIZE` is a different thing

jansupol (approved):
It would be good to document the new properties

jbescos (approved):

PR #5664, opened 2024-05-27
More release checks

jansupol:


senivam (approved):

tests/release-test/src/main/java/org/glassfish/jersey/test/artifacts/ClassVersionChecker.java
```diff
@@ -27,7 +27,14 @@
 class ClassVersionChecker {
     static TestResult checkClassVersion(JarFile jar, JarEntry entry, Properties properties) throws IOException {
         final String jerseyVersion = MavenUtil.getJerseyVersion(properties);
-        final int minVersion = jerseyVersion.startsWith("3.1") ? 11 : 8;
+        final int minVersion;
+        if (jerseyVersion.startsWith("4")) {
+            minVersion = 17;
+        } else if (jerseyVersion.startsWith("3.1")) {
```
jbescos, line 33: In case we create a new 3.2 version we will need to remember to update this or it will fall to version 8. Maybe it is preferable to use a regular expression, something like: `^3\.[1-9]+` (please test it because I am not good with regular expressions).

jbescos (approved):

PR #5669, opened 2024-06-03
prevent calling ServletRequest#getInputStream if FILTER_FORWARD_ON_404

### LINKED ISSUE #4867: Corrupted POST data of type x-www-form-urlencoded after forwardOn404
Dear community, In the most recent jersey 2.35 I observed a possible bug that causes my application to fail:

After POSTing application/x-www-form-urlencoded content to my application AFTER applying a org.glassfish.jersey.servlet.ServletContainer filter along with forwardOn404, the first byte is always missing.

The problem seems to be caused by calling containerRequest.hasEntity() in WebComponent.filterFormParameters(). I guess that EntityInputStream.isEmpty() consumes one byte from the original request input stream to check if there is some content but, if content (one byte) is found, it does only push it back to EntityInputStream.input but not the original input stream of the servlet request that will be forwarded later on.

The jersey code around seems to be quite complex, so fixing won't be that easy for me. It would be great if you could fix this issue soon. Thank you.

(Meanwhile I'll try to add two more filters to circumvent the problem: The first one changes the content type to application/x-www-form-urlencoded_fixme and the second reverts it afterwards to application/x-www-form-urlencoded. This should hopefully work.)

jansupol:
Introduce InputStreamWrapper

### DIFF
--- a/connectors/apache-connector/src/main/java/org/glassfish/jersey/apache/connector/ApacheConnector.java
+++ b/connectors/apache-connector/src/main/java/org/glassfish/jersey/apache/connector/ApacheConnector.java
@@ -60,6 +60,7 @@ import org.glassfish.jersey.client.innate.ClientProxy;
 import org.glassfish.jersey.client.innate.http.SSLParamConfigurator;
 import org.glassfish.jersey.client.spi.AsyncConnectorCallback;
 import org.glassfish.jersey.client.spi.Connector;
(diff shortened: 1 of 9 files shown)


senivam (approved):

jansupol:
Fixes #4867

jbescos (approved):

PR #5688, opened 2024-07-02
Wrap call of deprecated HttpServletResponse#setStatus into try catch

jansupol:


jansupol:
This PR should be merged into 3.x separately, as it is only for 3.0, but not for 3.1 where it makes no sense.

senivam (approved):

jbescos (approved):

PR #5698, opened 2024-07-15
Last-Modified header is garbled when accessing wadl document on Japanese locale

vavishal:
Last-Modified header is garbled when accessing wadl document on Japanese locale

Also calling javax.ws.rs.core.Response.getLastModified() on Japanese locale throws "org.glassfish.jersey.message.internal.HeaderValueException: Unable to parse "Last-Modified" header value:"

So, Locale.US was added as default locale in the fix

The testcase also exists at jersey/tests/e2e-server/src/test/java/org/glassfish/jersey/tests/e2e/server/wadl/WadlResourceTest.java

Signed-off-by: Vaibhav Vishal <vaibhav.vishal@oracle.com>


jbescos (approved):
LGTM

senivam (approved):

PR #5703, opened 2024-07-23
merge of the actual 3.0 into the 3.1

senivam:


core-server/pom.xml
```diff
@@ -252,6 +258,16 @@
     </dependencies>
 
     <profiles>
+        <profile>
```
jansupol: This profiles should go away eventually

pom.xml
```diff
@@ -2161,10 +2161,16 @@
         <xmlunit.version>2.10.0</xmlunit.version>
         <httpclient.version>4.5.14</httpclient.version>
         <httpclient5.version>5.3.1</httpclient5.version>
-        <jackson.version>2.17.0</jackson.version>
+        <jackson.version>2.17.1</jackson.version>
         <javassist.version>3.30.2-GA</javassist.version>
         <jboss.logging.version>3.5.3.Final</jboss.logging.version>
+        <jboss.logging.8.version>3.4.3.Final</jboss.logging.8.version>
+        <jersey1.version>1.19.3</jersey1.version>
```
jansupol: Do next:
The last is 1.19.4, check!
jboss logging 8 property, <jboss.vfs.jdk8.version>3.2.17.Final</jboss.vfs.jdk8.version> -> remove

jansupol (approved):

pom.xml
senivam: actually, those versions (*.jersey1, *.jdk8.*) should be removed from the 3.1 version, they are only valid for 2.x and 3.0. So, I've removed them for the 3.1

core-server/pom.xml
senivam: removed

PR #5705, opened 2024-07-24
Allow ChunkedInput#close to close the underlying stream

### LINKED ISSUE #5679: ChunkedInput.close is not sufficient to release client connections
Hi folks,

We observed a connection leak in our dropwizard service, using `jersey-client 2.41` to call an endpoint with chunked encoding. The code receives a `ChunkedInput`, processes it, then closes it. We expected this would be sufficient to close the underlying input stream and release the connection, but open connections just increase until the pool is exhausted.

I reproduced the issue [here](https://github.com/eddsteel/zipped-chunked) and raised a dropwizard [issue](https://github.com/dropwizard/dropwizard/issues/8884). Steffen's [reply](https://github.com/dropwizard/dropwizard/issues/8884#issuecomment-2169989863) clearly laid out the reasons for the behaviour (I won't add noise here). This situation seems to be by design in Jersey, specifically `jersey-common` 2.41.

We can release the resources by returning both a `ChunkedInput` and a `Response` object, and closing both. But this seems inconsistent with other parts of the API and the spec where closing an underlying input will have the same effect. I'm not necessarily asking you to change anything, but would be interested in your thoughts on what Steffen said.

Thanks!

jansupol:


senivam (approved):

jbescos (approved):

jansupol:
2.45, 3.0.15, 3.1.8, 4.0.0-M2

PR #5706, opened 2024-07-25
Propagate WebApplicationException from SseEventSource to provided error handler

### LINKED ISSUE #5693: `JerseySseEventSource` ignores `onError` callback supplied to `register()` method, making it impossible to surface HTTP/network errors
It is possible for an SSE endpoint to generate an error.
It can respond with 403 Forbidden or 401 Unauthorized if the user is not authorized or authenticated, or 404 Not Found if the client is communicating with an older version of the server that does not support the SSE endpoint.
Or there can be 5xx or some network error if there are other issues.

In all of these cases, it is useful to surface the actual error so that the user can understand what the problem is.
But the `JerseySseEventSource` seems to ignore the `onError` callback completely.

I have code that looks like this:

```java
try (SseEventSource sseSource = SseEventSource.target(target).build()) {
    // create SSE client and register consumer
    SettableFuture<InboundSseEvent> firstMessage = SettableFuture.create();
    sseSource.register(Functions.asConsumer(sseEvent -> {
        firstMessage.set(sseEvent);
        // do other stuff...
    }), e -> {
        firstMessage.setException(e);
    }, () -> {
        firstMessage.cancel(true);
    });
    // start SSE client and wait for initial message
    sseSource.open();
    firstMessage.get();
    // do other stuff...
} catch (ExecutionException e) {
    e.getCause().printStackTrace();
}
```

I'm expecting the `firstMessage.get()` invocation to generate an `ExecutionException` that carries the underlying cause of the SSE client failing if the endpoint cannot be used to establish an SSE connection, e.g. an exception containing HTTP status and response data. Instead only the `onClose` get invoked, without any useful information being surfaced.

Is this intentional, or is there something that I'm missing?

jansupol:


senivam (approved):

jbescos (approved):

jansupol:
2.45, 3.0.15, 3.1.8, 4.0.0-M2

PR #5744, opened 2024-09-13
Create delegate on InflaterInputStream and DeflaterOutputStream

ZongoForSpeed:
Those delegate are here to be sure the Inflater and Deflater resources are correctly released

### DIFF
--- a/core-common/src/main/java/org/glassfish/jersey/message/internal/ClosingDeflaterOutputStream.java
+++ b/core-common/src/main/java/org/glassfish/jersey/message/internal/ClosingDeflaterOutputStream.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright (c) 2024 Oracle and/or its affiliates. All rights reserved.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0, which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception, which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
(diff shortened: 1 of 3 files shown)


jansupol:
Please see our wiki about missing [ECA](https://github.com/eclipse-ee4j/jersey/wiki/Missing-ECA,-Missing-Sign-off-by).

jansupol:
The behaviour is the same in JDK 8, the Deflater and Inflater are not called there, either.

core-common/src/main/java/org/glassfish/jersey/message/internal/ClosingDeflaterOutputStream.java
```diff
@@ -0,0 +1,83 @@
...
+ * After Java 9, in the class Deflater the method finalize doesn't call end anymore
+ * The method end must be called explicitly. But when using the constructor
+ * DeflaterOutputStream(OutputStream out, Deflater def), the Deflater.end() method will
+ * not be called when closing the stream.
+ * This lead to memory leaks in the off-heap.
+ *
+ * @see java.util.zip.Deflater
+ * @see java.util.zip.DeflaterOutputStream
+ */
+public final class ClosingDeflaterOutputStream extends OutputStream {
```
jansupol, line 34: I would prefer to make those classes ~package-private~ private.

ZongoForSpeed:
@jansupol in JDK 8, the finalizer of Deflater & Inflater contains a call the end() method that will release correctly the Deflater/Inflater.
You in the source code of the Deflater/Inflater : 
in Deflater.java (JDK 9)
`
    @Deprecated(since="9", forRemoval=true)
    protected void finalize() {}
`
in Deflater.java (JDK 8)
`
    /**
     * Closes the compressor when garbage is collected.
     */
    protected void finalize() {
        end();
    }
`

core-common/src/main/java/org/glassfish/jersey/message/internal/ClosingDeflaterOutputStream.java
ZongoForSpeed: I can't make this class package-private and keep it in the internal package. 
Do you want me to move those classes in the package 'org.glassfish.jersey.message' or to create inner classes ?

Issue #5738, opened 2024-09-03
Jersey update from 3.1.3 to 3.1.4 slows down our external service response times drastically


rehmatt:
Hi,

we've recently updated our services from jersey `3.1.2` to `3.1.5` and encountered drastically worse response times for our external service calls which are using jerseys `HttpUrlConnector`.

With downgrading to `3.1.3` everything was fine again, but updating to `3.1.4` again let the response times rise by 50% to 120%.

Having a look on the changes between `3.1.3` and `3.1.4` I saw the following PR: https://github.com/eclipse-ee4j/jersey/pull/5359.
It introduced a synchronized block inside `getConnection(URL url, Proxy proxy)`:

```
default HttpURLConnection getConnection(URL url, Proxy proxy) throws IOException {
    synchronized (this){
        return (proxy == null) ? getConnection(url) : (HttpURLConnection) url.openConnection(proxy);
    }
}
```

If I interpret it correctly this will lead to only one thread being able to open a connection with the connector. As we have many threads processing incoming requests with one connector per external service, it might happen that there are many threads which want to get a connection, but only one at a time is possible. 

<img width="2359" alt="Screenshot 2024-09-03 at 15 25 05" src="https://github.com/user-attachments/assets/b3df91ab-26e9-4c29-8c56-b9b9b334a938">

Is this already a known issue? Is my assumption regarding the synchronized block correct?

fwippe:
Just to put this into perspective: If I'm not mistaken, the fact that `getConnection` is synchronized on `HttpUrlConnector` ultimately means that only one connection at a time may be created **across all services handled by Jersey**. This is potentially a catastrophic performance regression.

jankadel:
Great finding! I was wondering why my services became slower when I upgraded Jersey.

jansupol:
Should be fixed by #5794.

jansupol:
Jersey 3.1.10 has been released

PR #5746, opened 2024-09-23
Jersey update from 3.1.3 to 3.1.4 slows down our external service res…

jbescos:
https://github.com/eclipse-ee4j/jersey/issues/5738

It locks per URL, so the client can open different URLs in parallel and it locks if the URL is the same.


jbescos:
I created it to 2.x: https://github.com/eclipse-ee4j/jersey/pull/5749

PR #5751, opened 2024-09-30
test(flaky): fix paramter type assert in testBothAnnotatedConstructor test case

varsha-cl:
### Description

The `testBothAnnotatedConstructor` test in `CachedConstructorAnalyzerTest` fails randomly because constructor selection is non-deterministic when multiple constructors have the same number of parameters. Specifically, `BothAnnotatedConstructor` has two single-parameter constructors: one accepting a String and the other an Integer. Since `getDeclaredConstructors` does not guarantee order, the test may select either constructor, leading to intermittent failures.

### Fix

Adjust the test to validate that the parameter type of the returned constructor is either `String` or `Integer`, accommodating both possible outcomes.

### How Has This Been Tested

Identified this test to be flaky using [NonDex](https://github.com/TestingResearchIllinois/NonDex) tool (which explores different behaviors of under-determined APIs and reports test failures). Here are the commands to run this test case with NonDex tool

```
# Install the dependencies
mvn install -pl incubator/cdi-inject-weld -am -DskipTests

# Run the test with NonDex
mvn -pl incubator/cdi-inject-weld \
    edu.illinois:nondex-maven-plugin:2.1.7:nondex \
    -Dtest=org.glassfish.jersey.inject.weld.internal.injector.CachedConstructorAnalyzerTest#testBothAnnotatedConstructor
```
Here is the screenshot of the output generated by the NonDex tool 
Screenshot 2024-09-02 at 12 05 07 PM

In the above case, we can see that the test case fails intermittently, in the first instance it passed successfully, however in the second instance it failed with error `CachedConstructorAnalyzerTest.testBothAnnotatedConstructor:113 expected: <java.lang.Integer> but was: <java.lang.String>`

### DIFF
--- a/incubator/cdi-inject-weld/src/test/java/org/glassfish/jersey/inject/weld/internal/injector/CachedConstructorAnalyzerTest.java
+++ b/incubator/cdi-inject-weld/src/test/java/org/glassfish/jersey/inject/weld/internal/injector/CachedConstructorAnalyzerTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021, 2022 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2021, 2022, 2024 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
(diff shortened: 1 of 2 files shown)


jansupol:
@varsha-cl This looks correct. While we are not seeing this issue, I assume it can occur.

This is the same as in the other [CachedConstructorAnalyzerTest](https://github.com/eclipse-ee4j/jersey/blob/2.x/inject/cdi2-se/src/test/java/org/glassfish/jersey/inject/cdi/se/injector/CachedConstructorAnalyzerTest.java) test (yes, we duplicate the test in the incubator module because of its immaturity for the time being) which should be updated the same way.

The copyright year needs to be updated to 2024 before this PR's tests pass.

varsha-cl:
> @varsha-cl This looks correct. While we are not seeing this issue, I assume it can occur.
> 
> This is the same as in the other [CachedConstructorAnalyzerTest](https://github.com/eclipse-ee4j/jersey/blob/2.x/inject/cdi2-se/src/test/java/org/glassfish/jersey/inject/cdi/se/injector/CachedConstructorAnalyzerTest.java) test (yes, we duplicate the test in the incubator module because of its immaturity for the time being) which should be updated the same way.
> 
> The copyright year needs to be updated to 2024 before this PR's tests pass.

Thank you the feedback. I have updated the year and replicated the changes in other test file as well

senivam (approved):

jansupol (approved):

Issue #5675, opened 2024-06-11
AsyncContext completion attempt after AsyncListener.onError is called


wendigo:
We've observer an issue with Jersey 3.1.+ running in the Jetty EE10 container that applies to HTTP/2 protocol.

When client disconnects in the middle of the request (due to a timeout or whatever), it sends a RST_STREAM frame that on the server side, finishes an AsyncContext with an error through the AsyncListener.onError. It seems that Jersey doesn't understand this condition, and still tries to finish an AsyncContext when the AsyncResponse.complete is called. Since Jetty recycles request/response objects on AsyncContext error, there is no response to write output to, and the jersey code generates a non-actionable warning.

We've work around this issue by implementing https://github.com/trinodb/trino/pull/22315 that seems to do a trick. I'd like to see a permanent solution on the Jersey side instead.

wendigo:
@lorban anything else to add here?

lorban:
@wendigo no; you accurately described the problem.

jansupol:
It looks like creating the test case is the most difficult part here

wendigo:
@jansupol is there a plan to fix this issue?

jansupol:
Yes, we plan to check on the possibilities for this.

jbescos:
I don't know about this part but I would like to take a look.

My understanding is that the issue is happening in the server side. The server receives `cancel_stream_error` and in this case, the server should close the connection. From what you say, Jersey is invoking `#onError` and later it invokes `#complete`. The fix is about not invoking `#complete`. Am I right?.

Are you able to print the stacktrace when `#onError` is invoked?.

wendigo:
Actually async context can complete with error on a different thread than the competing thread that will try to complete with a result. It seems that this isn't synchronized in any way.

wendigo:
In Jetty it's pretty visible because when the context is completed (successfully or not) the resources are recycled which causes errors to be logged (request/response doesn't exist)

joakime:
> Are you able to print the stacktrace when `#onError` is invoked?.

Here you go, this is the stacktrace received by `AsyncContext.onError(Throwable)`.

In this case, the EOF results in a completed (and set to committed) response (if not completed yet).
[...]

PR #5773, opened 2024-10-18
 AsyncContext completion attempt after AsyncListener.onError is called #5675

jbescos:
[… #5675](https://github.com/eclipse-ee4j/jersey/issues/5675)

Draft for now

### DIFF
--- a/core-server/src/main/java/org/glassfish/jersey/server/ServerRuntime.java
+++ b/core-server/src/main/java/org/glassfish/jersey/server/ServerRuntime.java
@@ -277,6 +277,14 @@ public class ServerRuntime {
         }
     }
 
+    static boolean isIOException(Throwable t) {
+        boolean ioException = false;
+        while (t != null && !(ioException = t instanceof IOException)) {
+            t = t.getCause();
+        }
+        return ioException;
+    }
+
     /**
      * Get the Jersey server runtime background scheduler.
      *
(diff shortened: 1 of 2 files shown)


wendigo:
I've tested this change (backported to 3.1) in our project (trinodb/trino) with HTTP/2 enabled and errors/warnings are gone. I've been running our stress test for over 30 minutes so definitely an improvement :)

wendigo:
Can we get this into 3.1? 🙏🏻

core-server/src/main/java/org/glassfish/jersey/server/ServerRuntime.java
```diff
@@ -669,11 +669,18 @@ public OutputStream getOutputStream(final int contentLength) throws IOException
...
             } catch (final Throwable ex) {
                 if (response.isCommitted()) {
-                    /**
-                     * We're done with processing here. There's nothing we can do about the exception so
-                     * let's just log it.
-                     */
-                    LOGGER.log(Level.SEVERE, LocalizationMessages.ERROR_WRITING_RESPONSE_ENTITY(), ex);
+                    if (ex.getCause() instanceof IOException) {
+                        skipFinally = true;
+                        LOGGER.log(Level.WARNING, "Response was sent, but there is IO issue", ex);
```
joakime, line 682: There was an issue after the response was committed.
You don't know if the response was sent or not. (it could be sitting in a buffer somewhere locally, not even sent yet)
You just know you cannot change the response anymore.

core-server/src/main/java/org/glassfish/jersey/server/ServerRuntime.java
```diff
@@ -669,11 +669,18 @@ public OutputStream getOutputStream(final int contentLength) throws IOException
 
             } catch (final Throwable ex) {
                 if (response.isCommitted()) {
-                    /**
-                     * We're done with processing here. There's nothing we can do about the exception so
-                     * let's just log it.
-                     */
-                    LOGGER.log(Level.SEVERE, LocalizationMessages.ERROR_WRITING_RESPONSE_ENTITY(), ex);
+                    if (ex.getCause() instanceof IOException) {
```
joakime: The `ex instanceof IOException` is relevant here too.

joakime: Not sure if you want to check all nested causes and suppressed throwables for IOException too?

core-server/src/main/java/org/glassfish/jersey/server/ServerRuntime.java
```diff
@@ -669,11 +669,18 @@ public OutputStream getOutputStream(final int contentLength) throws IOException
 
             } catch (final Throwable ex) {
                 if (response.isCommitted()) {
```
joakime, line 679: This `if (response.isCommitted()) {` check is actually kinda racy.
The response could be failed in a different thread (eg: from an HTTP/2 goaway), causing the response to be flagged as committed as part of its stream close handling.
This could be false when you check, and then a microsecond later its true.

jbescos: I can refactor that, but in any case, if it is IOException it will throw up the exception (it does not matter whether the response was committed or not).

jbescos: I think if there is IOException it does not really matter if the response was sent or not, because we cannot respond. Probably we need to skip the finally block in any case for this type of exception.

PR #5784, opened 2024-10-29
 Jersey 3.1.9: java.lang.NoSuchMethodException: jakarta.inject.Inject.value() #5782 

### LINKED ISSUE #5782: Jersey 3.1.9: java.lang.NoSuchMethodException: jakarta.inject.Inject.value()
When bootstrapping Jersey 3.1.9 on Java SE 17 I came across this exception logged, which IMHO makes no sense to log at all:

```
org.glassfish.jersey.model.Parameter getValue
FINER: Unable to get the jdk.proxy2.$Proxy27 annotation value property
java.lang.NoSuchMethodException: jakarta.inject.Inject.value()
	at java.base/java.lang.Class.getMethod(Class.java:2277)
	at org.glassfish.jersey.model.Parameter.getValue(Parameter.java:452)
	at org.glassfish.jersey.model.Parameter.create(Parameter.java:269)
	at org.glassfish.jersey.server.model.Parameter.create(Parameter.java:68)
	at org.glassfish.jersey.server.model.IntrospectionModeller.checkResourceClassFields(IntrospectionModeller.java:206)
	at org.glassfish.jersey.server.model.IntrospectionModeller.doCreateResourceBuilder(IntrospectionModeller.java:118)
	at org.glassfish.jersey.server.model.IntrospectionModeller$1.call(IntrospectionModeller.java:91)
	at org.glassfish.jersey.server.model.IntrospectionModeller$1.call(IntrospectionModeller.java:88)
	at org.glassfish.jersey.internal.Errors.process(Errors.java:292)
	at org.glassfish.jersey.internal.Errors.process(Errors.java:274)
	at org.glassfish.jersey.internal.Errors.processWithException(Errors.java:232)
	at org.glassfish.jersey.server.model.IntrospectionModeller.createResourceBuilder(IntrospectionModeller.java:88)
	at org.glassfish.jersey.server.model.Resource.from(Resource.java:781)
	at org.glassfish.jersey.server.ResourceBagConfigurator.init(ResourceBagConfigurator.java:55)
	at org.glassfish.jersey.server.ApplicationHandler.initialize(ApplicationHandler.java:347)
	at org.glassfish.jersey.server.ApplicationHandler.lambda$initialize$1(ApplicationHandler.java:309)
	at org.glassfish.jersey.internal.Errors.process(Errors.java:292)
	at org.glassfish.jersey.internal.Errors.process(Errors.java:274)
	at org.glassfish.jersey.internal.Errors.processWithException(Errors.java:232)
	at org.glassfish.jersey.server.ApplicationHandler.initialize(ApplicationHandler.java:308)
	at org.glassfish.jersey.server.ApplicationHandler.<init>(ApplicationHandler.java:273)
	at org.glassfish.jersey.server.ApplicationHandler.<init>(ApplicationHandler.java:260)
	at org.glassfish.jersey.grizzly2.httpserver.GrizzlyHttpContainer.<init>(GrizzlyHttpContainer.java:310)
	at org.glassfish.jersey.grizzly2.httpserver.GrizzlyHttpServerFactory.createHttpServer(GrizzlyHttpServerFactory.java:142)
```

The reason for this exception is that [`@jakarta.inject.Inject`](https://jakarta.ee/specifications/platform/10/apidocs/jakarta/inject/inject) does not have a method called `value`, but Jersey apparently expects to find it, as shown in the source code:

https://github.com/eclipse-ee4j/jersey/blob/aafb9c413760cea7ba5e29ab23c2a555ffb25e7c/core-common/src/main/java/org/glassfish/jersey/model/Parameter.java#L450-L464

It is not a bug nor an exceptional case that an annotation does not have a `value` method, so I wonder what the benefit of logging this quite normal case shall have? It is quite normal to not find that method.

BTW, as exception handling is slow in Java, it might be a performance improvement to first check the result of `getMethods()` before actually requesting the `value` method.

jbescos:
https://github.com/eclipse-ee4j/jersey/issues/5782

### DIFF
--- a/core-common/src/main/java/org/glassfish/jersey/model/Parameter.java
+++ b/core-common/src/main/java/org/glassfish/jersey/model/Parameter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2010, 2024 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
@@ -447,14 +447,20 @@ public class Parameter implements AnnotatedElement {
                 parameterClass);
     }
 
-
     private static String getValue(Annotation a) {
         try {
-            Method m = a.annotationType().getMethod("value");
-            if (m.getReturnType() != String.class) {
-                return null;
+            Method[] methods = a.annotationType().getMethods();
+            for (Method method : methods) {
+                if ("value".equals(method.getName())) {
+                    if (method.getReturnType() != String.class) {
+                        return null;
+                    } else {
+                        return (String) method.invoke(a);
+                    }
+                }
             }
-            return (String) m.invoke(a);
+            LOGGER.log(Level.FINER, () ->
+                    String.format("Unable to get the %s annotation value property", a.getClass().getName()));
         } catch (Exception ex) {
             if (LOGGER.isLoggable(Level.FINER)) {
                 LOGGER.log(Level.FINER,


jansupol:
@jbescos I assume this is better performance-wise since the try-catch clause is slow in general. On the other hand, this code tests all the methods and iterates through them. I expect only a very limited number of methods and hence a speedup, but...have you tried to test the performance a bit with this (I do not mean a Jersey test of any kind, more a loop of numerous operations checking for the "value")? The code around the parameter class is utilized quite often in the runtime and can have a measurable impact.

jbescos:
I am running it, and I am surprised that the `#getMethod` performs only 25% slower when there are only 3 methods. When there are many methods (for example when you iterate the methods in String class), in the worst scenario, iterating all the methods is around 300% slower.

Till this point, it looks better to use `#getMethod`.

However, if exception is thrown because there is no such method, `#getMethod` performs 1200% slower than iterating 3 methods, and  200% slower iterating the methods of a String.

jansupol (approved):

senivam (approved):

Issue #5783, opened 2024-10-29
Broken pipe Exception from Jersey layer while closing response in ServerRuntime$Responder.writeResponse()  not  handled or re-thrown


logovind:
Exception from Jersey layer while closing response is not  handled and  implications  can very depending on the  implementation of ResponseWriter.

For Eg. Helidon is using a custom implementation of the container response writer([JAXRSResponseWriter](https://github.com/helidon-io/helidon/blob/helidon-4.1.x/microprofile/server/src/main/java/io/helidon/microprofile/server/JaxRsService.java)) which is using a CountdownLatch that is initialized with the value of 1 and decremented to 0 on commit and failure methods. If these methods do not get called then await will hang.  

we saw this behavior of await hanging when Broken pipe Exception occurred while closing the response in Jersey code.  More details available in https://github.com/helidon-io/helidon/issues/9442

In Jersey, the below code segment in ServerRuntime.writeResponse() method does not handle or re-throw the exception while closing the response. It simply logs the messages and ignores the exception.  

https://github.com/eclipse-ee4j/jersey/blob/3.1.9/core-server/src/main/java/org/glassfish/jersey/server/ServerRuntime.java#L467
 
```
ServerRuntime.writeResponse()
 
if (close) {
                        try {
                            response.close();
                        } catch (final Exception e) {
                            LOGGER.log(Level.SEVERE, LocalizationMessages.ERROR_CLOSING_COMMIT_OUTPUT_STREAM(), e);
                        }
                    }
 
```
For successful cases, the implementation of the close calls the “getResponseWriter().commit()” which would have decrement the countDownLatch to 0 and make sure that await() would not block. 
 
https://github.com/eclipse-ee4j/jersey/blob/3.1.9/core-server/src/main/java/org/glassfish/jersey/server/ContainerResponse.java#L404
```
ContainerResponse.close()
public void close() {
        if (!closed) {
            closed = true;
            messageContext.close();
            requestContext.getResponseWriter().commit();
            requestContext.setWorkers(null);
        }
    }
```
 However, for the error cases, failure() method never gets called and causes the await() to block in the case of Helidon use case. The broken pipe exception is thrown from  "messageContext.close()". 

stack below:

```
java.io.UncheckedIOException: java.net.SocketException: Broken pipe
  at io.helidon.common.buffers.GrowingBufferData.writeTo(GrowingBufferData.java:69)
[...]

PR #5786, opened 2024-11-05
Broken pipe Exception from Jersey layer while closing response in ServerRuntime.writeResponse() not handled or re-thrown #5783

jbescos:
https://github.com/eclipse-ee4j/jersey/issues/5783

That issue happened because Helidon throws `UncheckedIOException` instead of `IOException`.

If we catch `UncheckedIOExceptions` together with `IOException` in `OutboundMessageContext#close` , the issue should not happen. However, it would invoke `ContainerResponseWriter#commit` instead of `ContainerResponseWriter#failure`.

As there is an error, it should invoke `ContainerResponseWriter#failure` in my opinion.


jbescos:
[issue5783.zip](https://github.com/user-attachments/files/17645172/issue5783.zip)
Attaching it for version 3.1 so the customer can test. Rename jars accordingly.

senivam (approved):
LGTM

jansupol (approved):

PR #5792, opened 2024-11-07
[2.x] ParamConverterProvider constructors are now protected to allow extensibility

### LINKED ISSUE #5791: Jersey 2: Relax ParamConverters constructor visibility to public
PR https://github.com/eclipse-ee4j/jersey/pull/5349 changed the visibility of default ParamConverter constructors in Jersey 2.40+ to private. For example:
https://github.com/eclipse-ee4j/jersey/pull/5349/files#diff-739aa86c923eab6eead201b6b5da407758e6786b49473e9fd8507e0a172427c2R269

Our project extends default ParamConverters to handle custom encryption and correspondent annotations and 2.40 breaks this ability for us. The only workaround I see would be to copy-paste all the default implementations to our project, which I would love to avoid, or to use reflection to open up the constructors.

Looking at the PR, closing up the constructors doesn't seem to be the objective and this compatibility break is more of an unintended side effect. Constructors of OptionalCustomProvider and OptionalProvider are also left open, as the added fuature didn't touch them.

Could these constructors be opened up back and made public?

Spikhalskiy:
This PR addresses #5791 and opens back up constructors that were ?unintentionally? closed in #5349.
This allows users to decorate or extend the functionality of built-in providers without fully copy-pasting the implementations into user projects.

### DIFF
--- a/core-common/src/main/java/org/glassfish/jersey/internal/inject/ParamConverters.java
+++ b/core-common/src/main/java/org/glassfish/jersey/internal/inject/ParamConverters.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2012, 2023 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2012, 2024 Oracle and/or its affiliates. All rights reserved.
  * Copyright (c) 2018 Payara Foundation and/or its affiliates.
  *
  * This program and the accompanying materials are made available under the
@@ -121,7 +121,7 @@ public class ParamConverters {
     @Singleton
     public static class StringConstructor extends ParamConverterCompliance implements ParamConverterProvider {
 
-        private StringConstructor(boolean canReturnNull) {
+        protected StringConstructor(boolean canReturnNull) {
             super(canReturnNull);
         }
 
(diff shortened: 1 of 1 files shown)


senivam:
could you please make those constructors `protected` (it should be enough for extending) and sign ECA (as described at our [Wiki](https://github.com/eclipse-ee4j/jersey/wiki/Missing-ECA,-Missing-Sign-off-by))  for the PR could be merged.

Spikhalskiy:
@senivam I made them protected and I also added a protected no-op constructor to `OptionalProvider` to have them all consistent.
I already signed the ECA and the account email matches my github and commit email, but the check wasn't turning green. It looks like it did now after a force-push.
Thank you for the review!

core-common/src/main/java/org/glassfish/jersey/internal/inject/ParamConverters.java
```diff
@@ -206,7 +206,7 @@ public T _fromString(final String value) throws Exception {
     @Singleton
     public static class TypeFromStringEnum extends TypeFromString {
 
-        private TypeFromStringEnum(boolean canReturnNull) {
+        public TypeFromStringEnum(boolean canReturnNull) {
```
senivam: this is still public...

Spikhalskiy: fixed

senivam (approved):
LGTM

jansupol (approved):

PR #5794, opened 2024-11-08
Fix possible concurrent issue with SSL & default connector

### LINKED ISSUE #5797: Jersey core client 3.1.9 memory leak
If I'm not missing something this is a memory leak if you target unique URLs.

https://github.com/eclipse-ee4j/jersey/blob/f35691f4e4f1e6440ba2d384de798c4efdb2f932/core-client/src/main/java/org/glassfish/jersey/client/HttpUrlConnectorProvider.java#L301

jansupol:


tests/e2e-tls/src/test/java/org/glassfish/jersey/tests/e2e/tls/connector/ConcurrentHttpsUrlConnectionTest.java
```diff
@@ -0,0 +1,119 @@
...
+ * all parallel requests from other Jersey client instances fail
+ * with SSLHandshakeException: PKIX path building failed.
+ * <p>
+ * Once the first GET request is completed,
+ * all subsequent requests work without error.
+ * <p>
+ * BUG 5749
+ */
+public class ConcurrentHttpsUrlConnectionTest {
+    private static int THREAD_NUMBER = 2;
```
dtbaum: From my perspective, only 2 threads feels a bit too minimal for a concurrency test.

core-client/src/main/java/org/glassfish/jersey/client/internal/HttpUrlConnector.java
```diff
@@ -387,6 +387,10 @@ private ClientResponse _apply(final ClientRequest request) throws IOException {
             sniUri = request.getUri();
         }
 
+        if (!DEFAULT_SSL_SOCKET_FACTORY.isInitialized() && "HTTPS".equalsIgnoreCase(sniUri.getScheme())) {
```
dtbaum, line 390: The check !DEFAULT_SSL_SOCKET_FACTORY.isInitialized() seems to be not thread-save?

jansupol: The variable used in `isInitialized()` is volatile, so it should be thread-safe.

tests/e2e-tls/src/test/java/org/glassfish/jersey/tests/e2e/tls/connector/ConcurrentHttpsUrlConnectionTest.java
jansupol: The premise is that the first request passes and all the consecutive do not until the first is finished. Two seem to comply with the premise and the test failed indeed without the fix. I can certainly put back your 5, if you insist.

senivam (approved):

jbescos (approved):

PR #5800, opened 2024-11-13
NettyConnector - HOST header contains port & 307 works with buffered post

### LINKED ISSUE #5799: jersey-netty-connector: non standard port not added to Host: header when sending requests
[](https://www.rfc-editor.org/rfc/rfc9110.html#name-host-and-authority )
The “Host” header field in a request provides the host and port information from the target URI, enabling the origin server to distinguish among resources while servicing requests for multiple host names.

When I send a request to http://127.0.0.1:8080 I see
GET / HTTP/1.1
Accept-Encoding: gzip
Accept-Encoding: x-gzip
User-Agent: Jersey/2.45 (Netty 4.1.93.Final)
**host: 127.0.0.1**

If the port is non-standard then it should be present in the Host: header that is sent as part of the request.

jansupol:
Host header contains port for non 80 & 443 ports
Status 307 with post and buffered entity does not hang

### DIFF
--- a/connectors/netty-connector/src/main/java/org/glassfish/jersey/netty/connector/JerseyClientHandler.java
+++ b/connectors/netty-connector/src/main/java/org/glassfish/jersey/netty/connector/JerseyClientHandler.java
@@ -17,7 +17,6 @@
 package org.glassfish.jersey.netty.connector;
 
 import java.io.IOException;
-import java.io.InputStream;
 import java.net.URI;
 import java.util.Iterator;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


senivam (approved):

jbescos (approved):

sxsubram (commented):
jerseyRequest.setStreamProvider(new OutboundMessageContext.StreamProvider() {
                    @Override
                    public OutputStream getOutputStream(int contentLength) throws IOException {
                        replaceHeaders(jerseyRequest, nettyRequest.headers()); // WriterInterceptor changes
                        if (!nettyRequest.headers().contains(HttpHeaderNames.HOST)) {
                            nettyRequest.headers().add(HttpHeaderNames.HOST, jerseyRequest.getUri().getHost());
                        }
                        headersSet.countDown();

                        return entityWriter.getOutputStream();
                    }
                });

port needs to be added here as well.

jansupol:
@sxsubram nice catch! Fixed by #5801

PR #5809, opened 2024-11-26
Fixed memory leak in Micrometer code when exceptions are returned from the method

### LINKED ISSUE #5807: Memory leak in Micrometer code when exceptions are returned from the method
We've identified a memory leak in `ThreadLocal` when the application endpoint returns with an exception. E.g.:
```java
@GET
@Path("/404")
public String notFound() {
    throw new NotFoundException("Not found");
}
```

It happens because in case of exceptions, the observation is started twice in:
https://github.com/eclipse-ee4j/jersey/blob/619ebefb168eead0012b10ee7935f07862ca53d8/ext/micrometer/src/main/java/org/glassfish/jersey/micrometer/server/ObservationRequestEventListener.java#L67-L72
And:
https://github.com/eclipse-ee4j/jersey/blob/619ebefb168eead0012b10ee7935f07862ca53d8/ext/micrometer/src/main/java/org/glassfish/jersey/micrometer/server/ObservationRequestEventListener.java#L73-L75

This is an issue because Micrometer's `SimpleObservationRegistry` keeps the observation scope inside a `ThreadLocal` variable in [SimpleObservationRegistry.java#L30](https://github.com/micrometer-metrics/micrometer/blob/311efdb2f2079103e5a0e4b0b87b12d91e405af8/micrometer-observation/src/main/java/io/micrometer/observation/SimpleObservationRegistry.java#L30):
```java
private static final ThreadLocal<Observation.Scope> localObservationScope = new ThreadLocal<>();
```

The instance of `Scope` class contains a pointer to a previous scope [SimpleObservation.java#L285](https://github.com/micrometer-metrics/micrometer/blob/311efdb2f2079103e5a0e4b0b87b12d91e405af8/micrometer-observation/src/main/java/io/micrometer/observation/SimpleObservation.java#L285):
```java
static class SimpleScope implements Scope {
    @Nullable
    final Scope previousObservationScope;
```

And these 2 things are updated in [SimpleObservation.java#L290-L291](https://github.com/micrometer-metrics/micrometer/blob/311efdb2f2079103e5a0e4b0b87b12d91e405af8/micrometer-observation/src/main/java/io/micrometer/observation/SimpleObservation.java#L290-L291)
```java
this.previousObservationScope = registry.getCurrentObservationScope();
this.registry.setCurrentObservationScope(this);
```

Now, each request that ends with an exception causes this linked list of observations to grow, because the code in: https://github.com/eclipse-ee4j/jersey/blob/619ebefb168eead0012b10ee7935f07862ca53d8/ext/micrometer/src/main/java/org/glassfish/jersey/micrometer/server/ObservationRequestEventListener.java#L83-L91 is closing only one observation.

Here is an example test that reproduces the issue: https://github.com/eclipse-ee4j/jersey/compare/2.x...adpaste:jersey:memory_leak

Memory dump:
image

adpaste:
Proposed fix for https://github.com/eclipse-ee4j/jersey/issues/5807

### DIFF
--- a/ext/micrometer/src/main/java/org/glassfish/jersey/micrometer/server/ObservationRequestEventListener.java
+++ b/ext/micrometer/src/main/java/org/glassfish/jersey/micrometer/server/ObservationRequestEventListener.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2024 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
@@ -65,7 +65,7 @@ public class ObservationRequestEventListener implements RequestEventListener {
 
         switch (event.getType()) {
             case ON_EXCEPTION:
-                if (!isNotFoundException(event)) {
+                if (!isClientError(event) || observations.get(containerRequest) != null) {
                     break;
                 }
                 startObservation(event);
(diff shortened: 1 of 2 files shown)


senivam (approved):

jansupol (approved):
LGTM

PR #5811, opened 2024-11-28
Fix exception message on failure to create JakartaXmlBindAnnotationIntrospector

Spikhalskiy:
Commit https://github.com/eclipse-ee4j/jersey/commit/62ba351c6a1fde072b5634318551704eb277371d addressed the wrong class being used in Jersey 3.0.x but didn't adjust an error message accordingly. This PR fixes this.

### DIFF
--- a/media/json-jackson/src/main/java/org/glassfish/jersey/jackson/internal/jackson/jaxrs/json/JsonMapperConfigurator.java
+++ b/media/json-jackson/src/main/java/org/glassfish/jersey/jackson/internal/jackson/jaxrs/json/JsonMapperConfigurator.java
@@ -129,7 +129,7 @@ public class JsonMapperConfigurator
                     }
                     return _jaxbIntrospectorClass.newInstance();
                 } catch (Exception e) {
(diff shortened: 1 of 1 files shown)


senivam:
just FYI, the line you are changing (along with the whole Jackson in the ``**/jackson/jaxrs/**`` package) is repackaged from [here](https://github.com/FasterXML/jackson-jaxrs-providers/blob/c6cc58afb0aa68394dcfaa4f02cb2bd009fc26db/json/src/main/java/com/fasterxml/jackson/jaxrs/json/JsonMapperConfigurator.java#L130)

Spikhalskiy:
@senivam I’m not sure what the implication is. In that code version the class name in the message is aligned with the class used. After https://github.com/eclipse-ee4j/jersey/commit/62ba351c6a1fde072b5634318551704eb277371d it doesn’t and needs to be adjusted.

senivam:
@Spikhalskiy sure thing. That change was also made to the repackaged code. I just wanted to let you know that its origin is not ours. For us, it's important to keep track of those changes not to be re-written on the next version update. For you, it's just informative :)

senivam (approved):

jansupol (approved):

PR #5815, opened 2024-12-03
Use string as hashcode in lock map in HttpUrlConnectorProvider

### LINKED ISSUE #5814: Jersey update from 3.1.8 to 3.1.9 performance regression when using the JVM `http.proxyHost` mechanism
This PR introduced the regression: https://github.com/eclipse-ee4j/jersey/pull/5749 

The referenced fix introduces a map from `java.util.URL` to a `Lock`. That has the side effect of calling `getInetAddress(url.host)`, which significantly impacts the performance if you are running the JVM with a proxy with `-Dhttp.proxyHost` and `-Dhttp.proxyPort`.

Before the call to `hashCode` the JVM handled the proxy connection transparently.

ripdajacker:
See issue: https://github.com/eclipse-ee4j/jersey/issues/5814

This patch changes the key type of the `locks` map in `org.glassfish.jersey.client.HttpUrlConnectorProvider` to avoid calling `getInetAddress` on the URL's  every time a URL connection is opened.

This is an issue when using the built-in JVM http proxy mechanism.

### DIFF
--- a/core-client/src/main/java/org/glassfish/jersey/client/HttpUrlConnectorProvider.java
+++ b/core-client/src/main/java/org/glassfish/jersey/client/HttpUrlConnectorProvider.java
@@ -298,7 +298,7 @@ public class HttpUrlConnectorProvider implements ConnectorProvider {
 
     private static class DefaultConnectionFactory implements ConnectionFactory {
 
-        private final ConcurrentHashMap<URL, Lock> locks = new ConcurrentHashMap<>();
+        private final ConcurrentHashMap<String, Lock> locks = new ConcurrentHashMap<>();
 
         @Override
         public HttpURLConnection getConnection(final URL url) throws IOException {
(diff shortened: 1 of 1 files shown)


senivam:
please try to sign ECA ([Wiki](https://github.com/eclipse-ee4j/jersey/wiki/Missing-ECA,-Missing-Sign-off-by)) to merge your PR. Otherwise, we cannot accept it. And thank you for the fix!

ripdajacker:
I've signed the  ECA :-)

jansupol:
@ripdajacker Thank you for this.
However, #5813 drops the lock at all and would be even better performance-wise.

ripdajacker:
> @ripdajacker Thank you for this. However, #5813 drops the lock at all and would be even better performance-wise.

That seems to be for the 2.1 branch.

My fix does not fix the concurrency implications of using locks but it removes the very subtle potential DNS lookup when attempting to require a lock.

I am certain there are better ways to improve performance, but cannot see how the linked PR fixes the issue at hand.

jansupol:
@ripdajacker Sorry a wrong link, midnight work... #5794 should be the fix.

ripdajacker:
That will probably do, will it be merged to 3.1 too?

senivam:
@ripdajacker yes, before every release of 3.0.x and 3.1.x changes from 2.x and 3.0 branches are propagated respectively.

ripdajacker:
@senivam Cool, good to know :) Is there an ETA? 

I'll close the PR then.

PR #5816, opened 2024-12-04
Allow to configure Jackson's JaxRSFeature on Jersey DefaultJacksonJaxbJsonProvider

jansupol:


senivam (approved):

media/json-jackson/src/main/java/org/glassfish/jersey/jackson/internal/JaxrsFeatureBag.java
```diff
@@ -0,0 +1,58 @@
...
+    private static class JaxRSFeatureState {
+        /* package */ final JaxRSFeature feature;
+        /* package */ final boolean state;
+        public JaxRSFeatureState(JaxRSFeature feature, boolean state) {
+            this.feature = feature;
+            this.state = state;
+        }
+    }
+
+    private Optional<List<JaxRSFeatureState>> jaxRSFeature = Optional.empty();
```
jbescos, line 41: Does it makes sense to remove Optional and just have the list?:
`private final List<JaxRSFeatureState> jaxRSFeatures = new ArrayList<>();`

If empty, `hasJaxrsFeature` returns false

I found this is more understandable, although this is only a personal opinion.

jbescos (approved):

Issue #5834, opened 2025-01-14
OSGi / Spifly Provider for jakarta.ws.rs.ext.RuntimeDelegate cannot be found 


col-panic:
I have an OSGI environment with  jersey-common 3.1.9, jakarta.ws.rs 3.1.0 and spifly 1.3.7

On startup console says
```
Caused by: java.lang.RuntimeException: java.lang.ClassNotFoundException: Provider for jakarta.ws.rs.ext.RuntimeDelegate cannot be found
	at jakarta.ws.rs.ext.RuntimeDelegate.findDelegate(RuntimeDelegate.java:118)
	at jakarta.ws.rs.ext.RuntimeDelegate.getInstance(RuntimeDelegate.java:88)
	at jakarta.ws.rs.core.Response$ResponseBuilder.newInstance(Response.java:776)
	at jakarta.ws.rs.core.Response.status(Response.java:522)
	at jakarta.ws.rs.core.Response.status(Response.java:533)
	at jakarta.ws.rs.core.Response.ok(Response.java:566)
	at ch.elexis.core.services.es.NoRemoteEventService.<init>(NoRemoteEventService.java:8)
	at ch.elexis.core.services.es.ElexisServerService.<init>(ElexisServerService.java:72)
	at java.base/jdk.internal.reflect.DirectConstructorHandleAccessor.newInstance(DirectConstructorHandleAccessor.java:62)
	... 146 more
Caused by: java.lang.ClassNotFoundException: Provider for jakarta.ws.rs.ext.RuntimeDelegate cannot be found
	at jakarta.ws.rs.ext.FactoryFinder.find(FactoryFinder.java:163)
	at jakarta.ws.rs.ext.RuntimeDelegate.findDelegate(RuntimeDelegate.java:103)
	... 154 more
```

`jakarta.ws.rs.ext.RuntimeDelegate` is available within `jersey-common` and the `/META-INF/services/jakarta.ws.rs.ext.RuntimeDelegate` file exists.

It is, however, NOT being picked up by spifly. This is due to `/META-INF/MANIFEST.MF` not containing the required `Require-Capability` and `Provide-Capability` entries. See https://aries.apache.org/documentation/modules/spi-fly.html#specconf for details.

For example:

`logback-classic` correctly writes this info to `MANIFEST-MF` in https://github.com/qos-ch/logback/blob/5610c96b4d705a4fe6ded9c42d4f47cb92bbbd95/logback-classic/pom.xml#L343

[...]

PR #5851, opened 2025-02-03
#5384 add missing osgi capability info for spifly injection

col-panic:
As described in https://github.com/eclipse-ee4j/jersey/issues/5834

### DIFF
--- a/core-client/pom.xml
+++ b/core-client/pom.xml
@@ -95,6 +95,16 @@
                             *
                         </Import-Package>
                         <unpackBundle>true</unpackBundle>
+                        <!-- Needed to integrate ServiceLoader mechanism with OSGi -->
+                        <Require-Capability><![CDATA[
+                         osgi.extender;filter:="(&(osgi.extender=osgi.serviceloader.processor)(version>=1.0.0)(!(version>=2.0.0)))";
+                         resolution:=optional,osgi.serviceloader;filter:="(osgi.serviceloader=jakarta.ws.rs.client.ClientBuilder)";osgi.serviceloader="jakarta.ws.rs.ClientBuilder";cardinality:=multiple;
+                         resolution:=optional,osgi.extender;filter:="(&(osgi.extender=osgi.serviceloader.registrar)(version>=1.0.0)(!(version>=2.0.0)))";resolution:=optional,
+                         osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.8))"
+                         ]]></Require-Capability>
+                        <Provide-Capability><![CDATA[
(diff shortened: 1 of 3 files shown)


jansupol:
@col-panic why is this needed? 
[Section 13.3](https://docs.osgi.org/specification/osgi.enterprise/7.0.0/service.loader.html#d0e61102). Says:
> Consumers are classes that are not OSGi aware and directly use the Service Loader API

But that is not Jersey. Jersey is OSGi-aware, see [OsgiRegistry](https://github.com/eclipse-ee4j/jersey/blob/3.1/core-common/src/main/java/org/glassfish/jersey/internal/OsgiRegistry.java) class that has OsgiServiceFinder to lookup for the services in an OSGi environment.

Jersey is using [ServiceFinder](https://github.com/eclipse-ee4j/jersey/blob/3.1/core-common/src/main/java/org/glassfish/jersey/internal/ServiceFinder.java) class to lookup instead of using directly ServiceLocator. That has an OSGi-aware ServiceIteratorProvider at hand.

col-panic:
It is tightly interconnted with https://github.com/jakartaee/rest/pull/1308 - without this patch I can't seem to connect jakarta rest requiring the services provided by jersey! Only with both patches this and the one in jakartaee/rest repo I managed to get the entire functionality up and running in my Osgi environment! Despite jersey having the functionality jakarta rest seems not to?!

Do you see another solution?

jansupol:
Is it not enough to make the API OSGi-aware?

col-panic:
> Is it not enough to make the API OSGi-aware?

This seems to be the preferred approach to do just that. Take a look at the first post in https://github.com/eclipse-ee4j/jersey/issues/5834 - I referenced other projects realizing this compatibility just by extending the MANIFEST with the adaptations proposed in this patch.

jansupol:
@col-panic We have [OSGi tests](https://github.com/eclipse-ee4j/jersey/tree/3.1/tests/osgi/functional/src/test/java/org/glassfish/jersey/osgi/test/basic) in Jersey using Felix. Some magic is used there if you care to look.

The PR you created adds some code that is not covered by tests, and we do not understand it. If later someone comes saying there needs to be done something differently, we have no possibility to decide if yes or no. So we need to understand what is it you do.

huthomas:
The magic in the OSGI tests seems to be setting the following system properties.
```
		System.setProperty("org.osgi.framework.system.packages.extra", "jakarta.annotation");
		System.setProperty("jakarta.ws.rs.ext.RuntimeDelegate", "org.glassfish.jersey.internal.RuntimeDelegateImpl");
		System.setProperty("jakarta.ws.rs.client.ClientBuilder", "org.glassfish.jersey.client.JerseyClientBuilder");
```

PR #5859, opened 2025-02-06
Memory leak when using Jersey with HK2 #5796

jbescos:
https://github.com/eclipse-ee4j/jersey/issues/5796

To make sure ClientRuntime is not GCed meanwhile the InputStream or OutputStream is read or written, I added a WeakHashMap in JerseyClient, so ClientRuntime will exist meanwhile InputStream exists.

PR #5882, opened 2025-03-14
Release instances blocking GC when JAXB has PerThread injections

### LINKED ISSUE #5796: Memory leak when using Jersey with HK2
When using Jersey and Jackson to deserialize XML, any change to the configuration of the WebTarget causes a memory leak since hk2 3.0.4 (and up to 3.1.1).

I suspect this is caused by this change in HK2: https://github.com/eclipse-ee4j/glassfish-hk2/commit/9e5b5f576f7c4202a30708a68b79c4a9b2d2ef0f (FISH-6432 Applications Take Longer To Deploy on JDK 17) since the same occurs with the patched Payara version 3.0.1.payara-p2 (https://github.com/payara/patched-src-hk2/compare/3.0.1.payara-p1...3.0.1.payara-p2).

Not sure if this is a regression in HK2 or something that Jersey is doing incorrectly.

I've tested with Jersey version 3.0.0 to 3.1.9, and the behavior is the same.

See the attached example project.
[jersey-memory-leak.zip](https://github.com/user-attachments/files/17705432/jersey-memory-leak.zip)

Also posted on HK2: https://github.com/eclipse-ee4j/glassfish-hk2/issues/1126

jansupol:


jansupol:
Fixes #5796.
Alternative to #5859.
Another alternative is to change the binding for JAX-B classes from `PerThread` to `PerLookup`. Possibly, these two can be done in conjunction.

senivam (approved):

jbescos (approved):

Issue #5871, opened 2025-02-21
Update osgi-resource-locator to 3.0.0


akurtakov:
If one looks at https://github.com/eclipse-ee4j/jersey/blob/382f69ece53a82d58077142e0f71ff705a4f463f/pom.xml#L1599 it uses version 1.0.3 which is from 2019. It would be nice to update to version 3.0.0 https://mvnrepository.com/artifact/org.glassfish.hk2/osgi-resource-locator/3.0.0 .

jansupol:
Relates to #5860.

There is an interest in this osgi-resource-locator. What is the purpose? Does the older one not satisfy your needs?

jansupol:
OSGi locator 1.0.3 is replaced by OSGi locator 1.0.4 for EE10 (cannot be used for EE9 as it is built by JDK 17, but EE9 is JDK 11 compatible). Version 3.0.0 is for EE11.

PR #5860, opened 2025-02-10
Updated osgi-resource-locator version

abhilashmandaliya:

PR #5885, opened 2025-03-23
Upgrade HK2-extra osgi-resource-locator

OndroMih:
3.0.0 version is a major version only to fix versioning issues, otherwise it contains only minor changes.

See release notes: https://github.com/eclipse-ee4j/glassfish-hk2-extra/releases/tag/3.0.0-RELEASE

Releasing a new version of Jersey with this change would allow to integrate the new version of HK2-extra into GlassFish. Without this, GlassFish gets OSGi version conflicts, see: https://github.com/eclipse-ee4j/glassfish/pull/25366

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1524,7 +1524,7 @@
             <dependency>
                 <groupId>org.glassfish.hk2</groupId>
                 <artifactId>osgi-resource-locator</artifactId>
-                <version>1.0.3</version>
+                <version>3.0.0</version>
             </dependency>
             <dependency>
                 <groupId>org.glassfish.main.hk2</groupId>


jansupol:
@OndroMih This is for GF 8 only, correct?

Relates to #5871, #5860.

OndroMih:
This is for GlassFish 7, which now uses Jersey 3.1.10. As I wrote in the description, we decided to release 3.0.0 of HK2-extra to align to avoid conflicts with older Oracle 2.x versions. 

Now that I think about it, it’s probably better to release the same thing as 1.0.4 too, which could be used in GlassFish 7 without any change in Jersey. Then 3.0.0 would be only for GlassFish 8. I’ll do that and close this PR. Then 3.0.0 would target Jersey 4.0 only.

PR #5904, opened 2025-04-24
Fix for the InputStream caching cases for Servlets

### LINKED ISSUE #5739: Jersey not compatible with HttpServletRequests not always returning the same input stream from version 2.44
After [this pr](https://github.com/eclipse-ee4j/jersey/pull/5669) jersey is no longer handling correctly any override of `HttpServletRequest` which does not return always the same input stream when invoking `getInputStream`.

## Context about different implementation of ServletRequest#getInputStream
The servlet api does not seem to specify wether or not it is required for an implementation to always return the same input stream across multiple invocations and it therefore seems valid in some cases to override this method in an `HttpServletRequestWrapper` returning a new copy of the input stream for any invocation (to be able to consume the input stream multiple times). This is showcased for example in this [stack overflow answer](https://stackoverflow.com/a/17129256) about how a servlet input stream could be consumed multiple times.

## Current issue
With the aforementioned change the `InputStreamWrapper` will invoke the underlying `ServletRequest#getInputStream` whenever any operation is performed to the request input stream. This can cause the input stream to be read forever as invocations to `InputStreamWrapper#available` will create a new input stream every time that this has not yet been consumed.

## How to reproduce this issue
When overriding the http servlet request in a servlet filter with a request such as the one showed in the previous stack overflow answer no jersey invocation to an endpoint consuming json seems to work.

## Possible solutions
This use case seems quiet valid also considering that `HttpServletRequest` does not expose anywhere in its api a way to override the input stream and therefore allow for more easy input replayability. I therefore wondered if this could be fixed/supported by having `InputStreamWrapper` caching the result of `InputStreamWrapper#getWrapped` after it is invoked the first time?

senivam:


jansupol (approved):

jbescos (approved):

asharayev:
Is it possible to migrate the part of fix to 3.1.* versions branch?

Specifically the following changes, caching wrappedStream to avoid multiple calls to request.getInputStream from InputStreamWrapper:
````
        try {
            requestContext.setEntityStream(new InputStreamWrapper() {

                private ServletInputStream wrappedStream;
                @Override
                protected InputStream getWrapped() {
                    if (wrappedStream == null) {
                        try {
                            wrappedStream = servletRequest.getInputStream();
                        } catch (IOException e) {
                            throw new UncheckedIOException(e);
                        }
                    }
                    return wrappedStream;
                }
            });
        } catch (UncheckedIOException e) {
````
?

Let me clarify:
We have a servlet request wrapper that allows multiple consumptions of source input stream through caching of source request's stream bytes. However, current 3.1.* branch calls InputStreamWrapper.getWrapped() consistently while iterating bytes of the stream, not calling it once, and then iterating over the received stream object.

We get the following errors starting from 3.1.8 jersey version:
````
at java.util.Arrays.copyOf(java.base@21.0.2/Arrays.java:3539)
at java.io.ByteArrayOutputStream.toByteArray(java.base@21.0.2/ByteArrayOutputStream.java:187)
at security.util.MultiReadHttpServletRequest$CachedServletInputStream.<init>(MultiReadHttpServletRequest.java:53)
at security.util.MultiReadHttpServletRequest.getInputStream(MultiReadHttpServletRequest.java:34)
at org.glassfish.jersey.servlet.WebComponent$5.getWrapped(WebComponent.java:431)
at org.glassfish.jersey.innate.io.InputStreamWrapper.getWrappedIOE(InputStreamWrapper.java:40)
at org.glassfish.jersey.innate.io.InputStreamWrapper.available(InputStreamWrapper.java:81)
at sun.nio.cs.StreamDecoder.inReady(java.base@21.0.2/StreamDecoder.java:409)
at sun.nio.cs.StreamDecoder.implRead(java.base@21.0.2/StreamDecoder.java:370)
at sun.nio.cs.StreamDecoder.lockedRead(java.base@21.0.2/StreamDecoder.java:215)
at sun.nio.cs.StreamDecoder.read(java.base@21.0.2/StreamDecoder.java:169)
at java.io.InputStreamReader.read(java.base@21.0.2/InputStreamReader.java:188)
at java.io.Reader.read(java.base@21.0.2/Reader.java:265)
````
The cause is that org.glassfish.jersey.servlet.WebComponent$5.getWrapped(WebComponent.java:431) calls actually getInputStream for every single byte, thus doesn't iterate over a single stream, rather retrieves a stream every time from request.

If we want to make a stream wrapper that allow multiple reads of the source stream, then we need to return a new stream every time getInputStream() is called, otherwise we don't have any hint on whether this is just an iteration or a start of a new iteration of a stream.

Thanks in advance.

jansupol:
@asharayev Sure, it already is in [3.1](https://github.com/eclipse-ee4j/jersey/blob/3.1/containers/jersey-servlet-core/src/main/java/org/glassfish/jersey/servlet/WebComponent.java#L435)

asharayev:
> @asharayev Sure, it already is in [3.1](https://github.com/eclipse-ee4j/jersey/blob/3.1/containers/jersey-servlet-core/src/main/java/org/glassfish/jersey/servlet/WebComponent.java#L435)

Indeed! Thanks for the info!

PR #5911, opened 2025-05-13
Removed obsolete-for-removal UTF8 constant

mkarg:
# Aim

The target of this PR is to finally remove the `UTF8` custom constants which are currently deprecated-for-removal.

# Justification

There is no need to keep custom constants without a good reason.

4.0 is a major release, so removing code which is deprecated-for-removal is justified and does not break SemVer rules.

`StandardCharsets.UTF_8` exists since Java 7. Minimum supported JRE in 4.0 is 17+.


mkarg:
@jansupol Kindly asking for approval. I have introduced deprecated-for-removal in 3.x, so we should take the chance to finally remove needless code now in 4.x. Thanks. 🙂

jansupol (approved):

senivam (approved):

PR #5916, opened 2025-05-17
Replace UTF_8 as string in UriComponent

jansupol:


jansupol:
Closing as the method is since JDK 10

PR #5918, opened 2025-05-19
Fix wrong handling of duplicate cookies

jjqq2013:
----
Handling duplicate cookies on server side has no clear specification.

But if there is any hint or common-sense about it, it is the "longest **path**" wins, instead of currently implemented logic of "longest **value**" wins.

Section [5.4](https://www.rfc-editor.org/rfc/rfc6265#section-5.4), subsection 2 has this to say:

```
The user agent SHOULD sort the cookie-list in the following order:

Cookies with longer paths are listed before cookies with shorter paths.
NOTE: Not all user agents sort the cookie-list in this order, but this order reflects common practice when this document was written, and, historically, there have been servers that (erroneously) depended on this order.
```
though it is not for server side, it does mention how the path of cookie matters.

----
For duplicate cookies that has even same path, let us use the last value, as have been doing in earlier version (such as 2.22.1).

For example:

- Visit https://site1.domain/, it sets cookieX=value1, domain=.domain, path=/

- Visit https://site2.subdomain.domain/, it sets cookieX=value2, domain=.subdomain.domain, path=/

- Visit https://site2.subdomain.domain/ again, the Browser will just send cookies with the creation order (because the path are same, browsers have no idea which should be first), i.e., cookieX=value1;cookieX=value2

- Previous implementation of CookieParser will result in the cookieX=value1 being used, then the server side is not satisfied with the cookie value. (Actually in earlier version(2.22.1) it uses cookieX=value2.)

Let us use the last cookie value, so both web sites will work,
- Visit https://site2.subdomain.domain/ will use value2
- Visit https://site1.domain/ will use value1 because the domain of value2 can not be applied to this parent domain.

### DIFF
--- a/core-common/src/main/java/org/glassfish/jersey/message/internal/CookiesParser.java
+++ b/core-common/src/main/java/org/glassfish/jersey/message/internal/CookiesParser.java
@@ -94,7 +94,9 @@ public class CookiesParser {
     private static void checkSimilarCookieName(Map<String, Cookie> cookies, MutableCookie cookie) {
         if (cookie != null) {
             if (cookies.containsKey(cookie.name)){
-                if (cookie.path.length() >= cookies.get(cookie.name).getPath().length()){
+                int pathLen = cookie.path == null ? 0 : cookie.path.length();
+                int oldPathLen = cookies.get(cookie.name).getPath() == null ? 0 : cookies.get(cookie.name).getPath().length();
+                if (pathLen >= oldPathLen) {
                     cookies.put(cookie.name, cookie.getImmutableCookie());
                 }
             } else {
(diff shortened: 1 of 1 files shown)


jansupol:
Please see our [wiki](https://github.com/eclipse-ee4j/jersey/wiki/Missing-ECA,-Missing-Sign-off-by).

jjqq2013:
Thank you for your reply. This PR is just a suggestion, prompt, hope someone else can work for it. Actually I just wanted to create an issue but it is easier with a PR.

senivam:
Having your PR as a PoC, I've created the #5919 / is it what you expect to be changed?

senivam:
> Having your PR as a PoC, I've created the #5919 / is it what you expect to be changed?

@jjqq2013, having no response for a week, I suppose there are no objections to merge the #5919 and close this one.

senivam:
closing as #5919 was merged

PR #5806, opened 2024-11-20
Improve EntityInputStream#isEmpty handling for the Servlet case.

senivam:

PR #5924, opened 2025-05-29
improve EntityInputStream#isEmpty handling for the Servlet case. 

senivam:
porting #5806 to the 2.x branch.


senivam:
closing as NO-OP

PR #5928, opened 2025-06-04
merge of the current 3.0 into the 3.1

senivam:


core-client/pom.xml
```diff
@@ -152,6 +158,19 @@
     </dependencies>
 
     <profiles>
+        <profile>
```
jansupol: follow up: why for JDK 11 only, not 12, not 17?

tests/integration/security-digest/pom.xml
```diff
@@ -69,7 +69,7 @@
                     <loginServices>
                         <loginService implementation="org.eclipse.jetty.security.HashLoginService">
                             <name>my-realm</name>
-                            <config implementation="org.eclipse.jetty.ee10.maven.plugin.MavenResource">
+                            <config implementation="org.eclipse.jetty.maven.MavenResource">
```
jansupol, line 72: Is this correct?

senivam: yes

senivam:
I am converting this to DRAFT because it will be updated with more changes from  3.0. The profile in the core-client seems to be redundant in 3.1, and will be removed from the merge.

core-client/pom.xml
senivam: cleared

tests/integration/security-digest/pom.xml
senivam: link to the [user guide](https://jetty.org/docs/jetty/12/programming-guide/maven-jetty/jetty-maven-plugin.html#configuring-security-settings)

senivam:
Now, merge is updated with the latest 3.0, switching back to the ready-to-review state.

jansupol (approved):

PR #5941, opened 2025-06-24
Avoid NoSuchMethodError during normal operation with Servlet 6+

Spikhalskiy:
Currently, a normally operating application with Jersey produces a lot of `NoSuchMethodError`, which is shown on profilers and flight recordings, and also can be expensive if  `-XX:-OmitStackTraceInFastThrow` is specified.
This PR checks the existence of the method once during static initialization and avoids creating Errors.


senivam (approved):

jansupol:
@Spikhalskiy The problem is that you are using Servlet 6 with Jersey 3.0. But Jersey 3.0 is for Servlet 5. For Servlet 6, Jersey 3.1 should be used; Jersey 3.1.x does not use non-existent servlet methods with a reason phrase.

That being said, this code is better than catching an exception.

jansupol (approved):

Spikhalskiy:
@jansupol Thank you for your response! Yeah, you are right. It’s just Jersey 3.0 still working just fine for us with Servlet 6, so we didn’t have the urge to upgrade.

PR #5943, opened 2025-06-25
Eliminate unnecessary flushes

jansupol:


core-common/src/main/java/org/glassfish/jersey/message/internal/OutboundMessageContext.java
```diff
@@ -562,8 +562,12 @@ public void close() {
...
             try {
                 final OutputStream es = getEntityStream();
-                if (!FlushedCloseable.class.isInstance(es)) {
-                    es.flush();
+                if (!FlushedCloseable.flushOnClose(es)) {
+                    if (CommittingOutputStream.class.isInstance(es)) {
+                        ((CommittingOutputStream) es).flushOnClose();
+                    } else {
+                        es.flush();
+                    }
```
senivam: can this be:
```
                if (CommittingOutputStream.class.isInstance(es)) {
                    ((CommittingOutputStream) es).flushOnClose();
                } else if (!FlushedCloseable.class.isInstance(es)) {
                    es.flush();
                }
``` 
?

jansupol:
```
if (CommittingOutputStream.class.isInstance(es)) {
    ((CommittingOutputStream) es).flushOnClose();
} else if (!FlushedCloseable.flushOnClose(es)) {
    es.flush();
}
```
Yes, it can.

jansupol: Probably better.

senivam (approved):

PR #5957, opened 2025-07-23
Merge jersey-container-servlet-core and jersey-container-servlet modules

jansupol:


tests/integration/servlet-tests/src/test/java/org/glassfish/jersey/tests/integration/servlettests/FilterContextPathITCase.java
```diff
@@ -50,12 +50,12 @@ public void testContextPathResource() {
 
     @Test
     public void testNoContextPathResource() {
-        assertEquals("filter/resource", target("filter/resource").request().get(String.class));
+        assertEquals("filter/resource", target("filter/filter/resource").request().get(String.class));
```
jbescos, line 53: Why is it needed to modify the context path?

jansupol: It took me a long time to find out. Servlet 2.5 API does not support getters for FILTER registrations, only for servlet registrations. So whenever the filter used to register URL mapping with the jersey-container-servlet-core, the Jersey Server did not know about the URL mapping from the filter (the jersey-container-servlet, which supports API 3.1, which has a getter for filter URL mapping, had to be used). There was a trick in the tests, the filter mapping was identical to the first segment in the `@Path` mapping. The first was used by the servlet container for the match to pass the request to Jersey, and the second is used by Jersey. When the Servlet API 3.1 is used, Jersey knows about the filter mapping, and both filter mapping and the segment in the Path is used. So the segment needs to be (correctly) duplicated.

senivam (approved):

jbescos (approved):

PR #5972, opened 2025-08-15
merge of the current 4.0 into the 4.0.JPMS

senivam:


tests/integration/pom.xml
```diff
@@ -96,6 +97,7 @@
         <module>microprofile</module>
         <module>property-check</module>
         <module>reactive-streams</module>
+        <module>resteasy-client</module>
```
jansupol, line 101: where is  `<module>jersey-5796</module>`?

tests/e2e-server/pom.xml
```diff
@@ -223,6 +230,34 @@
     </dependencies>
 
     <profiles>
+        <profile>
```
jansupol: This can be left out

senivam:
as soon as those notes are valid for the origin 4.0 branch, I've created #5982 for 4.0. After it's validated and merged, I'll pick it up here

containers/jersey-servlet/src/main/java/module-info.java
```diff
@@ -16,14 +16,21 @@
...
-    requires jakarta.servlet;
+    requires static jakarta.servlet;
 
     requires org.glassfish.jersey.core.common;
     requires org.glassfish.jersey.core.server;
-    requires org.glassfish.jersey.container.servlet.core;
 
     exports org.glassfish.jersey.servlet.async;
     exports org.glassfish.jersey.servlet.init;
+    exports org.glassfish.jersey.servlet.internal;
```
jansupol: verify this is required

connectors/jetty-connector/pom.xml
```diff
@@ -124,6 +124,10 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
```
jansupol: This looks like a leftover from IDE, but it does not make sense here.

tests/e2e-server/pom.xml
senivam: removed via #5982

tests/integration/pom.xml
senivam: added via #5982

connectors/jetty-connector/pom.xml
senivam: fixed

containers/jersey-servlet/src/main/java/module-info.java
senivam: if you allow replacement of the 
```
if (ResponseWriter.class.getName().equals(stackTraceElement.getClassName())) {
```
statement by the 
```
if ("ResponseWriter".equals(stackTraceElement.getClassName())) {
```
 then this export is not required (now commented out from the ``module-info.java``

senivam: this replacement occurs in the file ``tests/integration/async-jersey-filter/src/main/java/org/glassfish/jersey/tests/integration/jersey2730/TestExceptionResource.java``

jansupol (approved):

Issue #5939, opened 2025-06-24
Memory leak when reusing response object


Spikhalskiy:
Our project contains the following code:

```
@Path("/foo")
public class Restlet {

    private static final Response RESPONSE_204 = Response.noContent().build();

    @GET
    @Path("/bar")
    public Response fooBar() {
        return RESPONSE_204;
    }
}
```

We use G1 collector. After making many requests (high load environment), the application starts to create a lot of humongous objects, which in turn causes STW GCs that can successfully clean the humongous objects up, but at a cost of STW.

Our G1 region size is 16Mb, so humongous objects need to be 8Mb+.

Taking the flight recording of the "Allocation outside TLAB" events to investigate the source of humongous objects, it was discovered that all the humongous objects were created with the following stacktrace:

```
Stack Trace	Count	Percentage
Object[] java.util.Arrays.copyOf(Object[], int)	1	100 %
ContainerResponse org.glassfish.jersey.server.model.ResourceMethodInvoker.lambda$invoke$2(RequestProcessingContext, ContainerResponse)	1	100 %
Object org.glassfish.jersey.server.model.ResourceMethodInvoker$$Lambda+0x00007f3b89c647c0.1160227937.apply(Object)	1	100 %
Stage$Continuation org.glassfish.jersey.process.internal.Stages$LinkedStage.apply(Object)	1	100 %
Object org.glassfish.jersey.process.internal.Stages.process(Object, Stage)	1	100 %
ContainerResponse org.glassfish.jersey.server.ServerRuntime$Responder.processResponse(ContainerResponse)	1	100 %
void org.glassfish.jersey.server.ServerRuntime$Responder.process(ContainerResponse)	1	100 %
void org.glassfish.jersey.server.ServerRuntime$1.run()	1	100 %
Void org.glassfish.jersey.internal.Errors$1.call()	1	100 %
Object org.glassfish.jersey.internal.Errors$1.call()	1	100 %
Object org.glassfish.jersey.internal.Errors.process(Callable, boolean)	1	100 %
Object org.glassfish.jersey.internal.Errors.process(Producer, boolean)	1	100 %
void org.glassfish.jersey.internal.Errors.process(Runnable)	1	100 %
void org.glassfish.jersey.process.internal.RequestScope.runInScope(RequestContext, Runnable)	1	100 %
void org.glassfish.jersey.server.ServerRuntime.process(ContainerRequest)	1	100 %
[...]

PR #5978, opened 2025-08-21
Prevent leaking of annotation in a reused response.

jansupol:
#5939

### DIFF
--- a/core-server/src/main/java/org/glassfish/jersey/server/model/ResourceMethodInvoker.java
+++ b/core-server/src/main/java/org/glassfish/jersey/server/model/ResourceMethodInvoker.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011, 2022 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2011, 2025 Oracle and/or its affiliates. All rights reserved.
  *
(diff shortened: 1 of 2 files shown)


jbescos (approved):

core-server/src/main/java/org/glassfish/jersey/server/model/ResourceMethodInvoker.java
```diff
@@ -487,6 +479,22 @@ private Response invoke(final RequestProcessingContext context, final Object res
         return jaxrsResponse;
     }
 
+    private static Annotation[] mergeDistinctAnnotations(Annotation[] existing, Annotation[] newOnes) {
```
senivam, line 482: what about
```
private static Annotation[] mergeDistinctAnnotations(Annotation[] existing, Annotation[] newOnes) {
        final Set<Annotation> merged = new HashSet<>(existing.length + newOnes.length);
        
        Collections.addAll(merged, existing);
        Collections.addAll(merged, newOnes);
        
        return merged.toArray(new Annotation[0]);
    }
```
?

jansupol: Much slower.

senivam (approved):

PR #5971, opened 2025-08-14
AbortTest contradicts TCK tests

senivam:

PR #5979, opened 2025-08-26
AbortResponse headers case insensitive header names

jansupol:


jansupol:
Updated methods from the original JDK's AbstractMap to compare case-insensitively

jansupol:
Fixes #5971.

senivam (approved):

jbescos (approved):

PR #5991, opened 2025-09-18
cd-inject-weld module-info

jansupol:


jansupol:
This seems to be working, but this might not be what we want, given the tests need to have -add-reads for org.jboss.logging

senivam (approved):

PR #5992, opened 2025-09-19
Adding maven ci friendly versions

cmakinen:
Hi,

These changes for [Maven cifriendly](https://maven.apache.org/guides/mini/guide-maven-ci-friendly.html) versions aren't really needed but should help avoid large diffs when changing snapshot versions.

I made the changes in the 4.0 branch. Please let me know if I need to do it in the default 2.x or another branch.

I initially changed the parent of the bom to the root but I see why the bom parent is different from the root parent after looking at https://github.com/eclipse-ee4j/jersey/pull/4013. Due to this, the flatten configuration needed to be added in the bom module too. So, snapshot version changes only need to happen in two places, the parent and bom modules.

If you would like pluginConfiguration for flatten-maven-plugin to be added in the [ee4j parent](https://github.com/eclipse-ee4j/ee4j), I can do that too. It'll avoid all the configuration being repeated twice here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,4 @@ nb-configuration.xml
 # Maven plugins noise
 dependency-reduced-pom.xml
 pom.xml.versionsBackup
+.flattened-pom.xml
--- a/archetypes/jersey-example-java8-webapp/pom.xml
+++ b/archetypes/jersey-example-java8-webapp/pom.xml
@@ -22,7 +22,7 @@
     <parent>
         <groupId>org.glassfish.jersey.archetypes</groupId>
         <artifactId>project</artifactId>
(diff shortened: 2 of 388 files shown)


senivam:
Thank you @cmakinen, the PR looks great! Having it, I've introduced the [487](https://github.com/mojohaus/flatten-maven-plugin/pull/487) PR to the `flatten-maven-plugin` to have the revision variable name configurable. Anyway, I think we can merge this PR as is and (if required) play with the revision variable name.

senivam (approved):

cmakinen:
Awesome! Thank you @senivam. We started using cifriendly versions last year in our project too. It'll be great if the revision property name can be configurable.

PR #5993, opened 2025-09-22
OSGi Package Incompatibility

Viii3:
Jersey 3.1.11 increased the imported OSGi package version of the microprofile rest client, causing compatibility issues with OSGi dependents.

This PR forces the import version to its prior value to fix the issue.

### DIFF
--- a/ext/microprofile/mp-rest-client/pom.xml
+++ b/ext/microprofile/mp-rest-client/pom.xml
@@ -121,6 +121,7 @@
                             ${cdi.osgi.version},
                             jakarta.decorator.*;version="[3.0,5)",
                             org.eclipse.microprofile.config.*;version="!",
+                            org.eclipse.microprofile.rest.client;version="[2.0,3)",
                             *
                         </Import-Package>
                     </instructions>


ext/microprofile/mp-rest-client/pom.xml
```diff
@@ -121,6 +121,7 @@
                             ${cdi.osgi.version},
                             jakarta.decorator.*;version="[3.0,5)",
                             org.eclipse.microprofile.config.*;version="!",
+                            org.eclipse.microprofile.rest.client;version="[2.0,3)",
```
jansupol, line 124: Thanks for telling us. But the version used to be 3, now it is 4, so the range better be [3.0, 5).

jansupol:
[Missing ECA](https://github.com/eclipse-ee4j/jersey/wiki/Missing-ECA,-Missing-Sign-off-by).

@Viii3 Do you plan to fix the ECA to keep your IP, or do you leave this to us?

ext/microprofile/mp-rest-client/pom.xml
jansupol: Maybe version="!" looks even better.

Pandrex247: This is referring to the explicit package version rather than the spec version - REST Client does per-package OSGi versioning

Pandrex247: > Maybe version="!" looks even better.

Is that the equivalent of simply removing version validation?

Ideally we'd need to do that for all packages (since they're all versioned separately) to prevent only validating _some_ of them

jansupol: Yes, that's to remove the version check, so that OSGi is happy with any version.

jansupol:
Closing in favor of #6003.

PR #5997, opened 2025-09-25
Expand MicroProfile Rest Client OSGi Range

Pandrex247:
Support for MicroProfile REST client 4.0 was added in this PR: https://github.com/eclipse-ee4j/jersey/pull/5831

Given that a TCK runner for version 3.0 was added alongside the new 4.0 version in the same PR I assume the intent was to support both 3.0 and 4.0 versions.

If so, then the OSGi version range configuration of the `org.eclipse.microprofile.rest.client` package needs to expanded to still include the version provided by the 3.0 API.

Without this change, an OSGi runtime still providing the REST Client 3.0 API will fail to resolve.


Pandrex247:
Ah, PR already exists - my mistake: https://github.com/eclipse-ee4j/jersey/pull/5993

PR #5999, opened 2025-09-25
Helidon Container SE Bootstrap

senivam:


containers/helidon/src/main/java/org/glassfish/jersey/helidon/HelidonHttpServer.java
```diff
@@ -0,0 +1,97 @@
...
+            try {
+                this.helidonHttpContainer.stop();
+            } catch (final Exception e) {
+                throw new CompletionException(e);
+            }
+        });
+    }
+
+    @Override
+    public <T> T unwrap(Class<T> nativeClass) {
```
jansupol, line 94: This method should return the Helidon WebServer instance, so that instance should have been kept somewhere, and that cast should be performed on that instance.

senivam: the ``HelidonHttpContainer`` is instance of ``io.helidon.webserver.WebServer``

senivam: However, to obtain the pure underlying instance, adding the ``unwrap()`` method to the container

jansupol (approved):

PR #6015, opened 2025-10-16
Issue #6002 | Ignore warnings for get method consumes entity

### LINKED ISSUE #6002: how to suppress "GET method should not consume any entity" - duplicate
Duplicate of #4563 but I don't have permission to reopen the issue.

Hi!

I have the same issue as in #4563. I am getting warning logs for get methods which take parameters.
I was thinking of filtering it out. However, since other errors are also logged in `org.glassfish.jersey.internal.Errors`, discarding all logs from this class is also suppressing other errors/warnings, which I don't want.

I am having these errors for endpoints for which I have openApi specs. According to the latest version of the [RFC9110](https://datatracker.ietf.org/doc/html/rfc9110#name-get):
> A client SHOULD NOT generate content in a GET request unless it is made directly to an origin server that has previously indicated, in or out of band, that such a request has a purpose and will be adequately supported. 

When reading it like this, if the openApi spec has been shared beforehand, this seems like a valid use-case of parameters in a get request.
Would it maybe be possible to make this warning configurable optional or maybe decrease the level to an INFO instead of warning for this specific error?

Luc-Veldhuis:
Hi! I am trying to solve #6002 as per this comment: https://github.com/eclipse-ee4j/jersey/issues/6002#issuecomment-3353186383 

I am trying to do so by propagating `ServerProperties` from the `RuntimeLocatorModelBuilder`/`ApplicationHolder` to `ComponentModelValidator` using a `Settings` class. The only problem I have is that I don't want to break the API by adding a new constructor, so I need a default Settings class, but that then duplicates the default values (Apparently that also happens in `ApplicationHandler` and `RuntimeLocatorModelBuilder`, I am not sure if that is okay. 

I am not very experienced with contributing to open source, any feedback is welcome!

### DIFF
--- a/core-server/src/main/java/org/glassfish/jersey/server/internal/routing/RuntimeLocatorModelBuilder.java
+++ b/core-server/src/main/java/org/glassfish/jersey/server/internal/routing/RuntimeLocatorModelBuilder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015, 2018 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2015, 2025 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
--- a/core-server/src/main/java/org/glassfish/jersey/server/internal/routing/RuntimeLocatorModelBuilder.java
+++ b/core-server/src/main/java/org/glassfish/jersey/server/internal/routing/RuntimeLocatorModelBuilder.java
@@ -228,7 +228,7 @@ final class RuntimeLocatorModelBuilder {
         Errors.process(new Runnable() {
             @Override
             public void run() {
-                final ComponentModelValidator validator = new ComponentModelValidator(valueSuppliers, messageBodyWorkers);
+                final ComponentModelValidator validator = new ComponentModelValidator(valueSuppliers, messageBodyWorkers, config);
                 validator.validate(component);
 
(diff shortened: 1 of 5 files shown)


jansupol:
Thank you for the contribution.

You are correct, adding an extra constructor could be problematic - maybe there could just be a new method that sets the Settings -

jansupol:
About the Settings - Should that be a new class? Does it make sense to create an extra instance? Is it enough to pass the `Configuration` object?

core-server/src/main/java/org/glassfish/jersey/server/internal/routing/RuntimeLocatorModelBuilder.java
```diff
@@ -141,6 +143,15 @@ public LocatorRouting load(final LocatorCacheKey key) throws Exception {
         });
     }
 
+    private ComponentModelValidator.Settings getSettingsFromConfig(Configuration config) {
```
jansupol: this utility method might be static.

Luc-Veldhuis:
Ah yeah indeed! Passing the `Configuration` is way easier and cleaner. 
When I try to use a setter to set it on the `ComponentModelValidator`, I later have to pass it to the validators, but then I would have to add a new method on the `ResourceModelVisitor` interface, which is a breaking change.
Now I overrode the constructor to either pass or not pass the configuration.
Would creating such an override structure be okay?

jansupol (approved):
Looks good.

senivam (approved):

Issue #6045, opened 2025-12-09
jersey-netty-connector v2.47 : Connector is returning 200 OK when decoded result in HttpResponse contains TooLongHttpHeaderException


sxsubram:
jersey-netty-connector v2.47
JDK 21

In jersey-netty-connector client configuration the following properties are set -
        defaultConfig.property(NettyClientProperties.MAX_HEADER_SIZE, 8192 * 2);
        defaultConfig.property(NettyClientProperties.MAX_INITIAL_LINE_LENGTH, 8192 * 2);

In the test the server is sending a header that exceeds the MAX_HEADER_SIZE. 
invocationBuilder.header("X-LARGE-HEADER", largeHeader);
invocationBuilder.header("X-MY-HEADER", "foo");

When JerseyClientHandler reads the response I see the following in the debugger. 

```
DefaultHttpResponse(decodeResult: failure(io.netty.handler.codec.http.TooLongHttpHeaderException: HTTP header is larger than 16384 bytes.), version: HTTP/1.1)
HTTP/1.1 200 OK
Date: Tue, 09 Dec 2025 22:05:37 GMT
Content-Type: application/json
Content-Length: 22
```
Even though the netty layer creates TooLongHttpHeaderException, this is not captured in the response processing. The future returned the client does not have the required headers causing downstream processing exceptions. 

<img width="578" height="261" alt="Image" src="https://github.com/user-attachments/assets/b4f61e50-11a3-4e14-97e9-ca6967b4028d" />
<img width="1285" height="278" alt="Image" src="https://github.com/user-attachments/assets/fa5904af-6986-49c1-adba-e6ef73556d4b" />

jbescos:
I am not really getting the issue, this is the test I have:

```
public class Issue6045Test extends JerseyTest {

    private static final Logger LOGGER = Logger.getLogger(Issue6045Test.class.getName());

    @Path("/test")
    public static class TestResource {

        @POST
        @Consumes(MediaType.TEXT_PLAIN)
        @Produces(MediaType.TEXT_PLAIN)
        public Response post(String input, @Context HttpHeaders headers) {
            return Response.ok("OK")
                    .header("X-LARGE-HEADER", headers.getHeaderString("X-LARGE-HEADER"))
                    .header("X-MY-HEADER", headers.getHeaderString("X-MY-HEADER"))
                    .build();
        }
    }

    @Override
    protected Application configure() {
        return new ResourceConfig(TestResource.class)
[...]

PR #6047, opened 2026-01-05
jersey-netty-connector v2.47 : Connector is returning 200 OK when decoded result in HttpResponse contains TooLongHttpHeaderException #6045

jbescos:
https://github.com/eclipse-ee4j/jersey/issues/6045


jansupol (approved):

jbescos:
The commit belongs to 2026, but if I update the copyright to 2026, it fails.

Is there some config file related to copyrights that needs to be updated every year?.

sxsubram:
In my standalone test with this change I get 

rsp has exception -
java.util.concurrent.CompletionException: javax.ws.rs.client.ResponseProcessingException: io.netty.handler.codec.http.TooLongHttpHeaderException: HTTP header is larger than 16384 bytes.

senivam (approved):

PR #6049, opened 2026-01-12
Fix a race condition that leads to ClassAnalyzer with name CdiInjecteeSkippingClassAnalyzer exceptions

OndroMih:
This solves the race condition - keep the effective injectionManager in a thread local. As multiple threads can modify the injectionManager and the `InjectionManagerInjectedCdiTarget` instances are global, effectiveInjectionManager must be scoped to the thread. It's actually scoped to a single WebTarget initialization and removed at the end of the initialization in the `done()` method of `CdiComponentProvider`

**The problem:**

A race condition happens when many threads execute REST client call at the same (more specifically, when `WebTarget.request()` is called.

Symptom:

Occasional 
`IllegalStateException: Could not find an implementation of ClassAnalyzer with name CdiInjecteeSkippingClassAnalyzer` from `WebTarget.request`

Reason for the exception:

* The `injectionManager` is set in the initialization phase, before it's used. It's stored to a shared volatile variable and later picked up
* If some thread sets the shared volatile variable to another `injectionManager`, before the original thread uses it, the original thread would use another thread's `injectionManager`
* In case the second thread doesn't bind the `ClassAnalyzer` class in time before the first thread uses it, the first thread won't have `ClassAnalyzer` bean available and throws exception

A reproducer is described in https://github.com/payara/Payara/pull/7753. 

The original fix in Payara fork (https://github.com/eclipse-ee4j/jersey/commit/53ae843516cd6d763c41aaeb7f263a26e32c99b5) addresses this by a workaround - a retry in case of exception, which works, because by the next time, the other thread has enough time to bind the `ClassAnalyzer` class. But that solution isn't efficient.

PR #6053, opened 2026-01-14
Fix a race condition that leads to ClassAnalyzer with name CdiInjecteeSkippingClassAnalyzer exceptions 

arjantijms:
This solves the race condition - keep the effective injectionManager in a thread local. As multiple threads can modify the injectionManager and the `InjectionManagerInjectedCdiTarget` instances are global, effectiveInjectionManager must be scoped to the thread. It's actually scoped to a single WebTarget initialization and removed at the end of the initialization in the `done()` method of `CdiComponentProvider`

**The problem:**

A race condition happens when many threads execute REST client call at the same (more specifically, when `WebTarget.request()` is called.

Symptom:

Occasional 
`IllegalStateException: Could not find an implementation of ClassAnalyzer with name CdiInjecteeSkippingClassAnalyzer` from `WebTarget.request`

Reason for the exception:

* The `injectionManager` is set in the initialization phase, before it's used. It's stored to a shared volatile variable and later picked up
* If some thread sets the shared volatile variable to another `injectionManager`, before the original thread uses it, the original thread would use another thread's `injectionManager`
* In case the second thread doesn't bind the `ClassAnalyzer` class in time before the first thread uses it, the first thread won't have `ClassAnalyzer` bean available and throws exception

A reproducer is described in https://github.com/payara/Payara/pull/7753. 

The original fix in Payara fork (https://github.com/eclipse-ee4j/jersey/commit/53ae843516cd6d763c41aaeb7f263a26e32c99b5) addresses this by a workaround - a retry in case of exception, which works, because by the next time, the other thread has enough time to bind the `ClassAnalyzer` class. But that solution isn't efficient.

(resubmit after https://github.com/eclipse-ee4j/jersey/pull/6049)

### DIFF
--- a/core-client/src/main/java/org/glassfish/jersey/client/ClientConfig.java
+++ b/core-client/src/main/java/org/glassfish/jersey/client/ClientConfig.java
@@ -1,4 +1,5 @@
 /*
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
  * Copyright (c) 2012, 2024 Oracle and/or its affiliates. All rights reserved.
  * Copyright (c) 2018 Payara Foundation and/or its affiliates.
  *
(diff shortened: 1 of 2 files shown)


arjantijms:
(this was already approved in the original PR)

OndroMih:
@arjantijms , isn't it better to reopen my PR and change the target branch?

arjantijms:
> isn't it better to reopen my PR and change the target branch?

Yes, I just did that. Sorry for the noise.

PR #6069, opened 2026-02-02
Exclude jersey-constants from the release test

arjantijms:
From 2.x branch


dmatej:
Could you add description why is that, please?

dmatej (approved):

mkarg:
> Could you add description why is that, please?

@arjantijms Yes please, as it makes it easier for others to review and confirm.

senivam:
I'd say it's only propagating the exclusion from the 2.x branch.

Regarding why it's required, the release test is developed to check modules after build (before release to Sonatype), if they are correctly assembled, if all multireleases are really multirelease jars, if OSGi headers are correct, if there are LICENSE and NOTICE files in place, if NOTICE files contain proper versions vs. the real version of dependencies. In this relation, the constants  module is very simple and should be excluded from the OSGi header check because its OSGi header is not required and is not present. That is why it's being excluded.

senivam (approved):

PR #6070, opened 2026-02-08
Removed ReaderWriter::readAllBytes

mkarg:
As planned for Jersey 4.0 (see notes in the removed code): Getting rid of a worarkound for a bug in one specific application server product (not a bug of Jersey, not a bug of OpenJDK). The vendor had several years of time to fix his own code. Keeping copies of OpenJDK-internals within libraries and applications is a no-go prevents JRE-internal optimizations.

Disclaimer: The contributor is part of the core-libs team at OpenJDK.


mkarg:
Kindly asking [all committers](@eclipse-ee4j/ee4j-jersey-committers)  to review my PR. Thanks.

jansupol (commented):
I am +0 about this.
Pros: less code, maybe JDK will come up with some better code than this one.
Cons: JDK by default does one condition less in the code ( if (nread == BUFFER_SIZE) { // This differs from JDK version
                    break;                  // prevents a bug (See ReaderWriterTest)
                })
This one less condition breaks with some stream implementation (which are incorrect, but ...)

jbescos (approved):
I am not against this, but are there any new enhancements in readAllBytes to want this change now?. It looks we are preventing a bug in some JDKs.

On the other hand, we cannot keep one eye on this forever.

mkarg:
The target of this MR is to allow the JDK to change and optimize the implementation at any time, without waiting for another Jersey release. At OpenJDK we are actively working on optimizations all the time, and my target is to remove anything that prevents applications from automatically gaining the benefits just by upgrading the JVM. So unless someone vetoes *with good reason*, I'm all for *merging* this PR ASAP. As a committer, I hereby veto *closing* it without having seen a veto *against* it.

Besides that, it is not the job of any Open Source Project to workaround bugs of downstream products. In particular, without clearly naming that product and the need to do that.

PR #6075, opened 2026-02-24
Fix memory leak

OndroMih:
`beanManager` was set to threadLocal from a configurator that was post-init  after the ComponentProvider configurator which is supposed to free the thread local. The beanManager  thread local was never removed afterwards. I switched the order of calling post-init so that configurators go in reversed order compared to calling init so that the configurators that depend on other configurators post-init before their dependencies so that everything is properly cleaned up after it's not needed anymore.

### DIFF
--- a/core-server/src/main/java/org/glassfish/jersey/server/ApplicationHandler.java
+++ b/core-server/src/main/java/org/glassfish/jersey/server/ApplicationHandler.java
@@ -1,4 +1,5 @@
 /*
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
  * Copyright (c) 2011, 2025 Oracle and/or its affiliates. All rights reserved.
  * Copyright (c) 2018 Payara Foundation and/or its affiliates.
  *
@@ -44,6 +45,9 @@ import jakarta.ws.rs.core.HttpHeaders;
 import jakarta.ws.rs.core.SecurityContext;
 import jakarta.ws.rs.ext.MessageBodyReader;
 import jakarta.ws.rs.ext.MessageBodyWriter;
+
+import java.util.Collections;
+import java.util.function.Consumer;
 
 import org.glassfish.jersey.CommonProperties;
 import org.glassfish.jersey.innate.inject.Bindings;
@@ -369,7 +373,11 @@ public final class ApplicationHandler implements ContainerLifecycleListener {
 
             injectionManager.completeRegistration();
 
-            bootstrapConfigurators.forEach(configurator -> configurator.postInit(injectionManager, bootstrapBag));
+            /* postInit must be called in reversed order
+             - some configurators (e.g. CDI component provider) perform cleanup (e.g. remove thread locals)
+               and must be called after other configurators that depend on them
+            */
+            forEachInReversedOrder(bootstrapConfigurators, configurator -> configurator.postInit(injectionManager, bootstrapBag));
             resourceModelConfigurator.postInit(injectionManager, bootstrapBag);
 
             Iterable<ApplicationEventListener> appEventListeners =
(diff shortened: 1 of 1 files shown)


core-server/src/main/java/org/glassfish/jersey/server/ApplicationHandler.java
```diff
@@ -369,7 +373,11 @@ private ServerRuntime initialize(InjectionManager injectionManager, List<Bootstr
 
             injectionManager.completeRegistration();
 
-            bootstrapConfigurators.forEach(configurator -> configurator.postInit(injectionManager, bootstrapBag));
+            /* postInit must be called in reversed order
+             - some configurators (e.g. CDI component provider) perform cleanup (e.g. remove thread locals)
+               and must be called after other configurators that depend on them
+            */
+            forEachInReversedOrder(bootstrapConfigurators, configurator -> configurator.postInit(injectionManager, bootstrapBag));
```
OndroMih: In case somebody's curious why not just `bootstrapConfigurators.reversed()` - because Jersey supports Java 17, and `reversed()` requires Java 21+.

dmatej (approved):

OndroMih:
@senivam , can you please help with Jenkins?

mkarg:
Hi @OndroMih, 

> @senivam , can you please help with Jenkins?

I asked Jenkins to perform another build, but it still fails with the same error (see https://ci.eclipse.org/jersey/job/ci-cd-integration/job/PR-6075/2/consoleText): `Could not find a suitable constructor in jakarta.ws.rs.sse.SseEventSink class.`

Does this test pass for you locally: `[INFO] Running org.glassfish.jersey.media.sse.SseEventSinkCloseTest`?

OndroMih:
Those tests also fail locally. The reversed order probably broke something in SSE. I'll investigate it.

OndroMih:
Reversing the order of calling postInit methods for all configurtors broke functionality in SSE injection. So I changed the solution to call specifically componentProviderConfigurator.postInit last, while keeping the original order for other configurators.

This is in line with the previous change in ClientConfig, which addressed a similar issue in the Client, although with different symptoms.

OndroMih:
In this case, ChunkedInputClosedOnErrorTest.testChunkedInputNotStuckedTimes failed. This test passed when I run it locally.

@mkarg @senivam , @arjantijms,  could you restart the Jenkins build?

mkarg:
> In this case, ChunkedInputClosedOnErrorTest.testChunkedInputNotStuckedTimes failed. This test passed when I run it locally.
> 
> @mkarg @senivam , @arjantijms, could you restart the Jenkins build?

Thank you for modifying your fix. I have restartet the Jenkins build.

OndroMih:
@mkarg , can you restart again? This time the build failed on GrizzlyHttpServerProviderTest due to some timeout.

mkarg:
> @mkarg , can you restart again? This time the build failed on GrizzlyHttpServerProviderTest due to some timeout.

Done. 🙂

mkarg:
@OndroMih Jenkins was stuck for six hours, so I killed the run and started another run.

core-server/src/main/java/org/glassfish/jersey/server/ApplicationHandler.java
```diff
@@ -465,6 +485,12 @@ private ServerRuntime initialize(InjectionManager injectionManager, List<Bootstr
         return serverRuntime;
     }
 
+    private static <T> void forEachInReversedOrder(List<T> list, Consumer<? super T> consumer) {
```
mkarg: You are not using this code anymore, so we should get rid of it.

mkarg (changes requested):

core-server/src/main/java/org/glassfish/jersey/server/ApplicationHandler.java
OndroMih: Of course, thanks for noticing it. I force pushed a change that removes it.

mkarg (approved):

jbescos (approved):
LGTM

PR #6083, opened 2026-04-13
Fix race condition in AsyncResponder.resume() and cancel() by re-checking state under write lock

### LINKED ISSUE #6068: Possible race conditions in ServerRuntime.AsyncResponder
Multiple places in ServerRuntime.AsyncResponder check conditions in a read lock, release the read lock, acquire a write lock, and then make changes.  Between releasing the read lock and acquiring the write lock changes could happen that would make the checks done in the read lock no longer valid.  Below is an example of what appears to be a race condition where resume and cancel operations can both run. 

 * Thread1 calls resume() and [here](https://github.com/eclipse-ee4j/jersey/blob/268486ffc7c727542dddcf3a60bfbce24d3efd7b/core-server/src/main/java/org/glassfish/jersey/server/ServerRuntime.java#L952) finds state is SUSPENDED and continues.
 * Thread2 calls cancel() and [here](https://github.com/eclipse-ee4j/jersey/blob/268486ffc7c727542dddcf3a60bfbce24d3efd7b/core-server/src/main/java/org/glassfish/jersey/server/ServerRuntime.java#L1014) find state is SUSPENDED and continues.
 * Thread1 releases read lock
 * Thread 2 release read lock
 * Thread1 obtains the write lock
 * Thread1 sets state to RESUMED [here](https://github.com/eclipse-ee4j/jersey/blob/268486ffc7c727542dddcf3a60bfbce24d3efd7b/core-server/src/main/java/org/glassfish/jersey/server/ServerRuntime.java#L959)
 * Thread1 release the write lock
 * Thread2 obtains the write lock
 * Thread2 sets state to RESUMED and sets canceled to true [here](https://github.com/eclipse-ee4j/jersey/blob/268486ffc7c727542dddcf3a60bfbce24d3efd7b/core-server/src/main/java/org/glassfish/jersey/server/ServerRuntime.java#L1022).  At this point Thread1 has already changed the state to RESUMED, so it does not seem like it should proceed.
 * Thread2 releases the write lock
 * Thread1 initiates resumption [here](https://github.com/eclipse-ee4j/jersey/blob/268486ffc7c727542dddcf3a60bfbce24d3efd7b/core-server/src/main/java/org/glassfish/jersey/server/ServerRuntime.java#L963)
 * Thread2 initiates cancelation [here](https://github.com/eclipse-ee4j/jersey/blob/268486ffc7c727542dddcf3a60bfbce24d3efd7b/core-server/src/main/java/org/glassfish/jersey/server/ServerRuntime.java#L1026)

Only looked at this code in isolation, so not sure if there are other mitigating factors outside this code that would handle this race condition.

anuragg-saxenaa:
## Summary

`ServerRuntime.AsyncResponder` uses a read-check / release / write-acquire pattern in both `resume()` and `cancel()`. A window exists between releasing the read lock and acquiring the write lock where another thread can change the state, making the initial check stale. This allows concurrent `resume()` and `cancel()` calls to both proceed past the state check and both set `state = RESUMED`, resulting in double-processing of a response.

### Race scenario

1. Thread A (`resume()`): acquires read lock → sees `SUSPENDED` → releases read lock
2. Thread B (`cancel()`): acquires read lock → sees `SUSPENDED` → releases read lock
3. Thread A: acquires write lock → sets `RESUMED` → releases write lock → processes response
4. Thread B: acquires write lock → sets `RESUMED`, `cancelled = true` → processes response again

## Fix

Add a second state check _inside_ the write-lock critical section in both `resume()` and `cancel()`. If state changed between the read and write lock acquisitions, return `false` immediately. Also wrapped write-lock body in `try/finally` so the lock is always released even if an exception occurs in the re-check.

```java
stateLock.writeLock().lock();
try {
    // Re-check under write lock: another thread may have changed state
    // between releasing the read lock and acquiring the write lock.
    if (state != SUSPENDED) {
        return false;
    }
    state = RESUMED;
} finally {
    stateLock.writeLock().unlock();
}
```

This is the standard double-checked locking pattern for `ReadWriteLock`.

Fixes #6068

### DIFF
--- a/core-server/src/main/java/org/glassfish/jersey/server/ServerRuntime.java
+++ b/core-server/src/main/java/org/glassfish/jersey/server/ServerRuntime.java
@@ -956,8 +956,16 @@ public class ServerRuntime {
                 stateLock.readLock().unlock();
             }
             stateLock.writeLock().lock();
-            state = RESUMED;
-            stateLock.writeLock().unlock();
+            try {
+                // Re-check under write lock: another thread may have changed state
+                // between releasing the read lock and acquiring the write lock.
+                if (state != SUSPENDED) {
+                    return false;
+                }
+                state = RESUMED;
+            } finally {
+                stateLock.writeLock().unlock();
+            }
 
             try {
                 responder.runtime.requestScope.runInScope(requestContext, handler);
@@ -1019,9 +1027,20 @@ public class ServerRuntime {
             }
 
             stateLock.writeLock().lock();
-            state = RESUMED;
-            cancelled = true;
-            stateLock.writeLock().unlock();
+            try {
+                // Re-check under write lock: another thread may have changed state
+                // between releasing the read lock and acquiring the write lock.
+                if (cancelled) {
+                    return true;
+                }
+                if (state != SUSPENDED) {
+                    return false;
+                }
+                state = RESUMED;
+                cancelled = true;
+            } finally {
+                stateLock.writeLock().unlock();
+            }
 
             responder.runtime.requestScope.runInScope(requestContext, new Runnable() {
                 @Override


dmatej:
Hi @anuragg-saxenaa , thanks for the contribution, however to merge your contribution we need you to create an account on the Eclipse website, and sign ECA License Agreement.
https://www.eclipse.org/legal/eca/

And yet another thing, could you add this line to the start of the file? That should fix the CI validation of the commit.
```diff
 /*
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation

```

dmatej:
@anuragg-saxenaa Hi, have you considered signing the ECA license? We cannot merge your PR without that.

PR #6088, opened 2026-06-07
Fix memory leak (port to 3.1)

OndroMih:
Backport of https://github.com/eclipse-ee4j/jersey/pull/6075 from 4.x

* Fix memory leak

beanManager was set to threadLocal from a configurator that was post-init  after the ComponentProvider configurator which is supposed to free the thread local. The beanManager  thread local was never removed afterwards. I switched the order of calling post-init so that configurators go in reversed order compared to calling init

* Fix memory leak - fix failing tests

Revert the change to run postInit on all configurators in reversed order. Instead run postInit only on the specific Component configurator last, while initialize runs in the original order. This cleans up thread locals safely, without changing the order of other configurators, which was breking other functionality.

### DIFF
--- a/core-server/src/main/java/org/glassfish/jersey/server/ApplicationHandler.java
+++ b/core-server/src/main/java/org/glassfish/jersey/server/ApplicationHandler.java
@@ -1,4 +1,5 @@
 /*
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
  * Copyright (c) 2011, 2025 Oracle and/or its affiliates. All rights reserved.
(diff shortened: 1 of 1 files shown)


OndroMih:
The Validate/ Api docs check failed on:

```
"Could not transfer artifact org.eclipse.persistence:org.eclipse.persistence.moxy:pom:4.0.7 from/to central (https://repo.maven.apache.org/maven2): Connection reset
```

@senivam , @arjantijms , can you restart the failed git action?

dmatej (approved):

dmatej (approved):

PR #6092, opened 2026-06-09
Fixed schemaLocation for poms for 5.0

dmatej:
All older schemalocations changed to the usual current URL.


dmatej:
The failure is fixed in https://github.com/eclipse-ee4j/jersey/pull/6093 - Grizzly NPN API 2.0.0 used just automatic module name and did not declare any module-info.java. 2.0.1 fixed that.

jbescos (approved):

PR #6093, opened 2026-06-09
Updated Grizzly dependencies - fixed jpms

dmatej:
Grizzly NPN API 2.0.0 used just automatic module name and did not declare any module-info.java. 2.0.1 fixed that.


mkarg:
@dmatej Thank you for this contribution. As I am a new committer on Jersey I have one question. Do we typically just merge directly our own commits without waiting for any review comments of other committers?

arjantijms:
Yes indeed. Of course after commit checking and reviews are still appreciated, and we can do new commits to fix anything if needed then.

PR #6105, opened 2026-07-20
Improved ApplicationInfoTest - avoiding NPE in logs

dmatej:
- I believe it is better to see 204 than 500, which is scary on the first sight


mkarg (commented):
I don't know *why* this test originally uses `500`, OTOH I have no problem with it, hence I abstain from a vote here. If other committers do not veto, feel free to proceed.

arjantijms (approved):

PR #6109, opened 2026-07-22
Updated Netty and stabilized ChunkedInputClosedOnErrorTest

dmatej:
Netty changelog: 
* https://github.com/netty/netty/releases
* https://github.com/netty/netty/compare/netty-4.1.126.Final...netty-4.1.136.Final

### ChunkedInputClosedOnErrorTest Changes

* Flush doesn't guarantee the end of input, however it should be close.
* Refactored ChunkedInputClosedOnErrorTest to give us more info
* The test randomly failed, however now it checks the exception, runs more iterations (300 or more instead of 10), and is more readable.


mkarg:
Basically +1 if not Jenkins would fail :-(

dmatej:
> Basically +1 if not Jenkins would fail :-(

That is the other flaky test on my radar, tls ... restarted.

PR #6113, opened 2026-07-25
Updated eclipse parent to 2.0.5 for 2.x branch

dmatej:
Note: Will require update of release jobs.


dmatej:
`/check-license`

EDIT: Proven that this works even if the change was not backported here yet. Proven also that when the job fails (http 504), it doesn't block merging (the original script was).

PR #6112, opened 2026-07-25
Enhanced License Check

dmatej:
- As we are getting blacklisted from data websites, we have to limit the usage
- Now the check can be executed manually using /check-license comment
- It will be still manually executed daily on main branch
- It can be executed also from the Actions menu
- Removed staging profile
- Switched to Temurin JDK
- Related GitLab issue: https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/work_items/7694

### Validation of this PR

* First I pushed the updated action file to the branch 4.x in my own repository and I set it as the main branch.
* Then I created another branch with any trivial change, and PR to 4.x
* Then I commented the PR with `/check-license` (`/license-check` works too)

The check passed, and the PR automatically received this comment: https://github.com/dmatej/jersey/pull/126#issuecomment-5078953607

Then I closed the PR. Failure should not block merge of the PR. 

### Notes

* The comment will not work until we merge this PR, that is why I had to use my own repo for testing.
* The summary in the comment is trimmed by GitHub, the original file can be downloaded from the action run.

PR #6120, opened 2026-08-17
Terminate SseEventSource gracefully on HTTP 204 during (re)connect

### LINKED ISSUE #6119: SseEventSource: HTTP 204 on (re)connect should terminate the event source gracefully instead of reconnecting forever
This issue surfaced while running the OSGi "Whiteboard Specification for Jakarta RESTful Web Services" TCK against a Jersey 4 based whiteboard implementation (eclipse-osgi-technology/jakartarest-osgi): the TCK's SSE test case fails with every Jersey version since 3.1.8, and the root cause turned out to be in the SSE client.

`SseEventSource` currently provides no way for a server to tell the client to stop reconnecting. According to the WHATWG HTML specification for `EventSource` : 
> A client can be told to stop reconnecting using the HTTP 204 No Content response code.

Jersey's `EventProcessor` does not implement this. When the handshake request of a re-connect attempt returns **204 No Content**, the client schedules the next reconnect and keeps polling the endpoint forever. The registered `onComplete` callback is never called, except oin the manual `close()` or, since 3.1.8, after an error.

## Why this became visible now

Up to and including 3.1.7 (and 3.0.x before the backport), connection errors during reconnect were silently swallowed and simply shut the event source down. This obviously accidentially acted as  "termination signal" for servers that answered a reconnect with
an error status. #5693 fixed that (correctly!) in 3.1.8 by delivering such errors to the registered `onError` consumers.

Since then there is **no response a server can send that terminates the client cleanly**:

| Response on reconnect  | Client behavior since 3.1.8                                  |
| ---------------------- | ------------------------------------------------------------ |
| 2xx stream that closes | schedule next reconnect (forever)                            |
| **204 No Content**     | **silent infinite reconnect loop, `onComplete` never fires** |
| 503 (+ Retry-After)    | reconnect                                                    |
| any other error status | `onError` + shutdown                                         |

This is exactly the scenario of the OSGi TCK test expoerienced. The server completes the stream, answers the subsequent reconnect with a termination signal, and the client is expected to complete without an error, which doesnt work with the current behaviore.

I already looked into it and would provide a fix PR.

maho7791:
- EventProcessor now inspects the handshake response status: a 204 shuts the event source down gracefully using onComplete callback instead of scheduling further reconnects.
- All other non-successful statuses are translated into the same WebApplicationException subtypes.
- The translation logic is extracted from JerseyInvocation into a reusable static method instead of being duplicated.
- Addes corresponding tests, that shows that new expecged behavior.
 
- Fixes #6119

### DIFF
--- a/core-client/src/main/java/org/glassfish/jersey/client/JerseyInvocation.java
+++ b/core-client/src/main/java/org/glassfish/jersey/client/JerseyInvocation.java
@@ -941,46 +941,7 @@ public class JerseyInvocation implements jakarta.ws.rs.client.Invocation {
             // leaking connections (see JERSEY-2157).
             response.bufferEntity();
 
-            final WebApplicationException webAppException;
-            final Response.Status status = Response.Status.fromStatusCode(statusCode);
-
-            if (status == null) {
-                final Response.Status.Family statusFamily = finalResponse.getStatusInfo().getFamily();
-                webAppException = createExceptionForFamily(finalResponse, statusFamily);
-            } else {
-                switch (status) {
-                    case BAD_REQUEST:
-                        webAppException = new BadRequestException(finalResponse);
-                        break;
-                    case UNAUTHORIZED:
-                        webAppException = new NotAuthorizedException(finalResponse);
-                        break;
-                    case FORBIDDEN:
-                        webAppException = new ForbiddenException(finalResponse);
-                        break;
-                    case NOT_FOUND:
-                        webAppException = new NotFoundException(finalResponse);
-                        break;
-                    case METHOD_NOT_ALLOWED:
-                        webAppException = new NotAllowedException(finalResponse);
-                        break;
-                    case NOT_ACCEPTABLE:
-                        webAppException = new NotAcceptableException(finalResponse);
-                        break;
-                    case UNSUPPORTED_MEDIA_TYPE:
-                        webAppException = new NotSupportedException(finalResponse);
-                        break;
-                    case INTERNAL_SERVER_ERROR:
-                        webAppException = new InternalServerErrorException(finalResponse);
-                        break;
-                    case SERVICE_UNAVAILABLE:
-                        webAppException = new ServiceUnavailableException(finalResponse);
-                        break;
-                    default:
-                        final Response.Status.Family statusFamily = finalResponse.getStatusInfo().getFamily();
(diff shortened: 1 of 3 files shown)


core-client/src/main/java/org/glassfish/jersey/client/JerseyInvocation.java
```diff
@@ -989,7 +950,63 @@ private ProcessingException convertToException(final Response response) {
...
+     * @param response the response to convert.
+     * @return the {@link WebApplicationException} subtype matching the response status code.
+     */
+    public static WebApplicationException convertToWebApplicationException(final Response response) {
+        final WebApplicationException webAppException;
+        final Response.Status status = Response.Status.fromStatusCode(response.getStatus());
+
+        if (status == null) {
+            final Response.Status.Family statusFamily = response.getStatusInfo().getFamily();
+            webAppException = createExceptionForFamily(response, statusFamily);
```
dmatej: Could you simplify that and simply return the result, then push the code off the `else` branch? And perhaps return from all cases, it will make the code shorter and easier to read - now I had to carefully read what else yet can happen to the `webAppException` variable.

pzygielo: @dmatej - could you please check https://github.com/eclipse-ee4j/jersey/pull/6112#issuecomment-5409165343?

dmatej: I did right before this comment, it took me a while to find and read definitions of those groups again. First I thought you could be a collaborator, but we don't have access to specify collaborators. 
https://docs.github.com/en/organizations/managing-peoples-access-to-your-organization-with-roles/roles-in-an-organization
I believe it is ok to allow all members to run the check. Bigger problem is with Jenkins in all projects, I don't know how to configure that waiting for approvals just like we have them for GitHub Actions. And especially with limited resources on Jenkins it is probably easy to overload it.

maho7791: I appended a commit with the suggested changes:
https://github.com/eclipse-ee4j/jersey/pull/6120/changes/22e60851106e55cc3374d56d5d24d7790e64bfa6

dmatej (approved):

